PR #12, opened 2017-09-06
Rework spring-cloud-function-web-handler

### LINKED ISSUE #11: Make spring-cloud-function-web-handler more versatile
Should be able to work with more that Function<String, String>

ericbottard:
Rework spring-cloud-function-web-handler to accept more functional types (including at input/output level)

Fixes #11

### DIFF
--- a/spring-cloud-function-web-handler/src/main/java/io/sk8s/handler/spring/web/invoker/FunctionConfiguration.java
+++ b/spring-cloud-function-web-handler/src/main/java/io/sk8s/handler/spring/web/invoker/FunctionConfiguration.java
@@ -17,28 +17,39 @@
 package io.sk8s.handler.spring.web.invoker;
 
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.net.URL;
 import java.net.URLClassLoader;
-import java.util.ArrayList;
-import java.util.List;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Consumer;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
+import org.springframework.beans.BeanUtils;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.deployer.resource.maven.MavenProperties;
 import org.springframework.cloud.deployer.resource.maven.MavenResource;
 import org.springframework.cloud.deployer.resource.maven.MavenResourceLoader;
-import org.springframework.cloud.function.support.FluxFunction;
-import org.springframework.cloud.function.support.FunctionFactoryUtils;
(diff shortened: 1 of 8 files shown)


spring-cloud-function-web-handler/src/main/java/io/sk8s/handler/spring/web/server/SpringWebHandlerController.java
```diff
@@ -47,37 +42,42 @@
...
-		this.function = function;
-	}
+	private volatile Object function;
 
 	@PostMapping("/init")
-	@SuppressWarnings("unchecked")
-	public void init(@RequestBody String uri, @RequestParam String classname) {
-		if (initialized.compareAndSet(false, true)) {
+	// TODO: use JSON with both uri and classname (eventually dependencies)
+	public void init(@RequestParam String uri, @RequestParam String className) {
```
markfisher: the uri is currently expected in the request body (not param)

spring-cloud-function-web-handler/src/main/java/io/sk8s/handler/spring/web/server/SpringWebHandlerController.java
```diff
@@ -16,27 +16,22 @@
...
-import org.springframework.web.bind.annotation.PostMapping;
-import org.springframework.web.bind.annotation.RequestBody;
-import org.springframework.web.bind.annotation.RequestParam;
-import org.springframework.web.bind.annotation.ResponseStatus;
-import org.springframework.web.bind.annotation.RestController;
-
-import io.sk8s.handler.spring.web.invoker.FunctionConfiguration;
-
-import reactor.core.publisher.Flux;
+import org.springframework.web.bind.annotation.*;
```
markfisher: should avoid wildcard imports

markfisher:
This doesn't work for me following the README steps for the demo. I did (locally) make the change for the URI param to be passed as the request body as I mentioned in a comment above.

spring-cloud-function-web-handler/src/main/java/io/sk8s/handler/spring/web/invoker/FunctionConfiguration.java
```diff
@@ -49,56 +60,71 @@
...
+		try (URLClassLoader cl = new URLClassLoader(urls, /* explicit null parent */null)) {
+			return Arrays.stream(properties.getClassName())
+					.map(name -> (Object) BeanUtils.instantiateClass(ClassUtils.resolveClassName(name, cl)))
+					.sequential()
+					.reduce(this::compose).get();
+		}
+	}
+
+	@SuppressWarnings("unchecked")
+	private Object compose(Object a, Object b) {
```
ericbottard: Also, did not mention this otp, but I guess something similar to this could move to s-c-function

ericbottard:
Updated the PR, using JSON object for init.
Also had to tweak stuff to support String payloads (I tested with `application/json` and integer `2` :( )

Not 100% sure this does the full roundtrip as I don't know if we're intended to witness the return from the function, but the return value reaches the `HandlerPool` alright

markfisher:
squashed and merged as https://github.com/markfisher/sk8s/commit/6727587a0dfc3c316e869e939391eac00c0a4856

PR #20, opened 2017-09-25
Introduce our variation of the fabric8 client Use Spring event publishing

ericbottard:


event-dispatcher/src/main/java/io/sk8s/event/dispatcher/JobLauncher.java
```diff
@@ -75,27 +78,35 @@ public void dispatch(String payload, Map<String, Object> headers, FunctionResour
...
+				.build()
+			)
 			.done().toString();
+		// @formatter:on
+
 		System.out.println("JOB: " + job);
 	}
 
-	private List<String> resolvePlaceholders(List<String> original, FunctionResource functionResource) {
+	private List<String> resolvePlaceholders(List<String> original, XFunction functionResource) {
```
ericbottard: Why not use the `$()` notation that K8S supports if we're ready to eg create an artificial ENV_VAR?
See http://jeffmendoza.github.io/kubernetes/v1.1/docs/design/expansion.html

markfisher: +1 I was planning to look into exactly that... thanks ;)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/JobLauncher.java
```diff
@@ -104,23 +115,17 @@ public void dispatch(String payload, Map<String, Object> headers, FunctionResour
...
-		int i = 0;
-		for (Map.Entry<String, String> entry : map.entrySet()) {
-			vars[i++] = new EnvVarBuilder()
-					.withName(entry.getKey())
-					.withValue(entry.getValue())
-					.build();
-		}
-		return vars;
+	private EnvVar[] buildEnvVars(List<FunctionEnvVar> envList, String payload) {
+		// Put the payload under the "MESSAGE" key and any envvar that had valueFrom() equal
```
ericbottard: Is this a long term thing?
What will we do when payload could be anything, not just String?
Should we dump it to a Volume?

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/JobLauncher.java
ericbottard: Done via eb6ba1b

README.adoc
```diff
@@ -8,6 +8,19 @@ these instructions assume minikube for now, for example:
 minikube start --cpus=4 --memory=4096 --disk-size=64g
 ```
 
+== Build the kubernetes-model jar Enhanced with Sk8s Types
```
trisberg: Seems a bit much to require this step for the demo, seems to require some serious go chops to get his to even work ...

ericbottard: See updated instructions, need to be cloned inside $GOPATH

markfisher:
rebased and merged via https://github.com/markfisher/sk8s/commit/5a5ffd29c8b5965f10cf42063621d2066ab54cb6

PR #21, opened 2017-09-25
Adding Helm chart for sk8s

trisberg:


publish
```diff
@@ -1,3 +1,4 @@
 #!/bin/bash
 
-curl -H "Content-Type: text/plain" $(minikube ip):30111/messages/$1 -d "$2"
+port=$(kubectl get svc | grep topic-gateway | cut -d ":" -f 2 | cut -d "/" -f 1)
```
jldec: it's not a lot cleaner, but you could extract one of the ports (http or https) using jsonpath 
`get svc topic-gateway -o jsonpath='{.spec.ports[?(@.name == "http")].port}'`

trisberg: Nice, updated the PR, it makes the intent clearer instead of chopping on delimiters

markfisher:
rebased, squashed, and merged via https://github.com/markfisher/sk8s/commit/aad0f82c7c41729f7658a8a03da360bb5ab62014

PR #22, opened 2017-09-28
Adding Prometheus

trisberg:
- adding /prometheus endpoint to spring web apps for scraping the metrics

- using 'com.moelholm:prometheus-spring-boot-starter' see: https://moelholm.com/2017/02/06/spring-boot-prometheus-actuator-endpoint/

- adding k8s yaml config files for prometheus and grafana

### DIFF
--- a/config/prometheus/prometheus-configmap.yaml
+++ b/config/prometheus/prometheus-configmap.yaml
@@ -1,0 +1,27 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: prometheus
+  labels:
+    app: metrics
+    component: prometheus
+data:
+  prometheus.yml: |-
+    global:
+      scrape_interval: 15s
+    scrape_configs:
+    - job_name: 'prometheus'
+      # Override the global default and scrape targets from this job every 5 seconds.
+      scrape_interval: 5s
+      # metrics_path defaults to '/metrics'
+      # scheme defaults to 'http'.
+      static_configs: 
+      - targets: ['localhost:9090']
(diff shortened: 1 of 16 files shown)


config/prometheus/prometheus-configmap.yaml
```diff
@@ -0,0 +1,27 @@
...
+      static_configs: 
+      - targets: ['localhost:9090']
+    - job_name: 'sk8s'
+      # Override the global default and scrape targets from this job every 5 seconds.
+      scrape_interval: 5s
+      # metrics_path defaults to '/metrics'
+      # scheme defaults to 'http'.
+      metrics_path: '/prometheus'
+      static_configs:
+      - targets: ['topic-gateway:80','topic-controller:80','event-dispatcher:80','spring-handler:80','greeter:80']
```
markfisher: is there any way to add targets dynamically ("greeter" is just one example of a Function... to an extent same applies for any handlers)

trisberg: I think we cold use `<dns_sd_config>` to query for dns hostnames to include 
- https://prometheus.io/docs/operating/configuration/#dns_sd_config

markfisher: ok, cool - something to keep in mind (maybe create an issue to make the targets dynamic?)

markfisher:
squashed, rebased, and merged via https://github.com/markfisher/sk8s/commit/b05b017a032b98fdb3ed98f2d9d0c14e061ecae7

PR #29, opened 2017-10-10
Add script handler with dynamic loading

### LINKED ISSUE #30: [sidecar] Provide a shell invoker image that only expects a user-provided shell script
e.g. its ENTRYPOINT would be something like:
['while read X; do $SCRIPT $X; done</pipes/input >/pipes/output']

ericbottard:
Fixes #30


function-invokers/shell-function-invoker/src/main/docker/image/runner.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+set -e
+
+source $FUNCTION_URI
+
+while read X; do echo $(userfunction "$X"); done </pipes/input >/pipes/output
```
markfisher: if we're going to be prescriptive about a function name, maybe it should be "main"?

samples/script/wordcount.sh
```diff
@@ -0,0 +1,3 @@
+function userfunction {
```
markfisher: why not just a shell script as the entry point?

function-invokers/shell-function-invoker/pom.xml
```diff
@@ -0,0 +1,111 @@
...
+								</image>
+							</images>
+						</configuration>
+					</execution>
+				</executions>
+			</plugin>
+		</plugins>
+	</build>
+
+	<profiles>
```
markfisher: I don't think we need the spring repos here. We are only depending on the fabric8 docker plugin.

function-invokers/shell-function-invoker/src/main/docker/image/runner.sh
dturanski: Couldn't the name of the function be an env var?

markfisher: well we could use the same approach as elsewhere (needs to change to either a proper "query param" or a `#` fragment I guess): `FUNCTION_URI=file:///foo/bar.sh?function=whatever`

That would also mean it could be optional, so that the script itself could be invoked directly when the user doesn't want to wrap their logic in a function (as could be the case with the simple `wc` example).

ericbottard:
Addressed review comments (without the `#` or `?function=` part for now)

markfisher:
merged along with https://github.com/markfisher/sk8s/commit/529934b6d5f8facf4135dc3b348ae7940c97bc0b

PR #38, opened 2017-10-12
separate k8s model build into bash script

jldec:
Trying to streamline the readme just a little
- moved git clone of the fabric8 repo up into the prerequisites section.
- added `build-model` bash script 

The bash script can be modified for other branches if they use different models.  
Developers or demo users run the same same build steps when they switch between branches.


README.adoc
```diff
@@ -1,24 +1,30 @@
...
 
-* a running 1.7 Kubernetes cluster with at least 4GB of memory
-these instructions assume minikube for now, for example:
+* A running 1.7 Kubernetes cluster with at least 4GB of memory
+these instructions assume minikube for now. Point your docker environment
+to the docker running in minikube.
 +
 ```
 minikube start --cpus=4 --memory=4096 --disk-size=64g
+eval $(minikube docker-env)
```
markfisher: can we remove line 36 if we have this here?

README.adoc
```diff
@@ -1,24 +1,30 @@
...
 ```
 
-* A working Go programming environment, with `GOPATH` set (defaults
-to `~/go`)
-
-== Build the kubernetes-model jar Enhanced with Sk8s Types
+* A working Go environment, with `GOPATH` set (defaults
+to `~/go`) and a clone of ericbottard's customized fabric8io/kubernetes-model.
```
markfisher: this is slightly confusing now (and in hindsight kind of was before), but the previous behavior was such that `GOPATH` did not need to be set *unless* something other than `~/go` was desired. Now, the script actually does require that it be set. Do we want to make the script more flexible or require it to be set? cc @ericbottard

README.adoc
jldec: i thought about removing it there, but since the docker-env needs to be set for every terminal session i thought it best to keep it in both places.

README.adoc
jldec: the script fails immediately if the GOPATH is not set so i think it's ok.
```
$ unset GOPATH
$ ./build-model 

# regenerate the fabric8 java API including k8s custom resources
pushd $GOPATH/src/github.com/fabric8io/kubernetes-model/
./build-model: line 5: pushd: /src/github.com/fabric8io/kubernetes-model/: No such file or directory
$ 
```

ericbottard: `go env GOPATH` retrieves the effective value of `GOPATH`

markfisher: so should we add that to the script?

ericbottard: I think we should replace occurences of `$GOPATH` with `go env GOPATH`

jldec: 👍  - i will push another commit

README.adoc
```diff
@@ -1,24 +1,30 @@
...
++
 ```
 cd $GOPATH   #default to ~/go
```
ericbottard: Change here as well

jldec: 👍

markfisher:
rebased, squashed and merged via https://github.com/markfisher/sk8s/commit/2ee43941238439ef97cbb67a1ea2200833a87db8

PR #41, opened 2017-10-13
Implement dynamic scaling based on consumer lag

### LINKED ISSUE #33: [sidecar] add support for scaling to/from 0
Assigned: @ericbottard

1) monitor topics (or the offset topic)
2) launch jobs on demand
3) add idle-timeout so that jobs can self-terminate
(also investigate job scaling... change to ReplicaSet/StatefulSet or manage Pods if necessary)

ericbottard:
Fixes #33


README.adoc
```diff
@@ -15,7 +15,7 @@ to `~/go`)
 cd $GOPATH   #default to ~/go
 git clone https://github.com/ericbottard/kubernetes-model src/github.com/fabric8io/kubernetes-model/
 cd src/github.com/fabric8io/kubernetes-model/
-git checkout sk8s-sidecar
+git checkout sk8s-sidecar-scaling
```
ericbottard: Pay attention to this (esp if merging after https://github.com/markfisher/sk8s/pull/38)

ericbottard:
This currently does not pass tests due to hardcoded `kafka:9092`

ericbottard:
Rebased, make tests pass if running against local kafka.
Ready for review

README.adoc
```diff
@@ -10,7 +10,7 @@ eval $(minikube docker-env)
 ```
 
 * A working Go environment, with a clone of ericbottard's customized fabric8io/kubernetes-model.
-
++
```
markfisher: is this `+` intentional? (I had removed it earlier I think)

ericbottard: Nope, certainly a rebase glitch

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -41,30 +51,122 @@
...
+					}
+				}
+			}
+		}
+
+		private void logOffsets(Map<KafkaConsumerMonitor.TopicGroup, List<KafkaConsumerMonitor.Offsets>> offsets) {
+			for (Map.Entry<KafkaConsumerMonitor.TopicGroup, List<KafkaConsumerMonitor.Offsets>> entry : offsets.entrySet()) {
+				System.out.println(entry.getKey());
+				for (KafkaConsumerMonitor.Offsets values : entry.getValue()) {
+					System.out.println("\t" + values + " Lag=" + values.getLag());
```
markfisher: this should probably be debug level logging? (same for the one 2 lines above)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/FunctionDeployer.java
```diff
@@ -17,80 +17,81 @@
...
+					.endSpec()
+				.done();
 		// @formatter:on
-		System.out.println("JOB: " + job);
 	}
 
 	private PodSpec buildPodSpec(XFunction function) {
 		PodSpecBuilder builder = new PodSpecBuilder()
-				.withRestartPolicy("OnFailure")
+				.withRestartPolicy("Always")
```
markfisher: this is the default for a `Deployment`

samples/python/uppercase/uppercase.yaml
```diff
@@ -12,6 +12,7 @@ spec:
   image: sk8s/grpc-py:latest
   protocol: grpc
   input: greetings
+  output: devnull
```
markfisher: we need a better solution for no-output cases (shouldn't involve any actual publishing)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/FunctionDeployer.java
ericbottard: Yeah, had "OnFailure" left over from rebase, got an error and changed explicitly to Always. Can remove.

samples/python/uppercase/uppercase.yaml
ericbottard: Yes. This is unrelated to this PR, rather a requirement of the go sidecar (which unconditionally creates a Publisher for now)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -41,30 +51,122 @@
...
+									.mapToLong(KafkaConsumerMonitor.Offsets::getLag)
+									.max().getAsLong()
+						));
+				functions.values().parallelStream().forEach(
+						f -> {
+							String expression = f.getSpec().getScalingStrategy() != null ? f.getSpec().getScalingStrategy().getReplicasExpression() : "1";
+							Expression spel = spelExpressionParser.parseExpression(expression);
+							EvaluationContext context = new StandardEvaluationContext();
+							context.setVariable("lag", lags.get(f.getMetadata().getName()));
+							int replicas = spel.getValue(context, Integer.class);
```
markfisher: I'm trying to see if I understand how this is working with the current example (using `#lag` as the expression)... that would create N replicas for N unconsumed events? Given that each instance can process multiple messages, that seems unnecessary in most cases. Also, there doesn't seem to be any relationship between the instance being created and the partition on which lag exists, correct?

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -41,30 +51,122 @@
...
+
 		logger.info("function added: " + functionName);
 	}
 
+
 	@EventListener
 	public void onFunctionUnregistered(ResourceDeletedEvent<XFunction> event) {
 		XFunction functionResource = event.getResource();
 		String functionName = functionResource.getMetadata().getName();
-		this.kubernetesClient.extensions().jobs().withLabel("function", functionName).delete();
```
markfisher: shouldn't we be deleting the `Deployment` in this case?

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
ericbottard: Yes this would. Obviously this is disproportionate, and we'd use some function (ah ah) to bring this number in a reasonable range. Including division, logarithm, Math.max, etc.

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -41,30 +51,122 @@
...
+							context.setVariable("lag", lags.get(f.getMetadata().getName()));
+							int replicas = spel.getValue(context, Integer.class);
+							// TODO no more replicas than topic.partitions
+							deployer.deploy(f, replicas);
+						}
+				);
+				
+				synchronized (mutex) {
+					try {
+						mutex.wait(scalingFrequencyMs);
```
markfisher: isn't there a way to make this whole thing based on push notifications from the actual offset monitoring consumer instead of a polling interval?

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
ericbottard: > Also, there doesn't seem to be any relationship between the instance being created and the partition on which lag exists, correct

Correct. I didn't look into that yet (seemed orthogonal to pure lag detection / scaling).
But some rebalancing should occur automatically, shouldn't it?

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
ericbottard: Yes, forgot to port that part

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -41,30 +51,122 @@
...
 
-	private final KubernetesClient kubernetesClient;
+	private volatile long scalingFrequencyMs = 0L; // Wait forever
 
-	public EventDispatchingHandler(KubernetesClient kubernetesClient) {
-		this.kubernetesClient = kubernetesClient;
-	}
+	private final Thread scalingThread = new Thread(new ScalingMonitor());
+
+	private final Object mutex = new Object();
```
markfisher: seems like there are simpler (and more efficient) ways to handle the rendezvous via `java.util.concurrent` types

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
ericbottard: Not sure without actually consuming the messages

samples/python/uppercase/uppercase.yaml
markfisher: understood - let's create an issue for that so we know it's tracked

README.adoc
ericbottard: Actually I'm sorry, YES it is intentional. It fuses the code snippet with the list item (otherwise it is not correctly indented)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
markfisher: for simple load-balancing cases, yes... but we should (eventually) support more fine-grained control (if there are events on 2 partitions of a topic having 10 partitions, we would ideally start 2 consumers and handle partition assignment)... or another way to consider it with the current implementation is that starting 10 instances will not be any better than 2 instance in such a case

samples/python/uppercase/uppercase.yaml
ericbottard: https://github.com/ericbottard/function-sidecar/issues/1

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
markfisher: regarding the various ways to determine the scaling value, I think initially we should just implement one approach and not worry about exposing it to users as a configurable option yet (that can come later)

ericbottard: I agree. This is also why I was interested in the HPA definition which has `min, max, targetMetricValue`. Still need to look into it in more details to see if the scaler "understands" what it's doing or is just assuming more pods = reduced metric value.
Here we'd strive for target=0

README.adoc
markfisher: nice - I didn't know about that

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
ericbottard: I'm far from a concurrency expert, but my primary goal here is the `wait`ing that can be woken up early. What `java.util.concurrent` primitive helps with that? Let's discuss on the call (if this stays at all, that is)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
markfisher: what's wrong with consuming the messages? (they are only "consumed" from this consumer group's perspective)

ericbottard: The fact that they're sent on the wire, processed, etc just to be thrown on the floor.
Maybe there is a way to get only the notification?

markfisher:
not sure if you've already looked at these examples @garyrussell created in his sandbox repo...

using boot:
https://github.com/garyrussell/sandbox/blob/master/consumer-offsets-boot15/src/main/java/com/example/ConsumerOffsetsBoot15Application.java

lower-level:
https://github.com/garyrussell/sandbox/blob/master/consumer-offsets/src/main/java/com/example/ConsumerOffsetsApplication.java

ericbottard:
Interesting. I originally thought about looking at the internal offset tracking topic but somehow disregarded that as no-one mentioned that as a viable solution. Seems like the most elegant solution

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
markfisher: that doesn't seem like an issue if we're monitoring the offsets topic itself... pasted the examples from Gary's repo in another comment

markfisher:
my mistake for not emphasizing it more I guess, but when we first discussed this I had mentioned my PoC using the offset topic... Gary pushed those projects to his sandbox at that time

we also briefly discussed the idea of a Flux consumer on the offset topic where we could determine scaling based on windowed operations, but that's something to revisit once we get the simple case in place (can probably think about this at the partition level initially though)

also worth a look: https://github.com/linkedin/Burrow

ericbottard:
Yeah, when I reached out to Gary he showed me something, but that wasn't that at all (creating a `Process` that invoked `kafka-consumer-groups` and reading stdout! ).

I was gonna say that we shouldn't make this too smart while it's in java, because this would hinder our port to Go, but Burrow may be exactly what we need.

Will first rework the java version, hopefully in time for our 4:30 call

markfisher:
awesome!

garyrussell:
I have discovered that the lag is made available via JMX and also by the consumer `metrics()` API...

>tags={client-id=consumer-1}]=org.apache.kafka.common.metrics.KafkaMetric@3a407b47, MetricName [name=rjug-0.records-lag, group=consumer-fetch-manager-metrics, description=The latest lag of the partition,

average and max lag are also available

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -41,30 +51,123 @@
...
+									.max().getAsLong()
+						));
+				functions.values().parallelStream().forEach(
+						f -> {
+							String expression = f.getSpec().getScalingStrategy() != null ? f.getSpec().getScalingStrategy().getReplicasExpression() : "1";
+							Expression spel = spelExpressionParser.parseExpression(expression);
+							EvaluationContext context = new StandardEvaluationContext();
+							context.setVariable("lag", lags.get(f.getMetadata().getName()));
+							int replicas = spel.getValue(context, Integer.class);
+							// TODO no more replicas than topic.partitions
```
markfisher: I feel like we need this ceiling enforcement before merging. Otherwise, it's too easy to spin up excess instances.

build-model
```diff
@@ -3,6 +3,7 @@ set -e -v
 
 # regenerate the fabric8 java API including k8s custom resources
 pushd $(go env GOPATH)/src/github.com/fabric8io/kubernetes-model/
-git checkout sk8s
+git fetch origin
+git checkout origin/sk8s-sidecar-scaling
```
markfisher: can we go ahead and push the revised model so that this PR doesn't need to change the branch here from "sk8s"? (seems possible since the only change is the addition of "replicase" for Topic resources, and that is not a required property since it defaults to 1)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatcherConfiguration.java
```diff
@@ -16,35 +16,42 @@
...
+	public EventDispatchingHandler eventDispatchingHandler() {
+		return new EventDispatchingHandler();
 	}
 
 	@Bean
 	public ResourceEventPublisher functionsEventPublisher(Sk8sClient client) {
 		return new ResourceEventPublisher(client.functions());
 	}
 
+@Bean
```
markfisher: this and the following `@Bean` don't appear to be indented

ericbottard: Needed to keep track of current nb of replicas and be notified of nb partitions

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -16,55 +16,264 @@
...
+
+	// TODO: Change key to ObjectReference or similar
+	private final Map<String, Topic> topics = new ConcurrentHashMap<>();
 
-	@Autowired // TODO: merge in here?
-	private JobLauncher launcher;
+	private final Map<String, CountDownLatch> topicsReady = new ConcurrentHashMap<>();
 
-	private final KubernetesClient kubernetesClient;
+	private final Map<String, Integer> actualNbReplicas = new ConcurrentHashMap<>();
```
markfisher: `actualReplicaCount` would be a better name IMO

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatcherConfiguration.java
ericbottard: Indentation is off though

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -16,55 +16,264 @@
...
 
-	public EventDispatchingHandler(KubernetesClient kubernetesClient) {
-		this.kubernetesClient = kubernetesClient;
-	}
+	private final Map<String, Long> edgeInfos = new ConcurrentHashMap<>();
+
+	@Autowired
+	private FunctionDeployer deployer;
+
+	private volatile long scalingFrequencyMs = 0L; // Wait forever
```
markfisher: technically this should be "interval" not "frequency"

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -16,55 +16,264 @@
...
+	}
+
+	@PreDestroy
+	public synchronized void tearDown() {
+		running = false;
+		this.notify();
+		kafkaConsumerMonitor.close();
+	}
+
+	private void updateWakeUpInterval() {
```
markfisher: might as well pick one name consistently: wakeup or scaling? (I prefer the latter)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -16,55 +16,264 @@
...
+									e -> e.getKey().group,
+									e -> e.getValue().stream()
+											.mapToLong(KafkaConsumerMonitor.Offsets::getLag)
+											.max().getAsLong()));
+					functions.values().stream().forEach(
+							f -> {
+								int desired = computeDesiredNbReplicas(lags, f);
+								String name = f.getMetadata().getName();
+								Integer currentDesired = actualNbReplicas.computeIfAbsent(name, k -> desired);
+								System.out.println("Want " + desired + " for " + name);
```
markfisher: change to logging

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -16,55 +16,264 @@
...
+							.collect(Collectors.toMap(
+									e -> e.getKey().group,
+									e -> e.getValue().stream()
+											.mapToLong(KafkaConsumerMonitor.Offsets::getLag)
+											.max().getAsLong()));
+					functions.values().stream().forEach(
+							f -> {
+								int desired = computeDesiredNbReplicas(lags, f);
+								String name = f.getMetadata().getName();
+								Integer currentDesired = actualNbReplicas.computeIfAbsent(name, k -> desired);
```
markfisher: the assigned var should be `actual` right?

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -16,55 +16,264 @@
...
+									if (edgeInfos.computeIfAbsent(name,
+											n -> System.currentTimeMillis()) < System.currentTimeMillis()
+													- DOWN_EDGE_LINGER_PERIOD) {
+										deployer.deploy(f, desired);
+										actualNbReplicas.put(name, desired);
+										edgeInfos.remove(name);
+									}
+									else {
+										System.out.printf("Waiting another %dms before scaling down to %d for %s%n",
+												DOWN_EDGE_LINGER_PERIOD
```
markfisher: change to logging

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
ericbottard: That's the bug I found. if absent, should be 0 now

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -16,55 +16,264 @@
...
+											.mapToLong(KafkaConsumerMonitor.Offsets::getLag)
+											.max().getAsLong()));
+					functions.values().stream().forEach(
+							f -> {
+								int desired = computeDesiredNbReplicas(lags, f);
+								String name = f.getMetadata().getName();
+								Integer currentDesired = actualNbReplicas.computeIfAbsent(name, k -> desired);
+								System.out.println("Want " + desired + " for " + name);
+								if (desired < currentDesired) {
+									if (edgeInfos.computeIfAbsent(name,
```
markfisher: "edgeInfos" should be "downEdgeLingerPeriods" (or change that constant name to match something else, perhaps "scaleDownDelays"?)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -16,55 +16,264 @@
...
+					}
+					catch (InterruptedException e) {
+						Thread.currentThread().interrupt();
+					}
+				}
+			}
+		}
+
+		/**
+		 * Compute the desired nb of replicas for a function. Each function defines 4 values:
```
markfisher: the function doesn't yet define these 4 values (we aren't sure yet want we want to expose to users as configuration options)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -16,55 +16,264 @@
...
+		 * <ul>
+		 * <li>minReplicas (>= 0, default 0)</li>
+		 * <li>maxReplicas (minReplicas <= maxReplicas <= nbPartitions, default nbPartitions)</li>
+		 * <li>l, the amount of lag required to trigger the first pod to appear, default 1</li>
+		 * <li>L, the amount of lag required to trigger all (maxReplicas) pods to appear. Default
+		 * 10.</li>
```
markfisher: I think the same value as "maxReplicas" would be a sensible default for the big L.

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -16,55 +16,264 @@
...
+		 * </ul>
+		 * This method linearly interpolates based on witnessed lag and clamps the result between
+		 * min/maxReplicas.
+		 */
+		private int computeDesiredNbReplicas(Map<String, Long> lags, XFunction f) {
```
markfisher: rename: "computeDesiredReplicaCount"

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -16,55 +16,264 @@
...
+			String fnName = f.getMetadata().getName();
+			long lag = lags.get(fnName);
+
+			// TODO: those 3 numbers part of Function spec?
+			int bigL = 10;
```
markfisher: we should name these more descriptively IMO ("initialLag" and "maxLag"?... or something more accurate if necessary)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -16,55 +16,264 @@
...
+				return 1;
+			}
+			else {
+				return spec.getPartitions();
+			}
+		}
+
+		private void waitForTopicEntry(String input) {
+			try {
+				topicsReady.computeIfAbsent(input, i -> new CountDownLatch(1)).await();
```
markfisher: seems like we might want a bounded wait timeout here (also I'd drop "Entry" from the method name... possibly "waitForTopicIfNecessary"?)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -16,55 +16,264 @@
...
+			}
+			catch (InterruptedException e) {
+				Thread.currentThread().interrupt();
+			}
+		}
+
+		private void logOffsets(Map<KafkaConsumerMonitor.TopicGroup, List<KafkaConsumerMonitor.Offsets>> offsets) {
+			for (Map.Entry<KafkaConsumerMonitor.TopicGroup, List<KafkaConsumerMonitor.Offsets>> entry : offsets
+					.entrySet()) {
+				System.out.println(entry.getKey());
```
markfisher: change the 2 sysouts in this method to use logging

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/FunctionDeployer.java
```diff
@@ -17,80 +17,94 @@
...
 		this.kubernetesClient = kubernetesClient;
 	}
 
-	public void dispatch(XFunction functionResource) {
+	/**
+	 * Requests that the given function be deployed with N replicas.
+	 */
+	public void deploy(XFunction functionResource, int replicas) {
 		String functionName = functionResource.getMetadata().getName();
+		System.out.printf("Setting %d replicas for %s%n", replicas, functionName);
```
markfisher: change to logging

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
ericbottard: Not exactly the same. The thread wakes up at said interval, but does not mean that it scales (as in "changes the value") on each wakeup. Still want `updateScalingInterval`?

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/FunctionDeployer.java
```diff
@@ -144,22 +158,20 @@ private VolumeMount buildNamedPipesMount() {
...
+		Map<String, Object> config = new LinkedHashMap<>();
+		config.put("spring.cloud.stream.bindings.input.destination", function.getSpec().getInput());
+		config.put("spring.cloud.stream.bindings.output.destination", function.getSpec().getOutput());
+		config.put("spring.cloud.stream.bindings.input.group", function.getMetadata().getName());
+		config.put("spring.profiles.active", function.getSpec().getProtocol());
+		config.put("spring.application.name", "sidecar-" + function.getSpec().getInput());
+
+		config.put("spring.cloud.stream.kafka.binder.brokers", System.getenv("SPRING_CLOUD_STREAM_KAFKA_BINDER_BROKERS"));
+		config.put("spring.cloud.stream.kafka.binder.zkNodes", System.getenv("SPRING_CLOUD_STREAM_KAFKA_BINDER_ZK_NODES"));
+		config.put("server.port", "-1");
```
markfisher: we might as well use a specific port for the sidecar, as long as it doesn't conflict with the function container port (8080)... 7433 mapped to "side" on a telephone, but we could use 8081 or whatever

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/KafkaConsumerMonitor.java
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * @author Eric Bottard
+ */
+public class KafkaConsumerMonitor implements Closeable {
+
+	private final String host;
+
+	private final Map<String, Consumer<?, ?>> consumersByGroup = new HashMap<>();
+
+	private final String monitoringConsumerGroupID = "monitoring_consumer_" + UUID.randomUUID().toString();
```
markfisher: I think `monitor_<UUID>` would be sufficient (maybe prefix with underscore(s) to further distinguish it from any actual consumer group)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
```diff
@@ -16,55 +16,264 @@
...
 /**
+ * Listens for function registration/un-registration and periodically scales deployment of
+ * function pods based on the evaluation of a SpEL expression.
+ *
+ * @author Eric Bottard
  * @author Mark Fisher
  */
 public class EventDispatchingHandler {
 
+	public static final int DOWN_EDGE_LINGER_PERIOD = 10_000;
```
markfisher: this should be private

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
ericbottard: That's unrelated. And remember the lag is understood _per partition_ (see computation below, which uses `max` as the aggregation function, not `sum`)

```java
Map<String, Long> lags = offsets.entrySet().stream()
	.collect(Collectors.toMap(
		e -> e.getKey().group,
		e -> e.getValue().stream()
			.mapToLong(KafkaConsumerMonitor.Offsets::getLag)
								.max().getAsLong()));
```

I still think there is no good, business agnostic default value. Can you explain your reasoning (incl if you think we change max to sum above)?

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/KafkaConsumerMonitor.java
```diff
@@ -0,0 +1,193 @@
...
+
+		long getLag() {
+			return endOffset - currentOffset;
+		}
+
+		@Override
+		public String toString() {
+			return "Offsets{" +
+					"partition=" + partition +
+					", endOffset=" + endOffset +
```
markfisher: the "end" is not actually an offset

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
ericbottard: What if we timeout? A NPE or similar will happen elsewhere. Absence of entry should never happen (famous last words)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/KafkaConsumerMonitor.java
```diff
@@ -0,0 +1,193 @@
...
+
+	private final Map<String, Consumer<?, ?>> consumersByGroup = new HashMap<>();
+
+	private final String monitoringConsumerGroupID = "monitoring_consumer_" + UUID.randomUUID().toString();
+
+	private volatile Consumer<?,?> logEndOffsetTrackingConsumer;
+
+	private final Set<TopicGroup> tracked = new HashSet<>();
+
+	private final static Collection<TopicPartition> FOR_ALL_PARTITIONS = Collections.emptyList();
```
markfisher: could this just be "PARTITIONS"? I'm not following the meaning of "FOR_ALL_".

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/FunctionDeployer.java
ericbottard: I used -1 to disable entirely (should that be 0?).

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/KafkaConsumerMonitor.java
```diff
@@ -0,0 +1,193 @@
...
+ */
+public class KafkaConsumerMonitor implements Closeable {
+
+	private final String host;
+
+	private final Map<String, Consumer<?, ?>> consumersByGroup = new HashMap<>();
+
+	private final String monitoringConsumerGroupID = "monitoring_consumer_" + UUID.randomUUID().toString();
+
+	private volatile Consumer<?,?> logEndOffsetTrackingConsumer;
```
markfisher: this could just be "offsetTrackingConsumer" as it does more than "logEnd" right?

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/KafkaConsumerMonitor.java
```diff
@@ -0,0 +1,193 @@
...
+												}
+												long position = logEndOffsetTrackingConsumer.position(tp);
+												return new Offsets(tp.partition(), position,
+														committed.offset());
+											})
+									.collect(Collectors.toList());
+						}));
+	}
+
+	static class TopicGroup {
```
markfisher: this name is confusing since it sounds like *a group of Topics*... maybe this is just the `ConsumerGroup`? (or ConsumerKey)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
ericbottard: Assume we're using `sum`. Are you voting for `bigL = max` because if lag (summed) == 10 with 10 partitions, then there is a chance that each partition has 1 msg pending?

samples/shell/echo/echo.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: extensions.sk8s.io/v1
 kind: Topic
```
markfisher: should the Topic be removed from this yaml (and this be renamed `echo-function.yaml`) now that you've added `greetings-topic.yaml` on its own?

samples/shell/echo/echo.yaml
```diff
@@ -12,3 +14,6 @@ spec:
   image: sk8s/echo:v0001
   protocol: stdio
   input: greetings
+  env:
+  - name: FUNCTION_URI
+    value: /resources/script/echo.sh
```
markfisher: why was this added here?

the current Dockerfile for the echo sample is:
```
FROM sk8s/shell-function-invoker:0.0.1-SNAPSHOT
ADD echo.sh /functions/script.sh
ENV FUNCTION_URI /functions/script.sh
```

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/KafkaConsumerMonitor.java
ericbottard: Using an empty collection instructs the method where it is used to reply for all assigned partitions. Used a constant to make that explicit, apparently I failed...

markfisher: I see. If it's only used in one place, that seems unnecessary (maybe a comment where invoked with the empty collection instead?)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/KafkaConsumerMonitor.java
ericbottard: No, it's really used for that: to track the "end" offset for partitions it tracks (by forcing a seek to end).
Tracking the _position_ of other consumers is done by `consumersByGroup`

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/KafkaConsumerMonitor.java
ericbottard: It's really a dumb Tuple used as a data structure in the set of what is tracked.
If functions were supporting multiple inputs (which they could, as we discussed) this would evolve to 1 Group+N topics

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
markfisher: what I mean is that inside this method the actual update applies to a var called "scalingFrequencyMs". I was suggesting that the method should describe what it's updating more accurately. I would probably call it the "scalingMonitorInterval" (that way it's not suggesting that it will actually *scale* just *monitor*, and as I pointed out elsewhere it's technically an interval and not a frequency, which would be N-per-some-timeunit)

samples/shell/echo/echo.yaml
ericbottard: Interesting. Artifact of previous rebase I assume

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
markfisher: I was just suggesting that "currentDesired" is confusing. I think this whole method reads better if the 2 values being compared are simply called "actual" and "desired".

samples/shell/echo/echo.yaml
ericbottard: This is weird though, seems to indicate that the image ENV has precedence over what we set programmatically

samples/shell/echo/echo.yaml
ericbottard: Sure. Note that this doesn't fix the root problem of the topic-controller not adding partitions.

Created https://github.com/markfisher/sk8s/issues/85

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
markfisher: no, I understand that little l and big L are both being evaluated against the max lag seen on any one partition (aside: I actually think it would be smoother if you took the average and rounded up), but the value 10 seems arbitrary, whereas it could be proportional to the max number of replicas you would ever instantiate; i.e. the overall tolerance for lag depends on how much headroom is available for scaling. As an example, if I had 1000 partitions, wouldn't it be too aggressive to scale to 1000 instances just because I happened to see lag of 10 on the most-backed up partition during a single test at a 100ms interval?

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatchingHandler.java
markfisher: as long as you're confident this can't lead to deadlock

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/FunctionDeployer.java
markfisher: I see. You're right: -1 disables (0 would be random). I think we would use `/health` for a readiness probe though. Is there a reason you included this in this PR? If not, let's treat it as a separate concern. If we truly want to disable, then we should avoid using a `WebApplicationContext` in the first place (and find an alternative for a readiness probe).

ericbottard: Note that the go sidecar doesn't (yet) create an http server, and we don't set up probes. So wouldn't it be better to disable for now, and re-enable when everything works together?

ericbottard: And to answer your question, this was a first attempt at speeding up the java sidecar

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/KafkaConsumerMonitor.java
markfisher: how about "lagTrackingConsumer"?

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/KafkaConsumerMonitor.java
markfisher: yeah, my point is that it's not representing a group of Topics yet that's what it suggests

maybe `Subscription`?

markfisher: ...or `Bindings`? ;)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/FunctionDeployer.java
markfisher: ok, then - fine to disable for now. Let's capture an issue to either switch to the Go Sidecar or optimize the Java Sidecar. We'd want an issue for adding zipkin tracing to the Go Sidecar as well.

ericbottard: Created https://github.com/ericbottard/function-sidecar/issues/2 and https://github.com/ericbottard/function-sidecar/issues/3

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/KafkaConsumerMonitor.java
markfisher: actually I now see where they use the term "log-end-offset"... think I see where you're coming from now (also on the name of the consumer)

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatcherProperties.java
```diff
@@ -24,18 +24,8 @@
 @ConfigurationProperties(prefix = "sk8s.event.dispatcher")
 public class EventDispatcherProperties {
 
-	private String namespace = "default";
```
markfisher: I believe @freynca is depending on the namespace being configurable.

ericbottard:
The getter was never called

ericbottard:
The déployer creates ressources in the same NS as the function

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatcherProperties.java
freynca: The namespace should absolutely be configurable

ericbottard:
The NS of what ? If you create a FN in the foo NS (using kubectl or our
cli), then it will create its pods in the foo NS

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatcherProperties.java
markfisher: this is what @ericbottard changed (namespace of the event-dispatcher is inherited by all of its deployments):
https://github.com/markfisher/sk8s/pull/41/files#diff-02a72fcc3a08713e28a5ab7f8d510f71R76

markfisher:
squashed and merged as https://github.com/markfisher/sk8s/commit/cf1163a5fe6a0b0e4d00690c14ec01ac1987a385

PR #45, opened 2017-10-14
Add Zipkin/Sleuth support for TopicGateway and Sidecars

trisberg:
- adding Zipkin server that reads traces from Kafka
- prepare topic-gateway for sleuth/zipkin tracing
  - update topic-gateway to use stream-binder for kafka
  - add explicit sleuth tracing for topics used by message publisher
- make sidecars participate in message tracing
- limit memory allocated for Java apps


markfisher:
merged as https://github.com/markfisher/sk8s/commit/3f537b298997d6b8808dca8895626a76b86eb7cc

PR #48, opened 2017-10-16
[DO NOT MERGE] Publish sample functions using http server

trisberg:
- eliminates the need for mounting samples folder

- all function launchers must support http resources

### DIFF
--- a/samples/Dockerfile
+++ b/samples/Dockerfile
@@ -1,9 +1,9 @@
 FROM python:3.6-slim
 VOLUME /tmp
 ADD ./java/greeter/target/greeter-1.0.0.jar /functions/java/greeter-1.0.0.jar
-ADD ./node/hello.js /functions/node/hello.js
+ADD ./node/square.js /functions/node/square.js
 ADD ./script/wordcount.sh /functions/script/wordcount.sh
 RUN sh -c 'touch /functions/java/greeter-1.0.0.jar'
-RUN sh -c 'touch /functions/node/hello.js'
+RUN sh -c 'touch /functions/node/square.js'
 RUN sh -c 'touch /functions/script/wordcount.sh'
 CMD cd functions && python -m http.server 8080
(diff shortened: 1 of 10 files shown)


trisberg:
Rebased and added suport for downloading node js module using the function URI

samples/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.6-slim
+VOLUME /tmp
+ADD ./java/greeter/target/greeter-1.0.0.jar /functions/java/greeter-1.0.0.jar
```
markfisher: We should probably restructure our samples so that this will be future-proof (can do that as a separate PR after this one is merged for now). For example, any new samples' code should be included in this. Likewise, all the yamls should be in one place (not necessarily the root "samples" dir).

trisberg: although, samples deploy/svc is separate from the core faas implementation and would belong outside of the config directory

samples/wordcount-function.yaml
```diff
@@ -9,4 +9,4 @@ spec:
   output: greetings
   env:
   - name: FUNCTION_URI
-    value: /resources/script/wordcount.sh
+    value: http://sample-functions/script/wordcount.sh
```
markfisher: does this work?

markfisher:
I'm concerned that this pushes a pull model onto all of the invokers (e.g. using wget in `server.js`) whereas before only the Java version was resolving the resource (and even in that case, it could have been a local file). We are still going to want to support a fully-baked container approach for those cases where dynamic loading of user-code is not necessary (i.e. when no warm start/pooling is involved), and in those cases having a file written in the top layer of a container image would be closer to the mounted volume approach.

samples/wordcount-function.yaml
trisberg: not sure, probably not, didn't have instructions for how to run this sample

trisberg:
Closed in favor of https://github.com/markfisher/sk8s/pull/51

PR #51, opened 2017-10-18
Build samples for Java and Node as Docker images

trisberg:
- adding a faas command line to create the images

- adding Dockerfile for java and node sample functions

- changing the function yaml to reference the new images

- removing sample docker images in teardown

- removing \resources mount in JobLauncher

- updating README

### DIFF
--- a/samples/greeter-function.yaml
+++ b/samples/greeter-function.yaml
@@ -3,10 +3,10 @@ kind: Function
 metadata:
   name: greeter
 spec:
-  image: sk8s/java-function-invoker:v0001
+  image: sk8s/greeter:v0001
   protocol: http
   input: names
   output: greetings
   env:
   - name: FUNCTION_URI
-    value: file:///resources/java/greeter/target/greeter-1.0.0.jar?functions.Greeter
+    value: file:///functions/java/greeter-1.0.0.jar?functions.Greeter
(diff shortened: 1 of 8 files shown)


samples/node/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM sk8s/node-function-invoker:latest
+ADD square.js /functions/node/square.js
```
markfisher: it seems like we could add this to a default location/name such as `/function.js` so that we don't need the `FUNCTION_URI` when that default is used

samples/greeter-function.yaml
```diff
@@ -3,10 +3,10 @@ kind: Function
...
 spec:
-  image: sk8s/java-function-invoker:v0001
+  image: sk8s/greeter:v0001
   protocol: http
   input: names
   output: greetings
   env:
   - name: FUNCTION_URI
-    value: file:///resources/java/greeter/target/greeter-1.0.0.jar?functions.Greeter
+    value: file:///functions/java/greeter-1.0.0.jar?functions.Greeter
```
markfisher: defaults would be nice here as well, but it's a bit trickier given that the function name might need to be specified... however, in the case where there's a single JAR and a single Function implementation in that JAR, the `java-function-invoker` could probably introspect the single JAR dropped into `/function`.

markfisher:
rebased and merged as https://github.com/markfisher/sk8s/commit/e8a4d188a313dcf271d5e6330076d6394e53dc53

PR #52, opened 2017-10-18
Add python-function-invoker

dturanski:
This adds a Python 2 function invoker. The image adds from [anaconda](https://hub.docker.com/r/continuumio/anaconda) which is based Python 2 using stdio.  There presumably would be a parallel one for Python 3. 

The sample is an ML tweet sentiment analyzer that imports a number of data analysis and ML libraries which are included with Anaconda and requires Python 2.  After several unsuccessful attempts to use an approach similar to [AWS Lambda for Python ](http://docs.aws.amazon.com/lambda/latest/dg/lambda-python-how-to-create-deployment-package.html) for this app, I decided to save that for a rainy day.

AWS Lambda requires the developer to create a deployment zip file with all dependencies installed in the root directory. This function invoker supports this, or tries to at least, if you set `FUNCTION_URI` to a zip file.  This works standalone but YMMV with docker.  I haven't yet tested with a simpler example.  Using Anaconda , the function pod actually loads faster since there is minimal initialization required. 

A characteristic of both a full Anaconda image and having the AWS approach is that neither require the installation of Python packages at runtime, as would be the case with creating a virtual environment or performing pip/conda installs. I tried creating a virtual environment at runtime as specified by the function. It takes too long to initialize for FaaS. Anaconda has a 3GB image but seems to spin up quickly.

Also, the URL requires `handler` as a query parameter. Based on previous discussions, I thought I would give it a try until we decide on the final format.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -159,6 +159,33 @@ kubectl create -f samples/square-function.yaml
 ./publish numbers 2
 ```
 
+
+== Add a Python Function
+
+=== Have the Samples Folder Mounted on the Minikube VM
+(see above if you haven't done this yet)
+
+=== Create a New Topic
+
+```
+kubectl create -f samples/tweets.yaml
+```
+
+=== Create the Function
+
+```
+kubectl create -f samples/sentiments.yaml
+```
+
+=== Publish a Message
+
+The function performs sentiment analysis on tweets. It accepts JSON and looks only at the `text` field. The input
+is in the form of an array (sorry about the escaped quotes).
+
+```
+./publish tweets "[{\"text\":\"happy happy happy\"},{\"text\":\"sad sad sad\"}]"
+```
+
 == Tear it all down
 
 ```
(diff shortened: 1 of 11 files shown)


markfisher:
rebased and merged as https://github.com/markfisher/sk8s/commit/ce06cadb41232cc0a10c6bb00e57b20759a0cd9c

...along with these changes to build a docker container rather than require the mount (which was removed yesterday via https://github.com/markfisher/sk8s/commit/e8a4d188a313dcf271d5e6330076d6394e53dc53): https://github.com/markfisher/sk8s/commit/c3b14397b254f00f1b46294238ab98bc2b6ab7de

the sentiments function is *not* currently working, since it is failing to find the file (even though it's in the container). Only `/resources` changed to `/functions` as far as that code is concerned, so I think it's an issue in the runner. @ericbottard is going to take a look

PR #53, opened 2017-10-19
only create volume for stdio protocol

markfisher:


markfisher:
rebased and merged as https://github.com/markfisher/sk8s/commit/981a61d02c705a4b924875876b06d1ab898d3eb7

PR #54, opened 2017-10-19
Fix shell functions

ericbottard:


README.adoc
```diff
@@ -54,6 +54,11 @@ kubectl apply -f config
 ```
 kubectl create -f samples/greetings-topic.yaml
 ```
+=== Build the Docker container for the function code
+
+```
+./faas create -f scripts -v v0001 -p samples/script/
```
markfisher: this should be `-f echo`

ericbottard: but then needs to create a second image for wordcount

markfisher: yes, that's exactly what we want for the pre-baked option ... we will revisit the dynamic option (where the base image is sufficient)

samples/script/Dockerfile
```diff
@@ -0,0 +1,2 @@
+FROM sk8s/shell-function-invoker:latest
+ADD *.sh /functions/script/
```
markfisher: this should also just be explicit about `wordcount.sh` for now. The goal as a follow-up will be to have a default file location/name so that the `FUNCTION_URI` env var is not needed for the pre-built immutable container cases.

README.adoc
markfisher: I will change while merging.

markfisher:
rebased and merged as https://github.com/markfisher/sk8s/commit/914617acfe41cba2fa28859b6f8fb04264f62ad2

...along with changes here: https://github.com/markfisher/sk8s/commit/f511d3955c154af8446e9a7ca2c84f688085db55

PR #56, opened 2017-10-19
Enhancing faas cli

trisberg:
- adding build and create commands

- aligning options with kubectl
  using -f for path/filename and -n for function name

- wrapping kubectl create as ./faas create

- updating README using the new commands

### DIFF
--- a/faas
+++ b/faas
@@ -3,23 +3,32 @@
 function print_usage() {
 cat <<EOF
 
-USAGE: faas create -f <function> -p <path> -v <version>
-  This command will build the function based on the code available in the path directory
+USAGE: 
 
-Flags:
+faas build -n <name> -v <version> -f <path>
+  This command will build the function based on the code available in the path directory, using the name 
+  specified for the image that is built
 
-[*] -f  | --function - the name of the function
-    -p  | --path - the path to the function code (defaults to current directory)
-    -v  | --version - the version of the function created (defaults to v0001)
+faas create -f <path>
+  This command will create the resource[s] based on the resource definition[s] that the path points to
 
-[*] = Required arguments
+Options:
+
+  -n  | --name - the name of the function (defaults to nake of current directory)
+  -v  | --version - the version of the function (defaults to v0001)
+  -f  | --filename - filename, directory, or URL for the code or resource (defaults to current directory)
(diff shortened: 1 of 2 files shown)


faas
```diff
@@ -3,25 +3,34 @@
...
-[*] -f  | --function - the name of the function
-    -p  | --path - the path to the function code (defaults to current directory)
-    -v  | --version - the version of the function created (defaults to v0001)
+faas create -f <path>
+  This command will create the resource[s] based on the resource definition[s] that the path points to
 
-[*] = Required arguments
+Options:
+
+  -n  | --name - the name of the function (defaults to nake of current directory)
```
markfisher: typo: make

faas
```diff
@@ -3,25 +3,34 @@
...
+  specified for the image that is built
 
-[*] -f  | --function - the name of the function
-    -p  | --path - the path to the function code (defaults to current directory)
-    -v  | --version - the version of the function created (defaults to v0001)
+faas create -f <path>
+  This command will create the resource[s] based on the resource definition[s] that the path points to
 
-[*] = Required arguments
+Options:
```
markfisher: shouldn't `-n` and `-v` only be supported when using `faas build`?

it seems like we need separate usage... the top level should just show sub-commands, and then for specific usage, we'd support something like `faas <command> -h`

trisberg: Yeah, we could do that. Let me add a commit

markfisher:
should we add `apply` in addition to `create`? I'm thinking that might be useful for topics since multiple functions are sharing them

markfisher:
I'd also like to add log-tailing here as `faas logs -n function` (can simplify that part of the README significantly). I've been using a script for that locally. In addition to the function name, we'd need an option for choosing sidecar vs. main container. My script takes `-c sidecar` (default is `main`) but really it's just a binary choice given our model, so it could perhaps be simpler than that.

markfisher:
rebased and merged as https://github.com/markfisher/sk8s/commit/a7a4797510c3707a0c48d26e35c6208125487c10
...along with these changes: https://github.com/markfisher/sk8s/commit/6a7745baad003770a162ca67a0e7dda8f9492759

PR #60, opened 2017-10-20
Change delete command to delete function and jobs (for now) resources

trisberg:
- jobs should probably be removed by controller at some point

- adding delete instructions to READMEs

- changing README links to be relative (helps when viewing locally)

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -49,7 +49,7 @@ kubectl apply -f config/kafka
 kubectl apply -f config
 ```
 
-== Try Some https://github.com/markfisher/sk8s/tree/master/samples[Samples]
+== Try Some link:samples/README.adoc[Samples]
 
 == Tear it all down
 
--- a/faas
+++ b/faas
@@ -73,15 +73,15 @@ EOF
 function print_delete_usage() {
 cat <<EOF
 
-Delete the resource definition[s] that the path points to.
+Delete the resource definition[s] for the function specified.
 
 Usage: 
 
-  faas delete -f <path>
+  faas delete -n <name>
 
 Options:
 
-  -f, --filename: filename, directory, or URL for the resource definition[s] (defaults to the current directory)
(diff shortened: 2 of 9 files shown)


markfisher:
merged with amended comment (for 50 char limit) as https://github.com/markfisher/sk8s/commit/6d5636f230e7365d8451d8645c665c472a0543ea

markfisher:
apparently setting the owner reference on the Job resource (referencing the Function resource) is not sufficient for cascading garbage collection. I added that here (but it doesn't work): https://github.com/markfisher/sk8s/commit/35338b1ba0e7a3e5b9f4e151186bf1c0a0cb97a9

so, I assume we need to set the `deleteOptions.propagationPolicy` on the owning resource (Function in our case) as described here: https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/

many resources (such as `Deployment` resources) have that set by default

cc @ericbottard

markfisher:
I realized the owner reference should be in the Job metadata (rather than its Pod template), so I changed that here (but it still doesn't cascade by default): https://github.com/markfisher/sk8s/commit/bab3da03123e252f2077c171282bee0c1363c293

trisberg:
Are you on k8s v1.8? I think it's a new feature. My minikube is v1.7.5.

markfisher:
I went ahead and pushed this for now (the controller deletes the jobs based on the function label), and the owner-reference is being set on the Job metadata... so maybe be that will provide an option for cascading deletes at some point: https://github.com/markfisher/sk8s/commit/703ff6832513e30b50ec87054e89a1291b7a1d54

ericbottard:
Yes we need 1.8+ for this to happen magically (haven't managed to run a 1.8+ minikube yet, hitting RBAC issues). Still doesn't hurt and serves "documentation" purposes nevertheless.

PR #63, opened 2017-10-20
Helm new version

trisberg:
Install with 

```
helm repo add sk8s-incubator http://sk8s-incubator.cfapps.io
helm repo update
helm install --name demo sk8s-incubator/sk8s
```


charts/sk8s/values.yaml
```diff
@@ -5,25 +5,36 @@ eventDispatcher:
   replicaCount: 1
   image:
     repository: sk8s/event-dispatcher
-    tag: v0001
+    tag: v0002
```
markfisher: did you intend to change these to v0002 or were you testing that locally?

trisberg: I did, I pushed v0002 to DockerHub, we could move the whole build to v0002 at some point

markfisher: okay - I was thinking that the `docker.tag` property is still at `v0001`, but I guess that's not relevant in this case, because the helm install will always be pulling from DockerHub... or would it ever be pulling locally if using minikube and the images are already local?

trisberg: it does pull from local, so if we set the new version in the pom we also need to update the v0001 used in other places

markfisher: we might want to decouple the build of the invoker images (currently triggered by the overall maven build), since with the helm install, all that should be necessary is to have those invoker images also available (we will eventually start publishing those to DockerHub also), and then the samples can be run

markfisher:
worked great!

rebased, squashed, and merged as https://github.com/markfisher/sk8s/commit/3db9e7ac70a8ab771095884ce638fadfb3723ea1

PR #66, opened 2017-10-23
Changing to use Spotify's dockerfile Maven plugin

trisberg:
- switching to use pom version as the only tag for images

- removing assembly.xml and adding Dockerfile instead


trisberg:
I noticed that some docker images are build during package while other during docker[file]:build. Should we change that to build all images in the same goal/phase?

markfisher:
The invokers are currently built within the Maven build. I agree we should be consistent. Should we keep the docker build separate for all or integrate into the Maven build for all?

PR #68, opened 2017-10-24
allow k8s to generate random nodePort

pivotal-jamie-klassen:
otherwise the port may be already allocated in case of multiple deployments, as seen here: https://ci.faas.to.cf-app.com/teams/pfs/pipelines/faas_integration/jobs/build-sk8s-and-invoker-java/builds/25


trisberg:
LGTM

PR #69, opened 2017-10-24
Switching all docker tags to latest

pivotal-jamie-klassen:
https://github.com/pivotal-cf/pfs-ci/issues/10

### DIFF
--- a/charts/sk8s/values.yaml
+++ b/charts/sk8s/values.yaml
@@ -28,7 +28,7 @@ zipkin:
   replicaCount: 1
   image:
     repository: sk8s/zipkin-server
-    tag: v0001
+    tag: latest
     pullPolicy: IfNotPresent
   service:
     name: zipkin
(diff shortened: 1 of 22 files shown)


trisberg:
Using `latest` is a pain when developing with minikube since k8s can't dor the `ifNotPresent` check and ends up always downloading the latest from DocherHub. What is the actual issue we are solving here?

markfisher:
we're using `0.0.1-SNAPSHOT` now in all the docker build configs (in the POMs), and the local dev experience is better with a version other than "latest" because you can replace locally without always pulling

freynca:
@trisberg We're not particular about `latest` vs anything else. However, CI needs to tag these images, and there is no semver-based `sk8s` release process at this moment. Can use a hardcoded `0.0.1-SNAPSHOT` for now.

trisberg:
Is this for the sk8s artifacts or for the samples? Seems to be a mix in the PR.

PR #71, opened 2017-10-25
Add docker image prefix to all artifacts

pivotal-jamie-klassen:
to ensure that images can be identified as associated with sk8s regardless of their org.


trisberg:
LGTM

PR #73, opened 2017-10-25
Adding liveness/readiness probes

### LINKED ISSUE #64: Add health checks to the Helm charts
This will make it easier to demo since we can see when components are ready

trisberg:
Resolves #64


markfisher:
LGTM

PR #79, opened 2017-10-26
Externalizing sidecar image name

freynca:
Externalizing sidecar image name to a configurable property.

### DIFF
--- a/event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatcherProperties.java
+++ b/event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatcherProperties.java
@@ -25,6 +25,7 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 public class EventDispatcherProperties {
 
 	private String namespace = "default";
+	private String sidecarImageName = "sk8s/function-sidecar";
 
 	public String getNamespace() {
 		return namespace;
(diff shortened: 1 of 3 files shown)


event-dispatcher/src/main/resources/application.properties
```diff
@@ -0,0 +1 @@
+sk8s.event.dispatcher.sidecarImageName=sk8s/function-sidecar:latest
```
markfisher: I don't think we want `latest` here as that would pull every time.

freynca:
Changed default to `stable`. The default will be overridden by CI to match the appropriate version of the image.

markfisher:
we need to keep the default as `0.0.1-SNAPSHOT`

event-dispatcher/src/main/java/io/sk8s/event/dispatcher/EventDispatcherProperties.java
```diff
@@ -25,6 +25,7 @@
 public class EventDispatcherProperties {
 
 	private String namespace = "default";
+	private String sidecarImageName = "sk8s/function-sidecar";
```
markfisher: we could leave this out if providing via `application.properties` (or if we're going to have a default here, then we don't need `application.properties` anyways, but this should also include the tag)

freynca:
Closing this PR; will re-open a new one

PR #86, opened 2017-10-27
Externalize version number into a property

freynca:
Externalize version number into a property


trisberg:
Not a fan of this approach, makes the pom look unusual with a property for the version, might cause problems downstream. There are more standard ways for changing the version on releases like - http://maven.apache.org/maven-release/maven-release-plugin/examples/update-versions.html

trisberg:
The release version references that are used in other places should be externalized so they can be generated by the build

freynca:
Closing. Will switch all versioning to be driven from the Concourse `semver` resource.

PR #89, opened 2017-10-29
Move log tailing after publish

trisberg:
- we can no longer tail the logs before the message is published since we scale to 0


markfisher:
Now that the function pods are ephemeral, I wonder if we should also change the publish examples to use the new support for additional args to specify a number of messages and a sleep interval. For example the following will publish 10 messages with 5 seconds delay between each publish:

```./publish greetings hello 10 5```

PR #105, opened 2017-11-08
Adding retries to http dispatcher

freynca:
Adding retries to http dispatcher

Signed-off-by: Jamie Klassen <cklassen@pivotal.io>

### DIFF
--- a/function-sidecar/src/main/java/io/sk8s/sidecar/HttpDispatcher.java
+++ b/function-sidecar/src/main/java/io/sk8s/sidecar/HttpDispatcher.java
@@ -16,19 +16,37 @@
 
 package io.sk8s.sidecar;
 
+import org.springframework.retry.backoff.ExponentialBackOffPolicy;
+import org.springframework.retry.policy.SimpleRetryPolicy;
+import org.springframework.retry.support.RetryTemplate;
 import org.springframework.web.client.RestTemplate;
+
+import javax.annotation.PostConstruct;
 
 /**
  * @author Mark Fisher
  */
 public class HttpDispatcher implements Dispatcher {
 
-	private static final String INVOKER_URL = "http://localhost:8080/";
+    private static final String INVOKER_URL = "http://localhost:8080/";
 
-	private final RestTemplate restTemplate = new RestTemplate();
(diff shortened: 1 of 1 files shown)


function-sidecar/src/main/java/io/sk8s/sidecar/HttpDispatcher.java
```diff
@@ -16,19 +16,37 @@
...
  * @author Mark Fisher
  */
 public class HttpDispatcher implements Dispatcher {
 
-	private static final String INVOKER_URL = "http://localhost:8080/";
+    private static final String INVOKER_URL = "http://localhost:8080/";
+
+    private final RestTemplate restTemplate = new RestTemplate();
+    private final RetryTemplate retryTemplate = new RetryTemplate();
+    private final ExponentialBackOffPolicy backOffPolicy = new ExponentialBackOffPolicy();
```
markfisher: What's the rationale for using exponential backoff here? If the purpose is to handle the case where a sidecar is available before its corresponding Function Container, we want to reach that Container as soon as possible (and we aren't expecting a situation where we ultimately have to give up). Also, the Sidecar has nothing else to do until that Container arrives, so it might as well be aggressive in the retry loop.

markfisher:
please use tabs instead of spaces

function-sidecar/src/main/java/io/sk8s/sidecar/HttpDispatcher.java
```diff
@@ -16,19 +16,37 @@
...
+    private final SimpleRetryPolicy retryPolicy = new SimpleRetryPolicy();
+
+    @Override
+    public String dispatch(String input) {
+        return this.retryTemplate.execute(retryContext -> restTemplate.postForObject(INVOKER_URL, input, String.class));
+    }
 
-	private final RestTemplate restTemplate = new RestTemplate();
+    @PostConstruct
```
markfisher: all of this could be handled in a default constructor for now, and we probably don't need instance vars for anything other than the `retryTemplate` (if we decide to expose these for configurability, we'd probably refactor to have each as a `@Bean`)

markfisher:
please don't combine unrelated changes into a single PR

PR #108, opened 2017-11-09
Fix payload types handling

ericbottard:


topic-gateway/src/main/resources/application.properties
```diff
@@ -1,4 +1,5 @@
 spring.cloud.stream.bindings.input.destination = replies
 spring.cloud.stream.bindings.input.group = gateway
 spring.cloud.stream.bindings.input.consumer.header-mode = embeddedheaders
-spring.sleuth.sampler.percentage = 1.0
\ No newline at end of file
+spring.sleuth.sampler.percentage = 1.0
+logging.level.io.sk8s.topic.gateway = debug
```
markfisher: do we want to keep this in here? (fine with me for now if you think it's helpful and not too noisy, but I assumed you might have left it inadvertently)

ericbottard: Left it on purpose, but either way is fine really

markfisher:
LGTM; incorporated into https://github.com/markfisher/sk8s/pull/109

PR #109, opened 2017-11-09
Added support for sync request/reply

markfisher:
- using embedded headers mode until we can support Kafka 0.11 in the sidecar
- handling any message payload type for replies


markfisher:
rebased and merged as https://github.com/markfisher/sk8s/commit/bc7906ff74fccfd563230182c7315629287c1bd8

PR #111, opened 2017-11-10
isolate topic-gateway build

markfisher:


markfisher:
Closing this as we are no longer using the Java `topic-gateway`. It has been replaced by the Go `http-gateway`: https://github.com/ericbottard/sk8s-http-gateway

PR #115, opened 2017-11-10
bake helm charts

pivotal-jamie-klassen:
the `generate_chart.sh` script is somewhat self-documenting. It puts a tarball in the `charts` folder.

Signed-off-by: Felix Reyn <freyn@pivotal.io>


markfisher:
@trisberg considering the helm charts we publish are themselves versioned, shouldn't those be locked down to specific versions of the images of our components? It doesn't seem like a good idea to allow end-users to override the versions of those images. They should be upgrading the helm chart itself as a unit. That said, I know you are more familiar with the Helm chart idioms. Are things like image versions typically exposed as configurable values?

markfisher:
I agree we should maintain the configurable `serviceType` values. That seems to be a common practice.

pivotal-jamie-klassen:
@trisberg You definitely cannot just run `helm install` referring to the chart directory anymore, it's true. The idea is that you _must_ run the `generate_chart.sh` script and then use helm install referring to the generated tarball, which will have the image tags baked in and not configurable by end users at install-time. It sounds like your concern is for the local experience of folks developing the sk8s platform itself, and how easy it is for them to make changes to the chart/install it locally and get feedback that way. Do you think it's too much cognitive load or is too anti-idiomatic to make them do two steps:
```
SIDECAR_IMAGE_TAG=<tag> HTTP_GATEWAY_IMAGE_TAG=<tag> SK8S_IMAGE_TAG=<tag> charts/generate_chart.sh <version>
helm install --name test charts/sk8s-<version>.tgz --debug --dry-run
```
The other option is the generate_chart script could expand to cover other "local development" workflows, amounting to a kind of domain-specific CLI.

I think it's OK to allow the serviceTypes to be configurable at install-time, as k8s seems able to handle services communicating in-cluster regardless of the configuration option, but I don't understand why it would be required, beyond it being common practice. If the idea is to allow something like the http-gateway to be configured to be NodePort instead of LoadBalancer because of some minikube-specific concern, that sounds to me like a workaround that intentionally introduces inconsistency between different k8s substrates (ideally the chart should be robust and the operator shouldn't have to know the logic for which service types to choose?)

trisberg:
First of all, @markfisher is right, we should version the charts and the default Sk8s component versions should be in the `values.yaml`. Once we bump any of the versions, change the configuration etc. we should add a new chart version with these changes.

Even with this approach, we should provide flexibility to override the image attributes. Someone might want to use a custom built image from their own repo. 

If you look at all the charts in https://github.com/kubernetes/charts/blob/master/stable every single one I have looked at allow for image options to be overridden including repository, tag and imagePullPolicy.

For some examples see:
https://github.com/kubernetes/charts/blob/master/stable/mysql/values.yaml 
https://github.com/kubernetes/charts/blob/master/stable/concourse/values.yaml 

For the script that generates the chart, IMO that's the wrong approach. Helm is used to template configuration files and to add another templating layer on top of it just seems wrong and is not idiomatic use of Helm. All Helm charts that are part of KubeApps(https://kubeapps.com/)  are usable directly out of the GitHub repo and packaging them up and publishing them in a repo is simply providing a convenient way for people to access them. I think our goal should be to make our Helm chart available as part of KubeApps once we are ready.

If we need additional ways of generating configuration artifacts maybe we should look at some other tool. Maybe Jsonnet is a better fit, haven't tried it but it might be worth a look.

Lastly, LoadBalancer vs NodePort is something that are always a consideration that vary from deployment to deployment and we need to let the user decide how they want to access the exposed endpoints. If someone use an Ingress controller then I think they would want to use NodePort.

markfisher:
The example in https://docs.helm.sh/developing_charts/#template-files shows these 3 configurable values:

```
imageRegistry: The source registry for the Docker image.
dockerTag: The tag for the docker image.
pullPolicy: The Kubernetes pull policy.
```

BUT, notably that does not include the image *name*.

trisberg:
So if you do a `helm create` you get the following:

```
image:
  repository: nginx
  tag: stable
  pullPolicy: IfNotPresent
```

where `image.repository` is the name of the repository wich in this example is just the image name since this is an "official" image, other charts have `account/image` as the the name, you can also provide a fully qualified name if you use a non-DockerHub registry. I fond the docker terminology a bit confusing.

pivotal-jamie-klassen:
https://github.com/pivotal-cf/pfs-ci/commit/860af232a7828a26e334ffa41c06824095d0e8dc just modifies the values.yaml by replacing occurrences of '0.0.1-SNAPSHOT' with the current sk8s version. Otherwise the above discussion suggests that everything else should be configurable, so no changes to the sk8s repo are required.

PR #116, opened 2017-11-11
Use spring-cloud-function for HTTP handling

dsyer:
The functions are all still in a separate class loader, but no need for a
separate application context (whose lifecycle was never managed properly).
Composition also happens natively in s-c-f so there isn't a lot of code
here now, just enough to load the function classes and register them.


markfisher:
One thing that wasn't very obvious from the existing state is that we were maintaining a separation between the base and the function for reintroducing the dynamic loading of a Function into a warm app. Originally, there was an `/init` endpoint that would receive the request to load a Function (the request included the JAR URI and classname of the Function). We moved that logic into the `afterPropertiesSet()` of the Controller since currently we are not dynamically loading Functions. We will be adding dynamic behavior back, but maybe we have some simpler options that could build upon the changes in this PR. We could use a proxy (refreshing the target on demand) or possibly a lazy-initialization of the Function based on the first call to `/` (invoking the Function).

dsyer:
Right. I guess that sort of explains the apparently superfluous code in master. IMO we should be able to add that back in later. However, there are quite likely to be multiple ways of doing it with different trade offs, so I wouldn’t want to commit to any particular implementation yet.

markfisher:
I agree. As we revisit the dynamic loading support, it will be better to start from scratch and consider a range of options.

function-invokers/java-function-invoker/src/main/java/io/sk8s/invoker/java/server/FunctionConfiguration.java
```diff
@@ -0,0 +1,141 @@
...
+		return new DelegatingResourceLoader(loaders);
+	}
+
+	/**
+	 * Registers a singleton bean for each of the function classes passed into the
+	 * {@link FunctionProperties}. They are named sequentially "function0", "function1",
+	 * etc. The instances are created in an isolated class loader, so the jar they are
+	 * packed in has to define all the dependencies (except core JDK).
+	 * 
+	 * TODO: Add reactor and maybe Message (not spring-messaging) to class loader in a way
```
markfisher: I just hit this same requirement as I'm working on a `Flux`-based Function that I intend to use in a sample. Do you already have some thoughts on sharing that dependency?

function-invokers/java-function-invoker/src/test/java/io/sk8s/invoker/java/function/FunctionConfigurationTests.java
```diff
@@ -31,52 +34,61 @@
...
 					"io.sk8s.invoker.java.function.FunctionConfigurationTests.NumberEmitter," +
 					"io.sk8s.invoker.java.function.FunctionConfigurationTests.Frenchizer"
 	})
 	public static class SupplierCompositionTests extends FunctionConfigurationTests {
 
 		@Autowired
-		@Qualifier("function")
-		private Supplier<String> function;
+		@Qualifier("function0")
+		private Supplier<Integer> function;
```
ericbottard: It seems the composition order is backwards now, or am I missing something?

function-invokers/java-function-invoker/src/test/java/io/sk8s/invoker/java/function/FunctionConfigurationTests.java
```diff
@@ -31,52 +34,61 @@
...
 	public static class FunctionCompositionTests extends FunctionConfigurationTests {
 
 		@Autowired
-		@Qualifier("function")
+		@Qualifier("function1")
 		private Function<Integer, String> function;
 
 		@Test
 		public void testComposition() {
-			assertThat(function.apply(2), is("quatre"));
```
ericbottard: Related to my comment above I guess. This definitely does not test the same thing as before.
I guess composition is only supported while hitting this via http, so most of this test code can be removed?

function-invokers/java-function-invoker/src/main/java/io/sk8s/invoker/java/server/FunctionConfiguration.java
dsyer: I know how to do it with a more complicated class loader structure (c.f. spring-cloud-function-deployer). I think we don't have to solve that problem right here, right now, though?

function-invokers/java-function-invoker/src/test/java/io/sk8s/invoker/java/function/FunctionConfigurationTests.java
dsyer: I don't know. Maybe I didn't understand the old code. But "function0" is the first function on the list you passed in (hence `NumberEmitter`). Doesn't that make sense?

function-invokers/java-function-invoker/src/test/java/io/sk8s/invoker/java/function/FunctionConfigurationTests.java
dsyer: Agree that this is not testing the same thing. You could pull a composed function out of the `FunctionCatalog` (which isn't even in the context of these tests) and test the same thing that way, but then you'd only be testing features of Spring Cloud Function. I just wanted a sanity check that the functions had been created bu `FunctionConfiguration`.

function-invokers/java-function-invoker/src/main/java/io/sk8s/invoker/java/server/FunctionConfiguration.java
markfisher: I am building a demo for S1P where the "user" Function has `Flux` for input and output, so we do need to solve at least that part. I am able to make it work from this branch by passing the parent classloader (where you explicitly pass `null`). I know that crude approach is not desirable, but maybe for now we could just have a parent classloader that only shares `reactor-core` and `reactivestreams`. WDYT?

markfisher: Just to be clear, it could be a follow-up PR to this one, but I just want to have a plan so we can have something to support that use-case ASAP.

dsyer: I think that approach might sometimes be useful. You could also go with a simple single classloader approach. If you want we can talk about options and tradeoffs out of band.

function-invokers/java-function-invoker/src/main/java/io/sk8s/invoker/java/server/JavaFunctionInvokerController.java
```diff
@@ -31,8 +31,8 @@
 	@Autowired
 	private FunctionProperties functions;
 
-	@PostMapping("/")
-	public Object invoke() {
+	@RequestMapping("/")
```
ericbottard: In the context of sk8s, the only client of this is going to be the function sidecar, which we control. So I don't think we need/want to support GET.

function-invokers/java-function-invoker/src/main/java/io/sk8s/invoker/java/server/JavaFunctionInvokerController.java
```diff
@@ -31,8 +31,8 @@
...
+	public String invoke() {
```
ericbottard: Why the change here? Ultimately, we want to support any kind of result, with HttpMessageConverter taking care of it

dsyer: Look at the method body. It only ever returns a String.

function-invokers/java-function-invoker/src/main/java/io/sk8s/invoker/java/server/JavaFunctionInvokerController.java
dsyer: Fine by me. Spring Cloud Function supports GET. I'm still debating with @markfisher what that really means, but he has some ideas.

function-invokers/java-function-invoker/src/main/java/io/sk8s/invoker/java/server/JavaFunctionInvokerController.java
ericbottard: Ah, sorry. Did not realize this was just the forward

function-invokers/java-function-invoker/src/main/java/io/sk8s/invoker/java/server/JavaFunctionInvokerController.java
markfisher: For now, we should stick with `POST` here as that's the current (intentionally constrained) "contract" with the Sidecar when using the "http" dispatcher. We need to document said contract. We might add GET support when we consider the role of an event source "trigger" function/supplier.

function-invokers/java-function-invoker/src/test/java/io/sk8s/invoker/java/function/FunctionConfigurationTests.java
ericbottard: This test no longer tests composition, does it?
Seems the property could only contain `NumberEmitter` (ie drop `Frenchizer`) and it wouldn't change a thing.

Hence, it needs to be renamed IMHO.

function-invokers/java-function-invoker/src/test/java/io/sk8s/invoker/java/function/FunctionConfigurationTests.java
ericbottard: Same as above, this only tests the second function (function1) and hence is not testing composition

function-invokers/java-function-invoker/src/test/java/io/sk8s/invoker/java/server/JavaFunctionInvokerApplicationTests.java
```diff
@@ -16,25 +16,44 @@
...
  */
 @RunWith(SpringRunner.class)
-@SpringBootTest
-@TestPropertySource(properties = "function.uri=file:target/test-classes" +
-		"?io.sk8s.invoker.java.function.FunctionConfigurationTests.Doubler")
-
+@SpringBootTest(webEnvironment = WebEnvironment.RANDOM_PORT)
+@TestPropertySource(properties = "function.uri=file:target/test-classes"
+		+ "?io.sk8s.invoker.java.function.FunctionConfigurationTests.Doubler")
+@AutoConfigureMockMvc
```
ericbottard: Any reason why this test has this annotation, while the other doesn't?

dsyer: No. It's a copy-paste error.

ericbottard:
Merged as 1c998a6, which is Dave's PR with https://github.com/ericbottard/sk8s/commit/137dd22371510886fb6bf115d06b6d7651e8f4d3 on top

PR #121, opened 2017-11-13
shorten service names

pivotal-jamie-klassen:
to correct build errors like

https://ci.faas.to.cf-app.com/teams/pfs/pipelines/sk8s_pr/jobs/deploy-sk8s-pr/builds/17


markfisher:
this was used as a test; no longer needed

PR #123, opened 2017-11-15
Don't close the class loader that is going to be used in functions

dsyer:
Otherwise it can't load any more classes at runtime, which is
obviously a bad thing.


markfisher:
Nice catch! Thanks Dave.

PR #144, opened 2017-11-16
Modified publish/request to work with service using LoadBalancer

### LINKED ISSUE #140: update publish and request script to detect NodePort vs LoadBalancer service type
we might also want to consider merging the two scripts (possibly just a flag to indicate a blocking request/reply)... not necessarily within the scope of this issue, but thought I'd mention it

trisberg:
Fixes #140


markfisher:
LGTM

PR #146, opened 2017-11-16
add slack command example

### LINKED ISSUE #147: Add an example that runs on GKE and listens to the outside world
A slash command for Slack, implemented in Javascript.

pivotal-jamie-klassen:
The "business value" inside the function is just parsing for the "text" param and echoing it back.

Expanding on the above:
If you were invoking a slack command, it might look like `/command-name a bunch of text`. The contract for Slack integrations works by sending a POST request to the configured endpoint with a form param called `text` (in the aforementioned case, more concretely, it looks like `text=a+bunch+of+text`), amongst a bunch of other form-encoded params.

Resolves #147.


trisberg:
Might be better to put this in a `samples/slack` directory with a note that you need a public facing http-gateway for it to work

samples/gke/README.adoc
```diff
@@ -0,0 +1,96 @@
...
+[horizontal]
+*Language*:: JavaScript
+*Function*:: https://github.com/markfisher/sk8s/blob/master/samples/gke/slack-command/slack_command.js[slack_command.js]
+*Protocol*:: HTTP
+*Input*:: slack
+*Output*:: none (replies, implicitly)
+
+=== TL;DR
+
+If you have push access to the `sk8s` org on dockerhub, you can probably run the `test.sh` script in this
```
trisberg: I think we should assume that readers don't have this access

samples/gke/slack-command/slack_command.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: slack
+---
+
+apiVersion: extensions.sk8s.io/v1
+kind: Function
+metadata:
+  name: slack-command
+spec:
+  image: sk8s/slack_command:test
```
trisberg: would prefer to have the image as `slack-command` since that is also the name of the function created, it also matches what we do for all other sk8s images, maybe also rename this file to `slack-command.yaml`

samples/gke/README.adoc
pivotal-jamie-klassen: what does that mean, concretely? Would you say something like
> If you have push access, ..., but you probably don't.
? Would you drop any mention of the test script?

trisberg: I would leave the test script out of this README unless it is easily runnable by all

samples/gke/README.adoc
```diff
@@ -0,0 +1,96 @@
...
+directory and watch the whole process work. It's a bit noisy and the output may be difficult to interpret,
+however. Therefore the behaviour will be described in detail in this document.
+
+=== Deploy the FaaS
+
```
trisberg: Deploying the FaaS shouldn't be part of the sample IMO, we should have that in a separate doc and maybe link to it with a not that this example needs a LoadBalancer with a public facing IP

samples/gke/README.adoc
```diff
@@ -0,0 +1,96 @@
...
+we will generally assume you are able to publish your images to dockerhub.
+
+```
+docker push <your-dockerhub-repo>:<your-tag>
+```
+
+=== Create the Function and Topic(s)
+
+Assuming your kube config is pointed at GKE,
+
```
trisberg: replace GKE with your cluster

samples/gke/README.adoc
```diff
@@ -0,0 +1,96 @@
...
+NAME                     TYPE           CLUSTER-IP      EXTERNAL-IP       PORT(S)        AGE
+sk8s-sk8s-http-gateway   LoadBalancer   10.51.247.194   104.154.118.171   80:30725/TCP   22m
+```
+
+We can see from the `EXTERNAL-IP` and `PORT(S)` columns that our gateway should be reachable at `http://104.154.118.171:80`
+in this case. Since the topic we configured in the previous step was called `slack`, then we should configure our
+slash command to send `POST` requests to `http://104.154.118.171:80/requests/slack`.
+
+=== Tear down the FaaS
+
```
trisberg: Again, this doesn't belong here IMO, maybe have instructions for removing the function

samples/gke/README.adoc
pivotal-jamie-klassen: I mention the deploy here since @jldec's idea was that this was an example that should run on GKE _by definition_. Reading between the lines it sounds like you're trying to relax that assumption in these docs. When I got a GKE cluster with all the defaults (except using k8s 1.8) everything worked out of the box. How exactly would you phrase the requirement that it "needs" one? Does that mean
> Your Google Cloud Platform Quotas should allow for the creation of a public-facing LoadBalancer

would be an appropriate note?

samples/gke/README.adoc
pivotal-jamie-klassen: So here it is a much easier (and future-proof) instruction to have the whole deployment namespaced and then annihilated after the test case. If we say how to remove the function, the instructions for doing so may change down the line and this coupling will have to be maintained.

samples/gke/README.adoc
trisberg: Yes, I don't think we should limit this to GKE, as long as the user has a public facing IP and a LoadBalancer configured, it should work. 

Maybe:
"Your kubernetes cluster should allow for the creation of a LoadBalancer with a public facing IP address"

samples/gke/README.adoc
pivotal-jamie-klassen: As above, are you certain of the need for a level of generality beyond GKE? Mostly I don't want to make claims that I am not confident I can substantiate, given that I have only run the demo on GKE. @jldec was _pretty_ specific when we talked about it, although the only definite statement he said was not to test on Kubo, rather than to only test on GKE.

samples/gke/slack-command/slack_command.yaml
pivotal-jamie-klassen: 👍

samples/gke/README.adoc
pivotal-jamie-klassen: 👍

samples/gke/README.adoc
trisberg: I used a 1.8.2 GKE cluster with RBAC enabled and it still worked fine.

pivotal-jamie-klassen: OK, if you're confident making these claims, I will fully decouple the instructions and leave the details of getting and destroying a cluster in the user's hands.

trisberg:
Merged as commit 2bfdd745049c0045f577e88616118cfaaeae3191

PR #150, opened 2017-11-17
Use function.container from updated model

### LINKED ISSUE #130: function yaml should accept any valid k8s env config
currently we only expose "name" and "value" for the env section of a function yaml, but we should allow "valueFrom" (including support for secrets) such as:

```
env:
- name: DB_PASSWORD
  valueFrom:
    secretKeyRef:
      name: db
      key: db-password
```

ericbottard:
Fixes #130

### DIFF
--- a/function-controller/src/main/java/io/sk8s/function/controller/FunctionDeployer.java
+++ b/function-controller/src/main/java/io/sk8s/function/controller/FunctionDeployer.java
@@ -18,16 +18,10 @@ package io.sk8s.function.controller;
 
 import java.util.Collections;
 import java.util.LinkedHashMap;
-import java.util.List;
 import java.util.Map;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
-
-import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.util.ObjectUtils;
-import org.springframework.util.StringUtils;
-
+import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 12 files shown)


function-controller/src/main/java/io/sk8s/function/controller/FunctionDeployer.java
```diff
@@ -123,29 +117,11 @@ private PodSpec buildPodSpec(XFunction function) {
 	}
 
 	private Container buildMainContainer(XFunction function) {
-		ContainerBuilder builder = new ContainerBuilder().withName("main")
-				.withImage(function.getSpec().getImage())
-				.withImagePullPolicy("IfNotPresent");
```
markfisher: so this is changing our default from `IfNotPresent` to `Always` (can be overridden by a user now though)... possibly this is a good thing since that's the normal default. WDYT @trisberg ?

ericbottard: Actually, it's only `Always` if image is tagged `latest` or no tag. See https://kubernetes.io/docs/concepts/containers/images/#updating-images

We can change, but I believe using the k8s semantics is better (users can specify now, as they have access to the full range of `container` props)

markfisher:
reverted `build-model` to rely on `sk8s` branch (coordinated with that push) and merged as:
https://github.com/markfisher/sk8s/commit/3ad0d58bfe8983d352a165beddde1208eb81a76b

PR #152, opened 2017-11-17
Add sidecar image var and chart values

freynca:
Add sidecar image variable (`SK8S_FUNCTION_CONTROLLER_SIDECAR_TAG`) and corresponding chart values, in order to enable the override of the tag during Helm deployments.

### DIFF
--- a/charts/sk8s/values.yaml
+++ b/charts/sk8s/values.yaml
@@ -18,9 +18,7 @@ functionController:
     pullPolicy: IfNotPresent
   sidecar:
     image:
-      repository: sk8s/function-sidecar
       tag: 0.0.1-SNAPSHOT
-      pullPolicy: IfNotPresent
   service:
     name: http
     type: ClusterIP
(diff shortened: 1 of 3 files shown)


charts/sk8s/values.yaml
```diff
@@ -16,6 +16,11 @@ functionController:
     repository: sk8s/function-controller
     tag: 0.0.1-SNAPSHOT
     pullPolicy: IfNotPresent
+  sidecar:
+    image:
+      repository: sk8s/function-sidecar
```
trisberg: repository and pullPolicy aren't used anywhere in the chart

config/function-controller-deployment.yaml
```diff
@@ -36,4 +36,6 @@ spec:
           value: kafka:9092
         - name: SPRING_CLOUD_STREAM_KAFKA_BINDER_ZK_NODES
           value: zookeeper:2181
+        - name: SK8S_FUNCTION_CONTROLLER_SIDECAR_TAG
+          value: sk8s/function-sidecar
```
trisberg: shouldn't this be `0.0.1-SNAPSHOT`

freynca: Fixed.

charts/sk8s/values.yaml
freynca: Fixed.

trisberg:
squashed and merged as commit 79145552d6d721f537c22e427b790b0bd30f806a

PR #156, opened 2017-11-19
Make use of FUNCTION_URI consistent 

### LINKED ISSUE #126: Make FUNCTION_URI consistent
In the samples, function invokers are inconsistent in how `FUNCTION_URI` is used.  They should all use similar conventions and patterns

dturanski:
Resolves #126 

* `ENV FUNCTION_URI` required in DockerFile
*  No default value provided by function invokers
*  Python URI consistent with JAVA URI (change query param `?handler=` to `?`)


ericbottard:
> ENV FUNCTION_URI required in DockerFile

Not sure what you mean by that (I see no enforcing mechanism, and I don't think we want one). Users should be able to use the java invoker "as is" and dynamically load their function from an external source (ie *not* bake)

ericbottard:
Also, I think it would be better to either
- use a named param (what name would suit all languages is another question) for the entrypoint
- use the fragment concept of a URI (`#`)
to indicate the entrypoint

dturanski:
FUNCTION_URI is required by the invoker.  But I agree not required to be defined in DockerFile. I mean if you're baking it in, as we do in the samples, it should go in DockerFile, not in the yaml.

dturanski:
Re named parameter. I was previously using `handler` for Python. Python is somewhat analogous to Java. I'm ok with that change. For Node JS and shell, a simple file path is sufficient, it would be awkward to require a consistent URL syntax that would suit all languages.

markfisher:
I like the idea of an optional (for the invokers that either require it, or where a default is being overridden) but consistent (when used) parameter. I guess @ericbottard's idea for using a fragment syntax could also be viewed as further qualified location info that's only appended when necessary.

markfisher:
I think the consensus is to go with `?handler=` for Java and Python (param-less URI should be sufficient for shell and Node).

dturanski:
Agreed

dturanski:
Should it be `handlers`,  e.g,  `function.uri=file:target/test-classes?handlers=io.sk8s.invoker.java.function.Doubler,io.sk8s.invoker.java.function.Frenchizer`

markfisher:
I think singular (`handler=`) is fine, because even in the case where we compose a Function chain, it will still be a single handler Function that results from that composition. Also, we have more to do before we advertise composition as a fully supported feature, including a consistent approach across our different invokers (not just java) and/or composition of polyglot functions (potentially as colocated containers within a single pod). This is a topic on the list for team discussions at S1P.

markfisher:
can you please rebase and see if you can figure out the test failures? (there are a couple conflicts to resolve as well but those are straightforward)

markfisher:
nevermind; I think I figured it out... trying samples now

markfisher:
rebased, squashed, resolved conflicts, and merged as https://github.com/markfisher/sk8s/commit/b636661a4c6ef582f47dcfcaebbba9c968f1d60f

PR #158, opened 2017-11-20
Use Spring to call bean lifecycle methods on created functions

dsyer:
This is quite useful if you want to add a @PostConstruct for
instance (could be used to initialize Spring in a mini-context
if not already there)

### DIFF
--- a/function-invokers/java-function-invoker/src/main/java/io/sk8s/invoker/java/server/FunctionConfiguration.java
+++ b/function-invokers/java-function-invoker/src/main/java/io/sk8s/invoker/java/server/FunctionConfiguration.java
@@ -28,7 +28,6 @@ import java.util.function.Function;
 
 import javax.annotation.PostConstruct;
 
-import org.springframework.beans.BeanUtils;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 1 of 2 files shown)


markfisher:
Could we also have a no-annotation-required option? Perhaps an optional env var `FUNCTION_INIT_METHOD` or even a default of "init" (since function implementations will typically not have too many methods). Also, looking at the example in https://github.com/markfisher/s1p2017-faas-demo/pull/5 makes me wonder if we should just create that context ourselves if the `Function` implementation defines any setters? (or trigger it via a boolean flag env var) Of course, we'd always want to back off if the user does take full control.

dsyer:
I think I get what you mean. You make 2 points:

1) User should not need an annotation to mark an init method. I disagree because `@PostConstruct` is in the JDK, so it's just best practice to use it and not a naming convention. For the end of lifecycle there is a naming convention built into Spring (and also `Closeable` in the JDK).

2) User shouldn't have to create the application context manually. Agree. That's going to be an enhancement in a PR later today, or maybe tomorrow.

This PR is finished IMO (and there was no measurable performance impact).

markfisher:
Makes sense (on both points). Thanks.

PR #159, opened 2017-11-20
Launch optionally with isolated classpath (support for Flux)

dsyer:
User has to add --function.runner.isolated=true on the command line. In
this case the app is launched in an isolated classpath with reactor in the
parent class loader. Then the function is launched in a class loader with
the same parent so it can support functions of Flux.

### DIFF
--- a/function-invokers/java-function-invoker/README.md
+++ b/function-invokers/java-function-invoker/README.md
@@ -1,0 +1,21 @@
+Testing locally:
+
+```
+$ mvn install dockerfile:build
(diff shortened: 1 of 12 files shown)


function-invokers/java-function-invoker/README.md
```diff
@@ -0,0 +1,13 @@
+Testing locally:
```
markfisher: should this also include instructions for a `Flux` function (requiring `--function.runner.isolated=true`)?

dsyer: Good point. But there is no README in master. Where are you seeing that?

dsyer: Got it (see latest commit).

markfisher:
squashed, updated comment, and merged as https://github.com/markfisher/sk8s/commit/0045dfda36ed6da4a22a25bc25ce5b462aa65b6d

PR #163, opened 2017-11-20
publish function replica counts to a topic

### LINKED ISSUE #162: publish function replica counts based on k8s deployment updates

markfisher:
The `FunctionMonitor` publishes updates to replica count per `Function` to the `function-replicas` topic (in the embedded-header format so its consumable by other `Functions`). Any `Function` that has that topic as its input will be excluded from the events to avoid a feedback loop. The `topics` shell script's `tail` command now uses the `function-replicas` topic as its default.

The `FunctionMonitor` now also creates a 0-replica `Deployment` immediately upon `Function` registration.

The `FunctionMonitor` now only tracks replica counts for `Deployments` that contain a "function" label.

fixes #162


trisberg:
LGTM

PR #165, opened 2017-11-20
expose configurable idleTimeout for functions

### LINKED ISSUE #149: expose the down-scale linger property in function.yaml
10 seconds is still probably a reasonable default

markfisher:
fixes #149

depends on:
https://github.com/ericbottard/sk8s-types/pull/1/files
and
https://github.com/ericbottard/kubernetes-model/pull/1

the value is interpreted as milliseconds (we could instead go with seconds?)

this property is optional, and the default is still 10 seconds


ericbottard:
Merged as 1e7195

PR #167, opened 2017-11-21
Set the context class loader during function creation

dsyer:
Lots of things use the thread context loader, including Spring
itself when creating services from spring.factories. So we need to
set it to something sane when the functions are being instantiated.

Also adds a new special URL for the function.uri. If the user uses
"app:classpath" then the classpath URLs from the app are re-used.
Very useful for testing, but might also be occasionally useful
for real functions that don't want to be a fat jar and
only need Spring stuff that is already in the invoker.


markfisher:
LGTM; rebased and merged as https://github.com/markfisher/sk8s/commit/dae5d38fa77034868e4d6060d6d059da81a571cf

Issue #157, opened 2017-11-19
Support @Autowiring for Spring Java functions


trisberg:
We need to load the app context and make sure the autowiring support works for Spring Boot apps following the spring-cloud-function model.

markfisher:
I assigned Dave as I believe he already has this on a branch.

ericbottard:
Closing this as
- it pertains to the java-function-invoker
- I assume it is working now (@dsyer please re-create if this is not the case)

PR #171, opened 2017-11-21
Automatically launch Spring Boot if possible and main class provided

dsyer:
User has to provide a main class in the function.uri, and the functions
and looked up by name and then by type, e.g.

function.uri=file:/path/to/function-sample-pof.jar?handler=greeter&main=functions.Application

The function jar has to be shaded still (or else use app:classpath).

See gh-157


markfisher:
LGTM; merged into the extracted `java-function-invoker` project via https://github.com/projectriff/java-function-invoker/commit/5fe673c851d5e08aaa6f3592d96a9e9fb441d905

Issue #161, opened 2017-11-20
Update helm instructions in the Getting Started doc


jldec:
https://github.com/projectriff/riff/blob/master/Getting-Started.adoc is a little out of date. Suggested fixes:

- [x] correct helm repo info
- [x] more details about package contents, image tags etc.
- [x] how to override tags
- [x] how to use helm-installed images for building samples (see #96)
- [x] how to override with a custom build of one component

trisberg:
We also need to move off the SNAPSHOT tags to a curated set of artifacts that work together and won't be overwritten by later builds

markfisher:
A subset of the checkbox items in this issue are being addressed by this PR:
https://github.com/projectriff/riff/pull/182

trisberg:
The rest by https://github.com/projectriff/riff/pull/214

PR #181, opened 2017-11-23
Use 'upstream' as the remote alias for repos.

### LINKED ISSUE #180: Don't use origin as the remote name

ericbottard:
Fixes #180

### DIFF
--- a/build-model
+++ b/build-model
@@ -1,9 +1,2 @@
 #!/bin/bash
-set -e -v
-
-# regenerate the fabric8 java API including k8s custom resources
-pushd $(go env GOPATH)/src/github.com/fabric8io/kubernetes-model/
-git fetch origin
-git checkout origin/riff
-make
-popd
+./build-helper "$(go env GOPATH)/src/github.com/fabric8io/kubernetes-model/" ${1:-''} ${2:-''} 'make'
--- a/build-python2-function-invoker
+++ b/build-python2-function-invoker
@@ -1,8 +1,2 @@
 #!/bin/bash
-set -e -v
-
-pushd $PWD/../python2-function-invoker/
-git fetch origin
-git checkout origin/master
-./build.sh
-popd
+./build-helper "$PWD/../python-function-invoker/" ${1:-''} ${2:-''} './build.sh'
--- a/README.adoc
+++ b/README.adoc
@@ -4,7 +4,7 @@
 
 You can install the system using link:Getting-Started.adoc#helm[a Helm chart] or by link:#manual[manually building and deploying] the components.
 
-=== [[manual]]Manual install of riff
+=== [[manual]] Manual install of riff
 
 ==== Prerequisites
 
@@ -12,6 +12,7 @@ You need:
 
 * A running 1.7+ Kubernetes cluster with at least 4GB of memory. These instructions assume minikube for now.
 +
+[source,bash]
 ----
 minikube start --cpus=4 --memory=4096 --disk-size=64g
 ----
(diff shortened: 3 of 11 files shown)


Getting-Started.adoc
```diff
@@ -1,101 +0,0 @@
-= Sk8s: a FaaS for Kubernetes - Getting Started
```
markfisher: I think we want to keep this guide. The helm charts are working, so this can be adapted (this issue is relevant: https://github.com/projectriff/riff/issues/161). In fact, I think the README should just change to point to either the helm install (for most people) or the manual install (for us and developers who want to operate at that level).

markfisher:
maybe we should take this opportunity to move all build scripts into a `build` sub-dir?

ericbottard:
> maybe we should take this opportunity to move all build scripts into a build sub-dir?

Depends if we expect more stuff to appear in this project in the end.
Having a quick way to build x or y is a good thing on a day to day basis.

I also have a `build-deploy-all` script locally:
```
#!/bin/bash
./build-model
eval $(minikube docker-env)
./build-function-sidecar
./build-http-gateway
./build-topic-controller
kubectl apply -f config/types
./build-function-controller
./build-function-invokers
kubectl apply -f config/kafka
kubectl apply -f config
```

Only downside is that it has a tendency to re-build the model when it's seldom needed

Getting-Started.adoc
ericbottard: Sure, my mistake. I assumed it was meant to go as it hadn't been `s/sk8s/riff`'d.
Seems like it's more "Install via helm" than "Getting started" though?

markfisher:
ok, we can wait on any sub-dir changes

I like the composite script, but maybe we should have separate `build-all` and `deploy-riff` commands? (the build script could still apply the `config/types` so those tests run but maybe that could be driven by a flag... and they could be removed after the build)

Getting-Started.adoc
markfisher: in general for users who aren't going to be working with the framework code, I think the helm install *is* the best getting started experience (the current `README` in `helm-charts` already shows a single command to deploy riff using locally built Docker images

markfisher: When https://github.com/projectriff/riff/pull/182 is ready for merging, it will be modifying this file, so we should leave it as is for now.

markfisher:
I created a separate issue for tracking the build/deploy-all scripts: https://github.com/projectriff/riff/issues/183

build-model
```diff
@@ -1,9 +1,2 @@
 #!/bin/bash
-set -e -v
-
-# regenerate the fabric8 java API including k8s custom resources
-pushd $(go env GOPATH)/src/github.com/fabric8io/kubernetes-model/
-git fetch origin
-git checkout origin/riff
-make
-popd
+./build-helper "$(go env GOPATH)/src/github.com/fabric8io/kubernetes-model/" ${1:-''} ${2:-''} 'make'
```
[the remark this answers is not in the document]
ericbottard: If we make a branch the default, then there is no way to use your local state.
I made it "the default" in the README, but these scripts need to have blank as the default

build-python2-function-invoker
```diff
@@ -1,8 +1,2 @@
 #!/bin/bash
-set -e -v
-
-pushd $PWD/../python2-function-invoker/
-git fetch origin
-git checkout origin/master
-./build.sh
-popd
+./build-helper "$PWD/../python-function-invoker/" ${1:-''} ${2:-''} './build.sh'
```
markfisher: this should be "python2-function-invoker"

markfisher:
LGTM

PR #182, opened 2017-11-23
Updated helm instructions

freynca:
Updated helm instructions

### DIFF
--- a/Getting-Started.adoc
+++ b/Getting-Started.adoc
@@ -1,13 +1,11 @@
-= Sk8s: a FaaS for Kubernetes - Getting Started
-
-IMPORTANT: We don't have a curated and published release yet, so please use the manual build process documented in the link:README.adoc[README] for the time being.
+= Riff: a FaaS for Kubernetes - Getting Started
 
 == Installation
 
-=== [[helm]]Helm install of Sk8s FaaS
+=== [[helm]]Helm install of Riff FaaS
 
-Sk8s offers a Helm Chart for deploying the FaaS and its required services to a Kubernetes Cluster.
-The following instructions cover how to initialize Helm and install Sk8s FaaS on the Minikube cluster.
+Riff offers a Helm Chart for deploying the FaaS and its required services to a Kubernetes Cluster.
+The following instructions cover how to initialize Helm and install Riff FaaS on the Minikube cluster.
 
 ==== Prerequisites
 
@@ -39,29 +37,31 @@ helm init --service-account=tiller
 
 NOTE: To verify that the Tiller pod has started execute the following command: `kubectl get pod --namespace kube-system` and you should see the `tiller` pod running.
 
-==== Add the Sk8s repository to your Helm environment
+==== Add the Riff repository to your Helm environment
 
-Run the following commands to add the `sk8s-incubator` repository:
+Run the following commands to add the repository:
 
 ----
-helm repo add sk8s-incubator http://sk8s-incubator.cfapps.io
-helm repo update
+helm repo add riffrepo https://riff_charts.storage.googleapis.com
+helm repo update riffrepo
 ----
 
-==== Install Sk8s FaaS helm chart
+==== Install Riff FaaS helm chart
 
-Install the Sk8s Helm chart using one of the following commands.
+Install the Riff Helm chart as follows:
+
+- `helm search riffrepo/riff` and identify the version you're looking to install (use the latest one if unsure)
 
 For a cluster that does not have RBAC enabled use:
 
 ----
-helm install --name demo sk8s-incubator/sk8s
+- `helm install riffrepo/riff --name=faas --version=<version>`
 ----
 
 For a cluster that does have RBAC enabled use:
 
 ----
-helm install --name demo sk8s-incubator/sk8s --set create.rbac=true
+- `helm install riffrepo/riff --name=faas --version=<version> --set create.rbac=true`
 ----
 
 You should see a number of resources get created in your Minikube. You can see them all using:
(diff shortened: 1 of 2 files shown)


Getting-Started.adoc
```diff
@@ -39,29 +39,35 @@ helm init --service-account=tiller
...
+- Download the starter install script: `curl -sL https://riff_charts.storage.googleapis.com/riff-<version>-install-example.sh > riff_install.sh`
+- Ensure the script has execute permissions: `chmod +x riff_install.sh`
+
+*Note:* the starter script above contains references to Docker image tags appropriate for the version of the Helm chart being deployed. If a value needs to be overridden, arguments to the script can be passed in the form `--set key=value`
 
 For a cluster that does not have RBAC enabled use:
 
 ----
-helm install --name demo sk8s-incubator/sk8s
+./riff_install.sh riffrepo/riff --name=riff
```
markfisher: I don't think we want to show "riff" as the example name since that leads to all of the components being named `riff-riff-<componentname>`, but maybe we want to reconsider that naming strategy (not sure if it's a helm idiom @trisberg ?)... an example from the chart templates:

```
metadata:
  name: {{ template "fullname" . }}-function-controller
```

Getting-Started.adoc
```diff
@@ -1,13 +1,13 @@
-= Sk8s: a FaaS for Kubernetes - Getting Started
+= Riff: a FaaS for Kubernetes - Getting Started
 
 IMPORTANT: We don't have a curated and published release yet, so please use the manual build process documented in the link:README.adoc[README] for the time being.
```
markfisher: this note should be removed

Getting-Started.adoc
```diff
@@ -39,29 +39,35 @@ helm init --service-account=tiller
...
 ----
 
-==== Install Sk8s FaaS helm chart
+==== Install Riff FaaS helm chart
 
-Install the Sk8s Helm chart using one of the following commands.
+Install the Riff Helm chart as follows:
+
+- `helm search riffrepo/riff` and identify the version you're looking to install (use latest if unsure)
+- Download the starter install script: `curl -sL https://riff_charts.storage.googleapis.com/riff-<version>-install-example.sh > riff_install.sh`
```
markfisher: I think here we should show the instructions for installing via a simple `helm install` with the latest stable release which will match the version in `Chart.yaml` as soon as we kick off the first such release, even if that means kicking off the first release immediately before merging the updated guide (cc @trisberg).

The instructions for downloading a script to install a specific build version via overrides should be either in a separate Developer Guide or at least in a section that follows completion of the "normal" Getting Started experience.

We also need to provide a `values-snapshot.yaml` that overrides all values to: `x.y.z.-snapshot` where `x.y.z` is the next planned release. Those tags would match the images built locally using the various `build-*` scripts within this project. Instructions for deploying the locally built images should also follow the "normal" steps. Personally I think the order should be:

1. full instructions with a simple `helm install` using the (latest stable) version as defined in `helm-charts/Chart.yaml`
2. supplement section that provides an example `helm install` command using `values-snapshot.yaml` to override for locally built image tags
3. supplement section that provides instructions for downloading the script used by a specific build version

thoughts?

markfisher:
overriding with `values-snapshot.yaml` works locally, but the other install does not currently work even though the "official" `Chart.yaml` [here](https://github.com/projectriff/helm-charts/blob/master/values.yaml) is currently listing `0.0.1-SNAPSHOT` versions as well. Based on `helm inspect` it appears as though the published chart's values' imagePullPolicies were all overridden to "Always" (whereas the `values.yaml` in the `helm-charts` repo has "IfNotPresent"). That might be the right thing to do for charts published for incremental builds (versions `x.y.x-build.n`), but we will want to make sure when we release the official `0.0.1` chart that it is exactly what we have under version control.

Getting-Started.adoc
```diff
@@ -87,15 +87,19 @@ These flags can be set when installing the chart or you can apply them during an
...
 ----
-helm delete demo --purge
+helm delete riff --purge
 ----
 
+==== Installing Local Builds
+
+When building snapshot images locally, pass in a reference to the `values-snapshot.yaml` file, which contains snapshot tags:
 
+- `helm install riffrepo/riff --name=faas --values=config/values-snapshot.yaml`
```
markfisher: this should probably include `--set httpGateway.service.type=NodePort` or possible we should provide 2 separate sections here, one for minikube and one for external clusters

@trisberg WDYT?

trisberg: I'm preparing a commit with `values-snapshot.yaml` and `values-nodeport.yaml` that containes only those values that we want to override so we can install using:

```
helm install --name demo --values helm/values-snapshot.yaml,helm/values-nodeport.yaml riffrepo/riff
```

I'd prefer to put these files somewhere other than the `config` directory

trisberg: See: https://github.com/projectriff/riff/pull/186

markfisher:
After reviewing several charts in https://github.com/kubernetes/charts default including PR comments by the maintainers, it appears the default `imagePullPolicy` values should be "IfNotPresent". Also, I don't think we should ever be generating charts (even in the build pipeline), since the values we need to override can always be exposed.

Getting-Started.adoc
markfisher: I've merged the PR for adding `values-snapshot.yaml` (with only versions) and `values-nodeport.yaml` into the `helm` subdirectory, so we need to adapt these instructions accordingly, and we can remove the `values-snapshot.yaml` included with this PR.

trisberg:
Yeah, not sure why we change the `imagePullPolicy` on the published charts, it should be `IfNotPresent` like we have in the `helm-charts` repo. I don't see the need to publish the charts each build, they are curated artifacts that only need to be published when they are modified in the repo.

markfisher:
Yes, I think we agreed to only publish the script that overrides the chart values with each build, but not the chart itself.

markfisher:
closing in favor of #187 which builds upon this and addresses the comments above as well as incorporating the new `helm/values-nodeport.yaml` and `helm/values-snapshot.yaml` files.

We still have the issue of charts currently in the repo overriding the `imagePullPolicy`. We should delete those charts.

PR #185, opened 2017-11-25
pass params to each build invoker

jldec:
Required for the build instructions in the README to pull latest changes for each sub-project.

### DIFF
--- a/build-function-invokers
+++ b/build-function-invokers
@@ -1,8 +1,8 @@
 #!/bin/bash
 set -e -v
 
-./build-java-function-invoker ${1:-''} ${2:-''}
-./build-node-function-invoker ${1:-''} ${2:-''}
-./build-shell-function-invoker ${1:-''} ${2:-''}
-./build-python2-function-invoker ${1:-''} ${2:-''}
+./build-java-function-invoker "$@"
+./build-node-function-invoker "$@"
+./build-shell-function-invoker "$@"
+./build-python2-function-invoker "$@"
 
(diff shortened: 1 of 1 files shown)


build-function-invokers
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
 set -e -v
 
-./build-java-function-invoker
-./build-node-function-invoker
-./build-shell-function-invoker
-./build-python2-function-invoker
+./build-java-function-invoker ${1:-''} ${2:-''}
```
jldec: i did a little more digging - using `"$@"` (quoted) should work as well to pass all parameters through and is a little simpler.

#### script1
```bash
#!/bin/bash
set -e -v

echo '$1='"$1" '$2='"$2"

./script2 $*

./script2 $@

./script2 "$*"

./script2 "$@"
```

#### script2
```bash
#!/bin/bash
set -e -v

echo '$1='"$1" '$2='"$2"
```
#### results
note that only  `"$@"` seems to pass multi-word parameters through correctly
```sh
$ ./script1 X "Y 2 3"

echo '$1='"$1" '$2='"$2"
$1=X $2=Y 2 3

./script2 $*

echo '$1='"$1" '$2='"$2"
$1=X $2=Y

./script2 $@

echo '$1='"$1" '$2='"$2"
$1=X $2=Y

./script2 "$*"

echo '$1='"$1" '$2='"$2"
$1=X Y 2 3 $2=

./script2 "$@"

echo '$1='"$1" '$2='"$2"
$1=X $2=Y 2 3
```

ericbottard:
Verified args are passed as expected, and merged as 638e90c

PR #186, opened 2017-11-26
Add values files to override provided values for Helm install

trisberg:
- use values-nodeport.yaml for overriding service type of LoadBalancer with NodePort

- use values-snapshot.yaml for overriding riff component tag with SNAPSHOT build tags

- install using `helm install --name demo --values helm/values-snapshot.yaml,helm/values-nodeport.yaml riffrepo/riff`


markfisher:
LGTM

PR #194, opened 2017-11-29
updates after 0.0.1 release

markfisher:
- Dockerfiles for samples all use 0.0.1 function invoker base images
- local config yaml for riff components all use 0.0.2-snapshot
- the values-snapshot.yaml used local helm install overrides also uses 0.0.2-snapshot


trisberg:
LGTM

PR #196, opened 2017-11-29
Switch back to snapshot images in samples

### LINKED ISSUE #195: Change samples to use snapshot base images again

ericbottard:
Fixes #195

### DIFF
--- a/samples/java/greeter/Dockerfile
+++ b/samples/java/greeter/Dockerfile
@@ -1,4 +1,4 @@
-FROM projectriff/java-function-invoker:0.0.1
+FROM projectriff/java-function-invoker:0.0.2-snapshot
 ARG FUNCTION_JAR=/functions/greeter-1.0.0.jar
 ARG FUNCTION_CLASS=functions.Greeter
 ADD target/greeter-1.0.0.jar $FUNCTION_JAR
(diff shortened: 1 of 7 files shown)


markfisher:
we discussed this yesterday and decided to leave samples on the latest stable version (devs can always change locally)... however, we can reopen the debate if you'd like. The argument in favor of changing to snapshots is that we do have a `v0.0.1` tagged version of the riff repo: https://github.com/projectriff/riff/tree/v0.0.1

cc @trisberg

samples/shell/echo/echo-function.yaml
```diff
@@ -3,7 +3,7 @@ kind: Function
 metadata:
   name: echo
 spec:
-  protocol: stdio
+  protocol: pipes
```
markfisher: I think you left this here on accident

ericbottard: ah, yes. `git add -A samples/` :(

trisberg:
I'd like to keep the samples at `0.0.1` for now, makes it easier to run them locally and on GKE. If I do local dev work I can change the sample. Just my 2c.

markfisher:
closing for now

PR #198, opened 2017-11-29
Simplify command

dturanski:


Getting-Started.adoc
```diff
@@ -48,7 +48,9 @@ helm repo add riffrepo https://riff-charts.storage.googleapis.com
 helm repo update riffrepo
 ----
 
-==== Install the riff helm chart
+==== Install Riff
```
trisberg: riff

trisberg:
I don't like the local vs. non-local classification - it's really "clusters supporting LoadBalancer" vs. "clusters without LoadBalancer support". Also, I think we should have the "--set httpGateway.service.type=NodePort" option for a non-customized install, there is no need to check out the riff repo for this. Same as here: --set httpGateway.service.type=NodePort

trisberg:
Squashed and merged as commit 52bb8aa215b60338f752bdf222e0f309c18e936b

Added an additional commit with some clarifications for manual and rbac installs.

PR #202, opened 2017-12-01
Slack sample

dturanski:


samples/node/slack/README.adoc
```diff
@@ -0,0 +1,148 @@
...
+
+```
+Forwarding                    http://a61d8bb0.ngrok.io -> 192.168.99.100:32595
+```
+
+=== Build the Docker container for the function code
+
+```
+eval $(minikube docker-env)
+docker build -t projectriff/slack-command:test samples/node/slack/slack-command
```
ericbottard: Should be replaced by `./riff build ...` to be consistent with other samples

ericbottard: Why the extra `slack-command` dir level?

samples/node/slack/README.adoc
```diff
@@ -0,0 +1,148 @@
...
+```
+=== Create the Function and Topic(s)
+
+```
+./riff apply -f samples/node/slack/slack-command
+```
+
+=== Test the function locally
+
+Since the input topic configured for the function is named `slack`, configure the slash command to `POST` to `http://<ngrokid>.ngrok.io/requests/slack`.
```
ericbottard: Without recreating Slash's own documentation, we should at least explain that the user needs to create a Slack command (+ link to where to start on their site / doco).
This sentence states "the slash command" but I as a reader have no idea what it is (unless I missed something)

samples/node/slack/README.adoc
```diff
@@ -0,0 +1,148 @@
...
+curl -H"Content-Type:application/x-www-form-urlencoded" http://<ngrokid>.ngrok.io/requests/slack -d text='riff is for functions'
+received:riff is for functions
+```
+[TIP]
+You can use the local URL `http://$NODE_IP:$NODE_PORT` here, but a publicly accessible URL is required for Slack integration.
+
+
+[slack-local]
+=== Set up your Slack integration
+Use the `ngrok.io` URL To create your slack https://api.slack.com/slash-commands[slash command].
```
ericbottard: Actually, I think the section I commented about above should be here.

And maybe put the "simulating locally" section *after* the real (with a link allowing to skip to it just before the "real" section)

samples/node/slack/README.adoc
```diff
@@ -0,0 +1,148 @@
...
+[NOTE]
+Running this sample on a cloud platform like GKE will require you to build and publish the image to a docker repository available to the platform.
+In GKE you can use the https://cloud.google.com/container-registry/[GCP Container Registry] if it is enabled. Here we will assume
+ dockerhub(https://hub.docker.com/). You will need to tag the image you build accordingly.
+ In these instructions, we refer to the tag as `<your-docker-tag>`. For dockerhub, this is typically something like `<docker-user>/slack-command:test`
+
+=== Build the Docker container for the function code
+
+```
+docker build -t <your-docker-tag> samples/node/slack/slack-command
```
ericbottard: Same remark as above

samples/node/slack/README.adoc
```diff
@@ -0,0 +1,148 @@
...
+==== Get the Gateway IP address and Port:
+
+```
+$ kubectl get svc -l component=http-gateway
+NAME                     TYPE           CLUSTER-IP      EXTERNAL-IP       PORT(S)        AGE
+riff-http-gateway        LoadBalancer   10.51.247.194   104.154.118.171   80:30725/TCP   22m
+```
+
+We can see from the `EXTERNAL-IP` and `PORT(S)` columns that our gateway is on `http://104.154.118.171:80`
+in this case. Since the topic we configured for the function is named `slack`, in this case, we would configure the
```
ericbottard: Double "in this case", I guess the second one can be removed

samples/node/slack/README.adoc
```diff
@@ -0,0 +1,148 @@
+= slack-command
+
+[horizontal]
+*Language*:: JavaScript
+*Function*:: https://github.com/projectriff/riff/blob/master/samples/node/slack/slack-command/slack_command.js[slack_command.js]
```
ericbottard: Why don't we use relative links? (this applies globally, obviously)

dturanski: agree

ericbottard:
Merged as e050f7f

PR #203, opened 2017-12-02
Fix the RBAC config scripts

trisberg:
- add ClusterRole allowing access to CustomResourceDefinitions

- add ClusterRoleBinding for new ClusterRole


markfisher:
LGTM

PR #211, opened 2017-12-12
Pull directly from github

dturanski:
Depends on https://github.com/projectriff/function-proto/pull/4


ericbottard:
Merged as 0e4cb2b

PR #216, opened 2017-12-15
Add a version command

trisberg:
- change the default for function version to be 0.0.1 instead of v0001

### DIFF
--- a/riff
+++ b/riff
@@ -1,9 +1,13 @@
 #!/bin/bash
 
+RIFF_VERSION=0.0.1
+
 function print_usage() {
 cat <<EOF
 
 riff is for functions
+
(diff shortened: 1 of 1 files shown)


riff
```diff
@@ -1,17 +1,22 @@
 #!/bin/bash
 
+RIFF_VERSION=0.0.1
```
markfisher: shouldn't this one be 0.0.2? (or will that be a single commit before the release?)

trisberg: I'll update that when we release

markfisher: cool

markfisher:
LGTM; if you want to handle it (with that one change)

PR #224, opened 2017-12-20
default riff build tag should match sample yaml image tag

jldec:
Ran into this during a demo today - i like to cd to the sample and then `riff build` using the defaults.


trisberg:
I discussed this with @markfisher and the thinking was that the default should apply to new functions that would start from `0.0.1` while existing samples should have instructions where the version is specified, currently at `0.0.2`. So I would prefer to keep the default as is.

jldec:
ok - will go with convention you propose - nice to see the VERSION can be overridden now.

PR #229, opened 2018-01-05
Change to work with updated function-proto

dturanski:
Depends on https://github.com/projectriff/function-proto/pull/8

Need to change `GIT_PATH` in samples/python/uppercase/dockerize


dturanski:
Merged

Issue #197, opened 2017-11-29
add high level `riff` command


markfisher:
this should accept the function artifact and the invoker (short) name, for example (not sure about the command name, but maybe we should use "create" and remove that existing command since "apply" can accomplish the same):

```
riff create -f myfunc.js -i node
```

could also take `-n` (name) or `-v` (version) but those could have defaults, with the default name being the artifact name minus its extension

it will generate the `Dockerfile` and then call `riff build` and `riff apply`

PR #230, opened 2018-01-07
Update riff to be able to create function definition files

trisberg:
- add init command to initialize function definitions with defaults based on language used

- change create to call init, build and apply

- add push option for build

Resolves https://github.com/projectriff/riff/issues/197

### DIFF
--- a/riff
+++ b/riff
@@ -43,8 +43,7 @@ Options:
   -f, --filename: filename or directory to be used for the function resources, 
                   if a file is specified then the file's directory will be used 
                   (defaults to the current directory)
-  -f, --filename: filename, directory, or URL for the resource definition[s] (defaults to the current directory)
-  -l, --language: the language used for the functiuon source
+  -l, --language: the language used for the function source
                   (defaults to filename extension language type or 'shell' if directory specified)
   -p, --protocol: the protocol to use for function invocations
                   (defaults to 'stdio' for shell and python, to 'http' for java and node)
@@ -52,8 +51,9 @@ Options:
   -o, --output: the name of the output topic (no default, only created if specified)
   -a, --artifact: path to the function artifact, source code or jar file
                   (defaults to function name with extension appended based on language: 
-                      '.sh' for shell, '.jar' for java, 'js' for node and '.py' for python)
+                      '.sh' for shell, '.jar' for java, '.js' for node and '.py' for python)
   --classname: the fully qualified class name of the Java function class (required for Java functions)
+  --riff-version: the version of riff to use when building containers
   --push: push the image to Docker registry
 
 EOF
@@ -97,8 +97,7 @@ Options:
   -f, --filename: filename or directory to be used for the function resources, 
                   if a file is specified then the file's directory will be used 
                   (defaults to the current directory)
-  -f, --filename: filename, directory, or URL for the resource definition[s] (defaults to the current directory)
-  -l, --language: the language used for the functiuon source
+  -l, --language: the language used for the function source
                   (defaults to filename extension language type or 'shell' if directory specified)
   -p, --protocol: the protocol to use for function invocations
                   (defaults to 'stdio' for shell and python, to 'http' for java and node)
@@ -106,8 +105,9 @@ Options:
   -o, --output: the name of the output topic (no default, only created if specified)
   -a, --artifact: path to the function artifact, source code or jar file
                   (defaults to function name with extension appended based on language: 
-                      '.sh' for shell, '.jar' for java, 'js' for node and '.py' for python)
+                      '.sh' for shell, '.jar' for java, '.js' for node and '.py' for python)
   --classname: the fully qualified class name of the Java function class (required for Java functions)
+  --riff-version: the version of riff to use when building containers
   --push: push the image to Docker registry
 
 EOF
@@ -212,7 +212,7 @@ EOF
 }
 
 function init() {
-  # assume the file extension based on languag - might be overridden as necessary later
+  # assume the file extension based on language - might be overridden as necessary later
   if [[ $FNLANG == "" || $FNLANG == "shell" ]] ; then
     FNEXT="sh"
   elif [[ $FNLANG == "java" ]] ; then
@@ -583,6 +583,10 @@ case ${key} in
   ;;
   -o|--output)
   TOPIC_OUT="$2"
+  shift
+  ;;
+  --riff-version)
+  RIFF_VERSION="$2"
   shift
   ;;
   -t|--tail)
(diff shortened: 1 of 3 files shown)


riff
```diff
@@ -23,6 +25,40 @@ Commands:
...
+
+Options:
+
+  -u, --useraccount: the Docker user account to be used for the image repository (defaults to current OS username)
+  -n, --name: the name of the function (defaults to the name of the current directory)
+  -v, --version: the version of the function (defaults to 0.0.1)
+  -f, --filename: filename or directory to be used for the function resources, 
+                  if a file is specified then the file's directory will be used 
+                  (defaults to the current directory)
+  -f, --filename: filename, directory, or URL for the resource definition[s] (defaults to the current directory)
```
markfisher: duplicate `-f` here (I think the 2nd one should be deleted?)

riff
```diff
@@ -23,6 +25,40 @@ Commands:
...
+  -l, --language: the language used for the functiuon source
```
markfisher: typo: "functiuon"

riff
```diff
@@ -23,6 +25,40 @@ Commands:
...
+                  (defaults to filename extension language type or 'shell' if directory specified)
+  -p, --protocol: the protocol to use for function invocations
+                  (defaults to 'stdio' for shell and python, to 'http' for java and node)
+  -i, --input: the name of the input topic (defaults to function name)
+  -o, --output: the name of the output topic (no default, only created if specified)
+  -a, --artifact: path to the function artifact, source code or jar file
+                  (defaults to function name with extension appended based on language: 
+                      '.sh' for shell, '.jar' for java, 'js' for node and '.py' for python)
+  --classname: the fully qualified class name of the Java function class (required for Java functions)
```
markfisher: this is fine for now, but perhaps we want to generalize this (e.g. "handler") or possibly allow it as an optional suffix on the artifact value (after a delimiter perhaps)

riff
```diff
@@ -31,29 +67,48 @@ Build the function based on the code available in the path directory, using the
...
 
 Options:
 
+  -u, --useraccount: the Docker user account to be used for the image repository (defaults to current OS username)
+  -n, --name: the name of the function (defaults to the name of the current directory)
+  -v, --version: the version of the function (defaults to 0.0.1)
+  -f, --filename: filename or directory to be used for the function resources, 
+                  if a file is specified then the file's directory will be used 
+                  (defaults to the current directory)
   -f, --filename: filename, directory, or URL for the resource definition[s] (defaults to the current directory)
```
markfisher: should delete this duplicate `-f` as well

riff
```diff
@@ -31,29 +67,48 @@ Build the function based on the code available in the path directory, using the
...
+  -l, --language: the language used for the functiuon source
```
markfisher: typo: "functiuon"

riff
```diff
@@ -31,29 +67,48 @@ Build the function based on the code available in the path directory, using the
...
+                  (defaults to filename extension language type or 'shell' if directory specified)
+  -p, --protocol: the protocol to use for function invocations
+                  (defaults to 'stdio' for shell and python, to 'http' for java and node)
+  -i, --input: the name of the input topic (defaults to function name)
+  -o, --output: the name of the output topic (no default, only created if specified)
+  -a, --artifact: path to the function artifact, source code or jar file
+                  (defaults to function name with extension appended based on language: 
+                      '.sh' for shell, '.jar' for java, 'js' for node and '.py' for python)
```
markfisher: should include a dot before "js" for consistency

riff
```diff
@@ -23,6 +25,40 @@ Commands:
...
+  -f, --filename: filename, directory, or URL for the resource definition[s] (defaults to the current directory)
+  -l, --language: the language used for the functiuon source
+                  (defaults to filename extension language type or 'shell' if directory specified)
+  -p, --protocol: the protocol to use for function invocations
+                  (defaults to 'stdio' for shell and python, to 'http' for java and node)
+  -i, --input: the name of the input topic (defaults to function name)
+  -o, --output: the name of the output topic (no default, only created if specified)
+  -a, --artifact: path to the function artifact, source code or jar file
+                  (defaults to function name with extension appended based on language: 
+                      '.sh' for shell, '.jar' for java, 'js' for node and '.py' for python)
```
markfisher: should include a dot before "js" for consistency

riff
```diff
@@ -133,24 +211,267 @@ Usage:
 EOF
 }
 
-function build() {
-  echo "Building $FUNCTION $FNPATH $VERSION"
-  docker build -t "projectriff/${FUNCTION}:${VERSION}" $FNPATH
+function init() {
+  # assume the file extension based on languag - might be overridden as necessary later
```
markfisher: typo: "languag"

riff
```diff
@@ -133,24 +211,267 @@ Usage:
...
+}
+
+function update() {
+  if [ -f "${FNPATH}" ] ; then
+    FNPATH=$(dirname "$FNPATH")
+  fi
+  FNDOCKER="$FNPATH/Dockerfile"
+  if [ -f $FNDOCKER ] ; then
+    echo "Updating $FNPATH resource[s]"
+    kubectl delete -f $FNPATH
```
markfisher: shouldn't we leave the topics alone?

riff
trisberg: yeah, not sure what handler options there will be for other languages and also not sure why we require the classname for Java since SCFn should be able to pull it from the Jar manifest "Function-Class" entry.

riff
trisberg: how do you change partition count? seems like our controllers don't apply what changed for modified resources

trisberg: also, if we don't delete the topic I get the `The Topic "hello" is invalid: apiVersion: Invalid value: "projectriff.io/__internal": must be projectriff.io/v1` error

markfisher: AFAIK that's an open issue for topic-controller (perhaps this one: https://github.com/projectriff/topic-controller/issues/3)

@ericbottard can clarify

One thing to keep in mind is that topics can be shared across multiple functions, so we'll need to decide how to handle that. I guess the proper behavior for `apply` with a topic resource YAML file would be to only consider changes (e.g. if the partition count is different than the existing topic). If multiple functions share a topic, and they have the same generated YAML for that topic, it wouldn't be an issue. But we might want to consider whether we should provide another flag for `init` so that topic YAML need not be generated. With a more sophisticated CLI at some point, we could actually add generate-if-not-exists functionality.

markfisher: Considering topic resource deletions are not yet processed by `topic-controller` (as you reported via https://github.com/projectriff/topic-controller/issues/17), I guess we might as well go ahead with this one as-is. We can revisit once that work is completed for the `topic-controller`.

markfisher:
LGTM; squashed and merged as https://github.com/projectriff/riff/commit/ea2c1368147437c039baf88350a4a4308316a155

PR #232, opened 2018-01-08
Add publish command

trisberg:
- use `-t` to "tail" waiting for the response -- `riff publish -d "world" -t`

- use `--count` to publish multiple messages (use ${i} variable to get the iteration) -- `riff publish -d 'My test ${i}' --count 3`

### DIFF
--- a/riff
+++ b/riff
@@ -212,9 +212,9 @@ Usage:
 Options:
 
   -i, --input: the name of the input topic (defaults to the name of the current directory)
-  -d, --data: the data to post to the http-gatewy using the input topic
+  -d, --data: the data to post to the http-gateway using the input topic
   --count: the number of times to post the data (defaults to 1)
-  --pause: the number of seconds to wait bewteen postings (defaults to 0)
+  --pause: the number of seconds to wait between postings (defaults to 0)
   -t, --tail: tail, waiting for the function response
 
 EOF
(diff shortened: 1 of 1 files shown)


riff
```diff
@@ -199,6 +200,26 @@ Options:
...
+Publish data to a topic using the http-gateway.
+
+Usage:
+
+  riff publish -i <input-topic> -d <data> [--count <count>] [--pause <pause>] [-t]
+
+Options:
+
+  -i, --input: the name of the input topic (defaults to the name of the current directory)
+  -d, --data: the data to post to the http-gatewy using the input topic
```
markfisher: typo: gatewy

riff
```diff
@@ -199,6 +200,26 @@ Options:
...
+  --count: the number of times to post the data (defaults to 1)
+  --pause: the number of seconds to wait bewteen postings (defaults to 0)
```
markfisher: typo: bewteen

riff
```diff
@@ -199,6 +200,26 @@ Options:
...
+  -t, --tail: tail, waiting for the function response
```
markfisher: not crazy about "tail" since it's just waiting for a single result every time in this case, but I don't have a better suggestion at the moment

trisberg: I'm not crazy about it either, but we already had that option. Would `--follow` be better for both use cases?

markfisher: I guess we can stick with it for now. We'll likely make changes when we move to a binary CLI.

markfisher:
squashed and merged as https://github.com/projectriff/riff/commit/e92a263bb4736623aa4f2a858d2d343b3553122e

PR #237, opened 2018-01-09
use riff create for square sample

markfisher:


samples/node/square/README.adoc
```diff
@@ -6,32 +6,26 @@
...
-./riff logs -n square -t
+riff publish -d 99 -t
 ```
 
-=== Delete the function
+=== delete the function and its input topic
 
 ```
-./riff delete -n square
+riff delete -n square --all
```
trisberg: you only need `riff delete --all` since the `--all` option uses the $FNPATH which  defaults to current dir unless `-f` is specified

trisberg: The usage was updated in the help text

```
Usage:

  riff delete -n <name>
    or
  riff delete -f <path> --all
```

markfisher: gotcha

trisberg:
Rebased and merged as commit bdf6c35daa91cc5b2f4b66dbf28c051d9b5c3007

PR #243, opened 2018-01-10
Update the help texts for the delete command

trisberg:
- the delete affects the resources and not their definitions


markfisher:
LGTM

PR #245, opened 2018-01-10
Use -r/--reply for publish and wait for reply

trisberg:


markfisher:
need to update the square sample README now as well

markfisher:
will do on merge

markfisher:
LGTM

PR #248, opened 2018-01-13
Multi build python

dturanski:
This modifies the Dockerfile to use multi-stage build. This removes the grpcio-tools components from the final image, decreasing its size by ~30%.


jldec:
closing since this code is moving out to a new repo under the github/projectriff-samples org.

PR #246, opened 2018-01-10
Update riff script to be on par with Windows batch file

trisberg:
Add explicit -e/--eval option to publish, you can do:

`riff publish -d 'test ${i}' -e --count 4` and get the following messages published:

```
Test 1
Test 2
Test 3
Test 4
```

Make it possible to run create from any directory, It should be possible to run:

```
cd ~
mkdir -p ~/Test/riff/square
echo 'module.exports = (x) => x**2' > ~/Test/riff/square/square.js
riff create -f ~/Test/riff/square
riff publish -r -d 22 -f ~/Test/riff/square
```

PR #250, opened 2018-01-16
Update samples to latest

trisberg:
Can be tested by setting RIFF_VERSION env var to `0.0.3-snapshot`

- requires riff script from https://github.com/projectriff/riff/pull/246

- removes output topic so all samples can be used with `--reply`

- I suggest moving remaining more complex samples to separate riff-samples project


markfisher:
since we no longer have any function with output, we should create a simple example in the new `riff-samples` repo that demonstrates a pipeline of 2 functions (such that the first of those has an explicit output topic defined)

markfisher:
rebased, squashed, and merged as https://github.com/projectriff/riff/commit/74e63a7d691150a6be66471bbbccc1b1a329d183

PR #251, opened 2018-01-24
update build for Go Function Controller

markfisher:


trisberg:
Why remove the probes? There is a PR adding the in.

Also, `JAVA_TOOL_OPTIONS` can go away

README.adoc
```diff
@@ -61,21 +60,14 @@ From now on, this README assumes you're running commands from the `riff` reposit
...
-
 ==== Point your local Docker environment to the Docker environment running in minikube:
 
 [source, bash]
 ----
 eval $(minikube docker-env)
 ----
 
-==== Build the Function Sidecar, HTTP Gateway and Topic Controller
+==== Build the Function Sidecar, HTTP Gateway, and Topic Controller
```
trisberg: Couldn't we add the build of the Function Controller here as well, looks weird to have that as a separate step later now that it's a Go project

markfisher: it was originally due to the tests expecting those resources to be created, but if that's no longer a requirement, I agree they should all be together

trisberg:
Squashed and merged as commit de937b3983eaf618c2464f73b17963c8c1962e87

README.adoc
trisberg: I'll submit a separate PR with some other changes as well

PR #272, opened 2018-02-03
Change protocol

making:
`pipes` is used as a `protocol` for java functions by default.

https://github.com/projectriff/riff/blob/v0.0.3/riff#L315-L316


trisberg:
LGTM, merged as commit 819e37c4e4717978bf01e036ad0caf75ac8871b6

PR #277, opened 2018-02-09
Force delete on update

glyn:
Fixes https://github.com/projectriff/function-controller/issues/5.

### DIFF
--- a/riff
+++ b/riff
@@ -480,7 +480,7 @@ function update() {
   FNDOCKER="$FNPATH/Dockerfile"
   if [ -f $FNDOCKER ] ; then
     echo "Updating $FNPATH resource[s]"
-    kubectl delete -n $NAME_SPACE -f $FNPATH
+    kubectl delete -n $NAME_SPACE --force=true -f $FNPATH
     build
     apply
   else


riff
```diff
@@ -480,7 +480,7 @@ function update() {
   FNDOCKER="$FNPATH/Dockerfile"
   if [ -f $FNDOCKER ] ; then
     echo "Updating $FNPATH resource[s]"
-    kubectl delete -n $NAME_SPACE -f $FNPATH
+    kubectl delete -n $NAME_SPACE --force=true -f $FNPATH
```
ericbottard: Why isn't this calling `delete()` somehow, and shouldn't the `force` apply there too?

trisberg: this was a"hack" that we put in until we can do a true update without the delete

glyn: Since `--force` is pretty brutal, I think we should restrict it to the scenarios that are known to need it. The Go rewrite can get the structure right and https://github.com/projectriff/function-controller/issues/9 can provide a better solution.

ericbottard: Also, obviously the meaning of `force` here is not at all the one found in the documentation of `kubectl apply`, which was mentioned before.

I'm then curious to understand how "immediate" deletion of the **function** resource can have any effect, as the `function-controller` does not "propagate" deletion attributes from the `function` resource to the `deployment` resource

ericbottard: This can't be a case of `kubectl` waiting for the server to delete, given that we post the deletion message in a go channel

glyn: For the meaning of `--force`, please see `kubectl delete --help`.

ericbottard: Obviously, I did look there, hence my comments: I don't see how this could have an impact (and given I have a very hard time reproducing the bug in the first place...)

glyn: I wonder if k8s has any knowledge of the dependency between the deployment and the function? Of so, deleting the function promptly might free up the deployment to be deleted more quickly.

markfisher: I think we were planning to add an `ownerReference`, but that would mean 1.8+ (probably fine)

ericbottard: > I wonder if k8s has any knowledge of the dependency between the deployment and the function

That's what I was trying to say above: it doesn't (currently). Moreover, everything is (should) be decoupled, given that we post the event in a go channel, to be handled asynchronously by another goroutine, so I'm really clueless as to why this would help.

I'd like to understand what's happening before merging :)

glyn: Me too. Maybe it was simply a fluke that the scenario appeared to fail consistently before this change and pass consistently afterwards.

trisberg:
Can we close this now that we changed the function-controller and switched to the Go CLI as well?

jldec:
closing per https://github.com/projectriff/function-controller/issues/5#issuecomment-366177890

PR #278, opened 2018-02-12
Switch samples to work with go shell-function-invoker

glyn:
This should be merged along with https://github.com/projectriff/shell-function-invoker/pull/2.

### DIFF
--- a/samples/shell/echo/echo.sh
+++ b/samples/shell/echo/echo.sh
@@ -1,1 +1,3 @@
-echo "received: $1"
+#!/bin/sh
+
+printf "received at `date -u`: `cat`"
--- a/samples/shell/wordcount/wordcount.sh
+++ b/samples/shell/wordcount/wordcount.sh
@@ -1,1 +1,3 @@
-echo "$1" | wc -w
+#!/bin/sh
+
+wc -w
--- a/config/delve/shell-function-invoker-debug.yaml
+++ b/config/delve/shell-function-invoker-debug.yaml
@@ -1,0 +1,16 @@
+apiVersion: v1
+kind: Service
+metadata:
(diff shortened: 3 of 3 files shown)


samples/shell/wordcount/wordcount.sh
```diff
@@ -1 +1,3 @@
-echo "$1" | wc -w
+#!/bin/sh
+
+echo `cat` | wc -w
```
ericbottard: Should be able to be just `wc -n`, shouldn't it?

samples/shell/echo/echo.sh
```diff
@@ -1 +1,3 @@
-echo "received: $1"
+#!/bin/sh
+
+echo -n `cat`
```
ericbottard: This doesn't add "received:".

ericbottard: Maybe add ``date`` or something similar here, so that each invocation can be obviously different

samples/shell/wordcount/wordcount.sh
glyn: I don't think `wc` supports `-n`: https://linux.die.net/man/1/wc

samples/shell/echo/echo.sh
glyn: Good idea. Will do.

samples/shell/wordcount/wordcount.sh
ericbottard: Sorry, I meant just `wc -w`, ie no `echo`

glyn: Yes, well spotted. Fixed.

glyn:
@ericbottard Fixed both your points. Please take another look.

PR #282, opened 2018-02-16
simplify shell echo sample

jldec:
```
~/riff/riff/samples/shell/echo (simpler-echo)$ riff publish -d world -r 
Posting to http://192.168.99.100:31982/requests/echo
hello
world
```


samples/shell/echo/echo.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/sh
 
-printf "received at `date -u`: `cat`"
+echo hello ; cat
```
markfisher: how about this? `xargs echo hello`

glyn: +1: no gratuitous newline.

jldec:
@glyn i took @markfisher's suggestion and added a "hello" sample. The echo sample just has `cat` now. What do you think?

trisberg:
Squashed and merged as commit ce2d769cbc18c4c5c469f1b5e91db5684c77273c

glyn:
@jldec `cat` LGTM, but you'll probably need to give a brief explanation in demos as not everyone will know that `cat` reads from standard input by default. (I had to look it up, for example.)

PR #287, opened 2018-02-19
Prepare for 0.0.4 release

trisberg:


Getting-Started.adoc
```diff
@@ -57,61 +61,75 @@ helm search riff -l
 
 == [[current]]Install the current riff release
 
-=== Install riff using a Kubernetes cluster supporting LoadBalancer
+Starting with the 0.0.4 version we have separated the Kafka installation from the riff chart. This requires an extra step of installing Kafka in order yo use riff. You can use the single-node Kafka chart provided by riff or the three-node kafka/zookeeper service provided by the Kubeapps.
```
ericbottard: In order *to* use riff, YO!

Will fix

Getting-Started.adoc
```diff
@@ -57,61 +61,75 @@ helm search riff -l
...
 [source, bash]
 ----
-helm install riffrepo/riff --version 0.0.3 --name demo --set httpGateway.service.type=NodePort
+helm install riffrepo/riff --version 0.0.4 --name demo --set create.rbac=false
 ----
 
 [NOTE]
 ====
-For a cluster that has RBAC enabled add `--set create.rbac=true`:
+For a cluster that has RBAC enabled use `--set create.rbac=true`:
```
ericbottard: Isn't that the default?

trisberg: Tes , it is now, maybe we should just remove it and add a comment that create.rbac=true is the new default

ericbottard: I think so, yes

Getting-Started.adoc
jldec: could also omit `the` at the end of the paragraph (before `Kubeapps`).

trisberg:
Rebased, fixed and merged as commit 029cfafe49b9c6189e31bba8791ac114f3615a8d

PR #289, opened 2018-02-20
Rename sentiments-service.py to sentiments.py

### LINKED ISSUE #288: Rename sentiment_service.py to sentiments.py
renaming would eliminate the need to override the filename when you do `riff create` using the cli.

https://github.com/projectriff/riff/blob/master/samples/python/sentiments/sentiment_service.py

dturanski:
Fixes #288


trisberg:
LGTM

PR #296, opened 2018-02-27
Going monorepo

ericbottard:
This merges the following repos into `riff` and changes so that the bare minimum still works:
```
function-controller 
function-proto 
function-sidecar 
helm-charts 
http-gateway 
kubernetes-crds 
message-transport 
riff-cli 
topic-controller
```

The first ~220 commits (up until 261cb68) are preserved history of each repo.

Then, on top of that:
- rewrote go import paths 
- removed `vendor/` dirs inside each project
- created `vendor/` dir at root of repo (converting `riff-cli` to glide along the way)

Each `make` default target seems to be working and produces a running executable.


ericbottard:
For reference, this is the helper script I used:
```
#!/bin/bash

# See https://stackoverflow.com/questions/1425892/how-do-you-merge-two-git-repositories/10548919#10548919

REPOS="function-controller function-proto function-sidecar helm-charts http-gateway kubernetes-crds message-transport riff-cli topic-controller"

function rewrite_paths() {
	git filter-branch -f --tree-filter "zsh -c 'setopt extended_glob && setopt glob_dots && mkdir -p $1 && (mv ^(.git|$1) $1 || true)'" -- --all
}


function merge_repos() {

	for repo in riff $REPOS ; do
		git clone https://github.com/projectriff/$repo
		pushd $repo
			echo $repo >> /tmp/counts
			git rev-list --count HEAD >> /tmp/counts
			echo "" >> /tmp/counts
			rewrite_paths $repo
		popd			
	done

	pushd riff
		for repo in $REPOS ; do
			git remote add  --fetch $repo ../$repo
			git merge --allow-unrelated-histories --no-edit $repo/master
		done

		echo "---------------" >> /tmp/counts
		git rev-list --count HEAD >> /tmp/counts
	popd

}

function rewrite_imports() {
	pushd riff

		for repo in $REPOS ; do
			pushd $repo
				GOFILES=$(find pkg cmd -type f -name '*.go')
				for f in $GOFILES ; do
					sed -i '' 's|"github.com/projectriff/\(.*\)"|"github.com/projectriff/riff/\1"|g' $f
				done
			popd
		done
	popd
}


rm /tmp/counts
mkdir work
pushd work
popd work

#merge_repos()
rewrite_imports

```

jchesterpivotal:
Does this show as branch-per-repo or are the commits interleaved by time? I can't tell from Github's UI, but it looks like interleaved.

ericbottard:
I believe it's branch-per-repo. I don't expect you'll be checking those commits anyway ;)

I left them with 2 parents because it's easier to spot that there is something special happening then

markfisher:
the `riff-cli` is not included in the top-level build

function-controller/Makefile
```diff
@@ -3,33 +3,28 @@ OUTPUT = function-controller
...
-glide.lock: glide.yaml
-	glide up -v --force
-
-gen-mocks:
-	mockery -name 'TopicInformer|FunctionInformer' -dir vendor/github.com/projectriff/kubernetes-crds/pkg/client/informers/externalversions/projectriff/v1
-	mockery -name 'SharedIndexInformer' -dir vendor/k8s.io/client-go/tools/cache
-	mockery -name 'DeploymentInformer' -dir vendor/k8s.io/client-go/informers/extensions/v1beta1
-	mockery -name 'LagTracker|Deployer' -dir pkg/controller
+#TODO
+#gen-mocks:
```
markfisher: does this mean testing is currently not working? (wondering why there is no `test` target in the root build)

ericbottard: no, just that re-generating mocks is not supported (given they involved paths inside vendor)

function-sidecar/Makefile
```diff
@@ -1,36 +1,22 @@
-.PHONY: build clean fetch-grpc grpc dockerize test
+.PHONY: build clean grpc dockerize test
```
markfisher: the "grpc" target was removed (not needed here)

ericbottard:
> the riff-cli is not included in the top-level build

Added that. Wanted to give it special treatment so we can `go get github.com/projectriff/riff`

PR #298, opened 2018-02-28
Fix ASCII doc nesting in README

ericbottard:


glyn:
Merged as ddd922fc2ed9a60bdca707c80bfcce9147809b89

PR #303, opened 2018-02-28
Fix mocks generation paths after monorepo move

ericbottard:


glyn:
Merged as 65b0a232c449e8cada919316fc9e12178c8b582f

PR #304, opened 2018-02-28
Use travis-ci

scothis:


scothis:
The build is failing, but this also seems to accurately reflect my local environment.

Issue #305, opened 2018-02-28
http-gateway tests fail intermittently


glyn:
Example of the failure (from [CI](https://ci.projectriff.io/teams/main/pipelines/riff/jobs/build-component-images-dev/builds/14?)):
```
2018/02/28 17:19:30 Listening on :16911
•! Panic [0.007 seconds]
HTTP Gateway
/go/src/github.com/projectriff/riff/http-gateway/pkg/server/gateway_integration_test.go:35
  should accept messages and fire&forget [It]
  /go/src/github.com/projectriff/riff/http-gateway/pkg/server/gateway_integration_test.go:99

  Test Panicked
  runtime error: invalid memory address or nil pointer dereference
  /usr/local/go/src/runtime/panic.go:491

  Full Stack Trace
  	/usr/local/go/src/runtime/panic.go:491 +0x283
  github.com/projectriff/riff/http-gateway/pkg/server_test.glob..func1.5()
  	/go/src/github.com/projectriff/riff/http-gateway/pkg/server/gateway_integration_test.go:112 +0x271
  github.com/projectriff/riff/vendor/github.com/onsi/ginkgo/internal/leafnodes.(*runner).runSync(0xc420492de0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
  	/go/src/github.com/projectriff/riff/http-gateway/pkg/server/server_suite_test.go:12 +0x64
  testing.tRunner(0xc4200102d0, 0x89dba8)
  	/usr/local/go/src/testing/testing.go:746 +0xd0
  created by testing.(*T).Run
  	/usr/local/go/src/testing/testing.go:789 +0x2de
  
------------------------------
••••••••••

Summarizing 2 Failures:

[Panic!] HTTP Gateway [It] should request/reply OK 
/usr/local/go/src/runtime/panic.go:491

[Panic!] HTTP Gateway [It] should accept messages and fire&forget 
/usr/local/go/src/runtime/panic.go:491

Ran 25 of 25 Specs in 0.043 seconds
FAIL! -- 23 Passed | 2 Failed | 0 Pending | 0 Skipped --- FAIL: TestHandler (0.04s)
FAIL
FAIL	github.com/projectriff/riff/http-gateway/pkg/server	0.087s
Makefile:10: recipe for target 'test' failed
make: *** [test] Error 1
```

glyn:
The test should assert that an error did not occur so we get better diagnostics:
```
[...]

PR #306, opened 2018-02-28
Make gateway test less racey

jchesterpivotal:
The http-gateway tests work on local systems, but tend to fail on
CI. This appeared to be due largely to a race condition between
the code that starts the server and the tests themselves.

- Add waitForHttpGatewayToBeReady()
- Add error panics in post()
- Replace Expect/To with Eventually/Should
- Pull gw.Run(done) into JustBeforeEach

[Github projectriff/riff#305]

### DIFF
--- a/http-gateway/pkg/server/gateway_integration_test.go
+++ b/http-gateway/pkg/server/gateway_integration_test.go
@@ -82,8 +82,8 @@ var _ = Describe("HTTP Gateway", func() {
 				message.Headers{server.CorrelationId: msg.Headers()[server.CorrelationId],
 					"Content-Type": []string{"bag/plastic"},
 				})
-			Eventually(msg.Headers()["Content-Type"]).Should(Equal([]string{"text/solid"}))
-			Eventually(msg.Headers()["Not-Propagated-Header"]).Should(BeNil())
+			Expect(msg.Headers()["Content-Type"]).To(Equal([]string{"text/solid"}))
+			Expect(msg.Headers()["Not-Propagated-Header"]).To(BeNil())
 		})
 
 		resp := doRequest(port, "foo", bytes.NewBufferString("world"), "Content-Type", "text/solid", "Not-Propagated-Header", "secret")
@@ -91,9 +91,9 @@ var _ = Describe("HTTP Gateway", func() {
 		b := make([]byte, 11)
 		resp.Body.Read(b)
 
-		Eventually(b).Should(Equal([]byte("hello world")))
-		Eventually(resp.Header.Get(server.CorrelationId)).Should(BeZero())
-		Eventually(resp.Header.Get("Content-Type")).Should(Equal("bag/plastic"))
+		Expect(b).To(Equal([]byte("hello world")))
+		Expect(resp.Header.Get(server.CorrelationId)).To(BeZero())
+		Expect(resp.Header.Get("Content-Type")).To(Equal("bag/plastic"))
 
 		defer resp.Body.Close()
 	})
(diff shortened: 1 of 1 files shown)


http-gateway/pkg/server/gateway_integration_test.go
```diff
@@ -143,18 +142,11 @@ func post(port int, path string, body io.Reader, headerKV ...string) *http.Respo
...
-		if err != nil {
-			fmt.Print(".")
-			time.Sleep(durationBetweenAttempts)
-			continue
-		}
-
-		break
-	}
+	Eventually(func() error {
+		_, err := http.Get(url)
```
glyn: There is no handler for the path `/`, so I'm not convinced this will work or, if it does work some of the time, whether it will always work. I was thinking it might be better to wrap the POST requests in the test in `Eventually`.

http-gateway/pkg/server/gateway_integration_test.go
```diff
@@ -143,18 +142,11 @@ func post(port int, path string, body io.Reader, headerKV ...string) *http.Respo
...
+		return err
+	}).Should(Succeed())
```
glyn: `Eventually` is taking the default timeout of 1 second. I know this seems like aeons, but on a really slow (CI) system, it might not be enough. How about upping it to 10 seconds?

glyn:
@jchesterpivotal "Add error panics in post()" is dubious and "Pull gw.Run(done) into JustBeforeEach" is inaccurate. Please could you correct the commit log with your next version?

http-gateway/pkg/server/gateway_integration_test.go
```diff
@@ -79,40 +82,37 @@ var _ = Describe("HTTP Gateway", func() {
 				message.Headers{server.CorrelationId: msg.Headers()[server.CorrelationId],
 					"Content-Type": []string{"bag/plastic"},
 				})
-			Expect(msg.Headers()["Content-Type"]).To(Equal([]string{"text/solid"}))
-			Expect(msg.Headers()["Not-Propagated-Header"]).To(BeNil())
+			Eventually(msg.Headers()["Content-Type"]).Should(Equal([]string{"text/solid"}))
```
glyn: Eventually should take a closure rather than a value, since a value will never change. ;-)

ericbottard:
Waiting for the GW to be ready in `JustBeforeEach()` is the right (and sole) approach IMO. There is no need for the `Eventually` constructs in the tests then (which are actually not changing anything as we realized with @glyn because you're not passing them a `func()`). So these should be changed back.

As for the `waitForHttpGatewayToBeReady()` function, it can try to hit the `/application/status` endpoint of the GW, which should reply with `200`

Lastly, should the panics be changed to Gomega' `Should()` on `err` for concistency?

http-gateway/pkg/server/gateway_integration_test.go
```diff
@@ -126,10 +126,27 @@ func doMessage(port int, topic string, body io.Reader, headerKV ...string) *http
 
 func post(port int, path string, body io.Reader, headerKV ...string) *http.Response {
 	client := http.Client{}
-	req, _ := http.NewRequest("POST", fmt.Sprintf("http://localhost:%v%v", port, path), body)
+	req, err := http.NewRequest("POST", fmt.Sprintf("http://localhost:%v%v", port, path), body)
+	if err != nil {
```
glyn: As @ericbottard was intimating, this block should be replaced by `Expect(err).NotTo(HaveOccurred())`, throughout.

http-gateway/pkg/server/gateway_integration_test.go
jchesterpivotal: I can step down to the previous tactic of dialling the socket instead of performing a full HTTP `GET` request. Not sure I want to add logic to fix a test.

http-gateway/pkg/server/gateway_integration_test.go
jchesterpivotal: Ah, good call.

http-gateway/pkg/server/gateway_integration_test.go
ericbottard: See my comment about hitting the health handler

jchesterpivotal:
@ericbottard I looked at your suggestion of changing the `panic()`s to be `Eventually`s.

It adds a surprising amount of noise to the test helper methods. For example, here's one take:

```Go
func post(port int, path string, body io.Reader, headerKV ...string) *http.Response {
	var response *http.Response
	var request *http.Request

	Eventually(func() error {
		req, err := http.NewRequest("POST", fmt.Sprintf("http://localhost:%v%v", port, path), body)
		if err != nil {
			return err
		}

		for i := 0; i < len(headerKV); i += 2 {
			req.Header.Add(headerKV[i], headerKV[i+1])
		}
		request = req
		return err
	}).Should(Succeed())

	Eventually(func() error {
		client := http.Client{}
		resp, err := client.Do(request)
		response = resp

		return err
	}).Should(Succeed())

	return response
}
```

jchesterpivotal:
Ah! I see that the intention is to use `Expect` for those too, as @glyn was pointing out.

ericbottard:
Merged as 4cb98b8f

PR #307, opened 2018-03-01
Fix function-proto VCS URLs after going monorepo

ericbottard:


function-proto/java/pom.xml
```diff
@@ -47,9 +47,9 @@ limitations under the License.
 	</licenses>
 
 	<scm>
-		<url>https://github.com/projectriff/function-proto</url>
-		<connection>scm:git:git://github.com/projectriff/function-proto.git</connection>
-		<developerConnection>scm:git:ssh://git@github.com/projectriff/function-proto.git</developerConnection>
+		<url>https://github.com/projectriff/riff/tree/master/function-proto</url>
```
glyn: The previous value of `<url>` didn't include `master`, so was it in error? (If not, the new value may be in error.)

ericbottard: This is just a "to be viewed by humans URL" as far as I can tell.
The previous url did not include master because it was a "link to repo". Now it has to be a "link to directory inside a repo (and in some branch)"

So I'd argue it was ok before, and it's ok now. It's good you have sharp eyes though :)

ericbottard: To put it another way, the previous URL could have been `https://github.com/projectriff/function-proto/tree/master/` if we wanted to compare apples to apples

glyn: Ok, thanks.

PR #332, opened 2018-03-01
Add clean to top-level Makefile. Re-add dependency on vendor/

ericbottard:


function-controller/Makefile
```diff
@@ -20,11 +20,18 @@ gen-mocks:
...
+debug-dockerize: $(GO_SOURCES) ../vendor
 	# Need to remove probes as delve starts app in paused state
 	-kubectl patch deploy/function-controller --type=json -p='[{"op":"remove", "path":"/spec/template/spec/containers/0/livenessProbe"}]'
 	-kubectl patch deploy/function-controller --type=json -p='[{"op":"remove", "path":"/spec/template/spec/containers/0/readinessProbe"}]'
 	docker build . -t projectriff/function-controller:$(TAG) -f Dockerfile-debug
+
+../vendor: ../glide.lock
+	$(MAKE) -C .. vendor
+
+../glide.lock: ../glide.yaml
```
glyn: This allows a “top level” make operation to be initiated from the component makefile. This could result in undefined behaviour if component makefiles are invoked concurrently, but I guess that's unlikely to happen in practice.

ericbottard: yeah, maybe I overused recursive `make` here.
Maybe just `cd .. && glide up -v etc...`?

ericbottard:
Merged after simplifying vendor/ thing as dicussed in above comment

PR #339, opened 2018-03-02
Ignore coverage artefacts

glyn:


ericbottard:
I'll trust you on this, but how do I generate those? 
They're not generated by `go test -cover` are they?

glyn:
`ginkgo -cover` produces them.

PR #340, opened 2018-03-02
Publish docker images from travis

scothis:
Pushes images for function-controller, function-sidecar, http-gateway
and topic-controller.

When dockerizing, each component will look at the `VERSION` file to
determine the current version. It can be overridden by the TAG env var.

Docker imagers are only published from the master branch if the other
tests pass.


.travis.install-kafka.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+set -e
+
```
jchesterpivotal: I usually use this blurb:
```
set -o errexit
set -o nounset
set -o pipefail
```

A little more verbose. I think `nounset` is a bashism, so it might not work here.

function-controller/Makefile
```diff
@@ -1,7 +1,7 @@
 .PHONY: build clean dockerize gen-mocks test kubectl-apply
 OUTPUT = function-controller
 GO_SOURCES = $(shell find pkg cmd -type f -name '*.go')
-TAG = 0.0.5-snapshot
+TAG ?= $(shell cat ../VERSION)
```
jchesterpivotal: Out of curiousity, why not `$(cat ../VERSION)`?

scothis: I thought I tried that and it didn't work, but my make skills are weak.

.travis.install-kafka.sh
scothis: no reason we can't use bash

function-controller/Makefile
jchesterpivotal: And my noticing-it's-not-bash skills are weaker.

.travis.docker-push.sh
```diff
@@ -0,0 +1,28 @@
...
+
+version=`cat VERSION`
+images=(
+    "function-controller"
+    "function-sidecar"
+    "http-gateway"
+    "topic-controller"
+)
+
+docker login -u "$DOCKER_USERNAME" -p "$DOCKER_PASSWORD"
```
jchesterpivotal: Might be worth using the `--stdin` flag for this: https://docs.docker.com/engine/reference/commandline/login/#parent-command. They claim it'll prevent leaking to logs. That said I think travis blocks it too.

scothis: This is the format travis uses in their docs, the warning is annoying, but `echo`ing the password doesn't feel better to me

jchesterpivotal: Works for me

PR #341, opened 2018-03-04
Issue 338

### LINKED ISSUE #338: riff delete --all doesn't work if not in function-dir (used with -n)
I assume it tries to read the local `yaml` file. It should query the function resource instead

dturanski:
Fixes #338 .Since `delete --all` functionality requires executing `kubectl get function -o json` to determine the topics, I made some changes to enable stubbing the `kubeclt.Exec` functions. Most of the delete tests no longer set `--dry-run`.

### DIFF
--- a/riff-cli/pkg/kubectl/kubectl.go
+++ b/riff-cli/pkg/kubectl/kubectl.go
@@ -21,8 +21,11 @@ import (
 	"time"
 )
 
+var EXEC_FOR_STRING = ExecForString
+var EXEC_FOR_BYTES = ExecForBytes
+
 func ExecForString(cmdArgs []string) (string, error) {
-	out, err := ExecForBytes(cmdArgs)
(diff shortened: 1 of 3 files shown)


riff-cli/pkg/kubectl/kubectl.go
```diff
@@ -21,8 +21,11 @@ import (
 	"time"
 )
 
+var EXEC_FOR_STRING = ExecForString
```
trisberg: doesn't look like this is used

trisberg: never mind, saw it is used in `riff-cli/cmd/delete.go`

trisberg:
LGTM

trisberg:
Squashed, rebased and merged as commit 7b529d0e0d72d5e9fb4aa905e2e11534ea01f30f

PR #346, opened 2018-03-06
Use riff-system as the namespace for kafka.

ericbottard:
This aligns with what we use for helm charts and makes the topics script work again

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -50,7 +50,7 @@ make dockerize
 
 [source, bash]
 ----
-kubectl apply -f config/kafka
+kubectl apply -n riff-system -f config/kafka
 ----
 
 === Deploy the riff Custom Resource Definitions and deploy the riff core components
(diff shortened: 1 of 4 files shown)


README.adoc
```diff
@@ -50,7 +50,7 @@ make dockerize
 
 [source, bash]
 ----
-kubectl apply -f config/kafka
+kubectl apply -n riff-system -f config/kafka
```
trisberg: this depends on the namespace being there - we should add instructions to create it like we do https://github.com/projectriff/riff/blob/master/Getting-Started.adoc#install-kafka-chart[here]

ericbottard: Can you add that one line, or do you want me to do it and merge?

trisberg: Sure, I can add it on merge

trisberg: Also adding `-n riff-system` to the teardown script

trisberg:
Rebased and merged as commit 6717f5af8908591c9ab344976204cc19fa633c3e

PR #351, opened 2018-03-07
Move away from all commands being global vars.

### LINKED ISSUE #350: Re-architecture CLI to rely less on global vars

ericbottard:
Move away from use of init() functions.

Some commands still mutate state that is kept globally (in opts package). TBC
Tests still refer to commands as global vars. Added temporary workarounds for that

Fixes #350

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -19,11 +19,19 @@ package main
 import (
 	"github.com/projectriff/riff/riff-cli/cmd"
 	"github.com/projectriff/riff/riff-cli/global"
+	"fmt"
+	"os"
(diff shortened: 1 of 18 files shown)


ericbottard:
Still a lot to do, but this is a first step.

Delete wasn't consistent in the way it dealt with global state compared to other functions, so I made that explicit (this is bad, but at least it's consistent now)

trisberg:
Rebased and merge as commit 65b8effb8b2b89882ce07267bfd6771f6ce53948

PR #357, opened 2018-03-07
use --namespace=riff-system with helm

jldec:
Changes the helm instructions (both devel and release) to use riff-system instead of default namespace for the riff components.

Does not change the manual install instructions in the main README or the PUBLISH_NAMESPACE default in the CLI.


trisberg:
Merged as commit 7d7698177ac495626b5d9d15101f1eb42aeb1c17

PR #359, opened 2018-03-07
Drop support for stdio and pipes function-sidecar protocols

scothis:
All invokers now support `grpc`. We can standardize on `grpc` as the sidecar-invoker protocol with `http` as a fallback.

The default protocol for python in the riff-cli is now `grpc`.


markfisher:
LGTM

PR #363, opened 2018-03-08
Publish helm charts via travis

scothis:
The charts now have release agnostic versions that are pinned down when
published to the helm repository. If the riff version does not end with
"-snapshot" then the current release is marked as the latest.

Helm charts are only published from the master branch after the test
complete successfully.


.travis.helm-publish.sh
```diff
@@ -0,0 +1,38 @@
...
+
+  mkdir -p $work_dir
+  gsutil cp "gs://$helm_charts_bucket/index.yaml" $work_dir/
+
+  if [[ $riff_version != *-snapshot ]]; then
+    echo "Setting latest version to $riff_version"
+    echo $riff_version > $work_dir/latest_version
+  fi
+
+  sed -E -i '' "s/tag:[ ]*latest/tag: $riff_version/" riff/values.yaml
```
trisberg: we should also update the README.md file changing "|latest|" to "|$riff_version|"

.travis.helm-publish.sh
```diff
@@ -0,0 +1,38 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+riff_version=`cat VERSION`
+helm_charts_bucket='riff-charts'
+helm_charts_url="https://${helm_charts_bucket}.storage.googleapis.com/"
+work_dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )/helm-charts/.work"
+
+if [ "$GCLOUD_CLIENT_SECRET" != ""]; then
```
trisberg: should this be:

`if [[ "$GCLOUD_CLIENT_SECRET" != "" ]]; then`

my bash complained

.travis.helm-publish.sh
scothis: done

.travis.helm-publish.sh
scothis: done

trisberg:
Squashed and merged as commit 11834f2beb9a34a312c27273d3a5b3b5277b9517

PR #364, opened 2018-03-08
Don't default namespace to "default"

### LINKED ISSUE #345: riff CLI default namespace should respect the default namespace of the kubectl context
The following will create new function and topic resources in the `default` namespace instead of `funkytown`.

```
kubectl config set-context $(kubectl config current-context) --namespace=funkytown
riff create
```

### LINKED ISSUE #360: change CLI default for Publish Namespace to riff-system
#357 changed the helm instructions to use the `riff-system` namespace for riff installs.

Please modify the CLI to make this the default so that `riff publish` works without extra parameters or configuration.

trisberg:
- if --namespace option is provided use it, otherwise don't include it in the commands

- this helps when namespace is specified in the kubeconfig, now we will not override that

- no need to specify namespace for publish, we'll search all namespaces

Fixes #345
Fixes #360


ericbottard:
LGTM, I'll merge with regenerated docs

ericbottard:
Merged as 69058eaf

PR #365, opened 2018-03-09
Don't hardcode target as the path for a Java function artifact

### LINKED ISSUE #347: CLI generated Dockerfile always looks for JAR artifact in 'target' directory
When running:

```
riff create java -a uppercase.jar --handler functions.Uppercase
```

The path to the JAR artifact is clearly defined as the working directory.

The Dockerfile includes this line

```Dockerfile
ADD target/uppercase.jar $FUNCTION_JAR
```

Which is looking for the JAR in the target directory, which doesn't exist.

trisberg:
Fixes #347


trisberg:
Rebased and merged as commit 4540bfc4c91988804dbf469940c98fa6fdb17888

PR #377, opened 2018-03-12
Remove DeleteAllOptions struct

trisberg:
- rename func delete to avoid collision with Go's built-in delete function

- fix handling of namespace option in deleteFunctionByName


trisberg:
Rebased after adding Go support

ericbottard:
Rebased and merged as 165b59e5

PR #390, opened 2018-03-14
Try to prevent forgetting to run 'make docs'

ericbottard:
So that we're less likely to forget to update docs.

Also, the js doc is gone, not sure if this is a consequence of the refactoring or if it should have been for a while.


trisberg:
Rebased and merged as commit d48f6b9855451ffc5d03e781fb68467175443a80

PR #391, opened 2018-03-14
Python update

dturanski:
Updates CLI to use Python 3 and cleans up samples


samples/python/upper/README.adoc
```diff
@@ -0,0 +1,31 @@
+= square
```
trisberg: should be "upper" - can fix on merge

samples/python/concat/README.adoc
```diff
@@ -0,0 +1,31 @@
+= square
```
trisberg: Will change to "concat" on merge

dturanski: Thanks I missed a spot

trisberg:
Squashed/Rebased merged as commit b20ff0264908b96f99d2b136819c5b90ea463843 and commit aff04abf09b5a0d086212f1d1b63e974bf18343d

PR #393, opened 2018-03-15
Use approach that is consistent with the rest of flags mgmt for handler.

ericbottard:


riff-cli/cmd/init.go
```diff
@@ -160,18 +159,13 @@ func InitGo(initOptions *options.InitOptions) (*cobra.Command, *options.InitOpti
...
-			err := initializers.Go().Initialize(*initOptions)
-			if err != nil {
-				return err
-			}
-			return nil
+			return initializers.Go().Initialize(*initOptions)
 		},
 	}
-	initGoCmd.Flags().String("handler", "", "the name of the function handler")
+	initGoCmd.Flags().StringVar(&initOptions.Handler, "handler", "", "the name of the function handler (Exported go symbol)")
```
trisberg: What does "Exported go symbol" mean here?

ericbottard: That the handler should be the name of the Exported go symbol in the go plugin

ericbottard: Maybe change to "Exported go function name"?

ericbottard: or "name of Exported go function"?

trisberg: yeah, that sounds better - can change on merge

trisberg:
Rebased and merged as commit 210fe186a1fc518cb24024b448c047cc1dd09c1a

PR #394, opened 2018-03-15
Spike: FunctionInvoker CRD

### LINKED ISSUE #370: Function Invoker CRD
The `riff-cli` currently compiles in support for all known function invokers that enable commands like `riff init` and `riff create`. As new function invokers are introduced, users will have a substandard experience until the cli is updated. A few cases we have already seen where the invokers and cli can drift include:
- brand new invokers (`go`)
- renamed invokers (`shell` -> `command`)
- backwards incompatible language versions (`python2` vs `python3`)
- function upgrades (`0.0.4` -> `0.0.5`)

All of this change is within our own, very small, set of invokers. We also want to encourage community invokers (ruby and kotlin have been explored), and certainly do not want to make `riff-cli` a bottleneck for new invokers.

We need a way for invokers to be discoverable by the `riff-cli`. Each invoker will provide configuration so that the cli can:
- detect which invoker to use for the given artifact by default, user should be able to override
- parameters needed for *-function.yaml (like protocol)
- scaffold the appropriate resources for the function (Dockerfile, .dockerignore, etc)

We have two simple data stores available right now: the filesystem, k8s resource configuration. A major advantage of defining the invokers in k8s is all users access the cluster will have a common set of invokers.

## CRD:

```yaml
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: invokers.projectriff.io
versions:
- name: v1
spec:
  scope: Cluster
  group: projectriff.io
  version: v1
  names:
    kind: Invoker
    plural: invokers
    singular: invoker
  # todo validation
```

This proposal is for a configuration only CRD without any runtime behavior enforced by an invoker-controller. It is only used in this proposal as a central configuration that can be accessed by `riff-cli`, and has no runtime impact on the cluster.

In the future, we may want to make invokers a richer object. For example, the Function resource could default the protocol property to the protocol property of the invoker, or updating an invoker could trigger a update to all functions using that invoker. Nothing in this proposal is meant to limit a richer invoker types in the future.

## Sample Invoker

```yaml
apiVersion: projectriff.io/v1
kind: Invoker
metadata:
  name: node
spec:
  version: 0.0.5-snapshot
  matchers:
  - package.json
  - "*.js"
  properties:
    handler: # not actually used by node now, but maybe with ES Modules
      default: "<default export>"
      description: Name of exported function
    function:
      protocol: grpc
  files:
  - path: Dockerfile
    template: |
      FROM projectriff/node-function-invoker:{{.RiffVersion}}
      {{ if .FileExists("package.json") -}}
      ENV FUNCTION_URI /functions/
      COPY . ${FUNCTION_URI}
      RUN (cd ${FUNCTION_URI} && npm install --production)
      {{- else -}}
      ENV FUNCTION_URI /functions/{{.Artifact}}
      ADD {{.ArtifactBase}} ${FUNCTION_URI}
      {{- end }}
  - path: .dockerignore
    template: |
      node_modules
```

- the handler property indicate if the invoker requires a handler to be specified for the artifact
- `function` and `topics` properties represent the `spec` of the function and topics yaml files
- matchers are globs matching from the target directory or the artifact if specified. Regex may also be useful
- files are templated using the same set of properties currently used in riff-cli, plus methods that allow very basic environment testing:
  - Artifact string
  - ArtifactBase string
  - RiffVersion string
  - Handler string
  - func FileExists(string filename) bool

## CLI usage

When executing the `riff init` or `riff create` commands it should fetch the list of invokers for the cluster. For each invoker and flavor within the invoker, the matcher patterns should be checked. More specific matches (like a full file name, `package.json`) take precedence over partial matches (like a file extension, `*.js`). If there are multiple matches with the same precedence, the user should be prompted to specify the invoker.

Once a match is found, the function.yaml and topics.yaml files should be created using the respective properties merging into the `spec` section the cli currently generates. For file defined, the template should be applied to the properties and written to the file name.

When `riff update` is invoked, the current invoker version for the function should be verified against the latest invoker. If a newer version of the invoker is published, the cli can prompt to upgrade to the newer version.

scothis:
Move logic for scaffolding Function and Topic resources into an Invoker
resource. Invokers are installed as cluster scoped resources and fetched
by the riff-cli to provide invoker specific `riff init` and
`riff create` subcommands.

Invoker resources are currently only used by riff-cli to scaffold
resources for a new function. They is no runtime impact; although this
is likely to change in the future.

Invokers are currently registered with the cluster directly via
`kubectl`:

```
kubectl apply -f path/to/python3-invoker.yaml

kubectl get invokers

kubectl delete invokers python3

kubectl delete invokers --all
```

```
---
apiVersion: projectriff.io/v1
kind: Invoker
metadata:

  # Cluster unique name of the invoker. Used specify the invoker in the
  # `riff init [invoker]` and `riff create [invoker]` commands.
  name: python3

spec:
  # Default Docker image tag for functions using this invoker.
  # Overridden by `--invoker-version`
  version: 0.0.5-snapshot

  # List of file patterns used to autodetect this inovker
  matchers:
  - "*.py"

  properties:

    # Function handler (only needed for languages that need to hint
    # which function to invoke)
    handler:
      # Default handler value
      #
      # Template params:
      # - FunctionName: name of the function
      default: "{{ .FunctionName }}"

      # Documentation for the `--handler` cli property
      description: the name of the function handler

    # Default properties for the *-function.yaml 'spec' key
    function:
      protocol: grpc

    # Default properties for the *-topics.yaml 'spec' key (not common)
    topics:
      partitions: 1

  # List of files to generate (*-topics.yaml and *-functions.yaml are
  # always generated)
  files:
  - # Path within the working directory
    path: Dockerfile

    # Content of the file
    #
    # Template params:
    # - Artifact: path to the artifact
	  # - ArtifactBase: base filename of the artifact (e.g. 'echo.py')
	  # - InvokerVersion: tag of invoker base image
	  # - Handler: function handler (if any)
    # - FileExists: test if a file exists within the working directory
    template: |
      FROM projectriff/python3-function-invoker:{{.InvokerVersion}}
      ARG FUNCTION_MODULE={{.ArtifactBase}}
      ARG FUNCTION_HANDLER={{.Handler}}
      ADD ./{{.ArtifactBase}} /
      {{- if .FileExists "requirements.txt" }}
      ADD ./requirements.txt /
      RUN  pip install --upgrade pip && pip install -r /requirements.txt
      {{- end }}
      ENV FUNCTION_URI file:///${FUNCTION_MODULE}?handler=${FUNCTION_HANDLER}

  # Long form documentation exposed via `riff init [invoker] --help`
  #
  # Template params:
  # - Process: verb describing the command
  # - Command: 'init [invoker]' or 'create [invoker]'
  # - Result: expected output of the command
  doc: |
    {{.Process}} the function based on the function source code specified as the filename, handler,
    name, artifact and version specified for the function image repository and tag.

    For example, type:

        riff {{.Command}} -i words -n uppercase --handler=process

    to {{.Result}}.
```

The `RIFF_DISABLE_KUBECTL` envvar can be set to prevent riff-cli from
loading invokers from the kubernetes cluster.

TODO:
- make shell completion interactive? Currently, the completion scripts
  grow stale after every invoker change
- commands to list, install and remove invokers
- invoker discovery
- built-in invokers

Resolves #370


kubernetes-crds/config/function-invoker-resource.yaml
```diff
@@ -0,0 +1,14 @@
...
+metadata:
+  name: functioninvokers.projectriff.io
+versions:
+- name: v1
+spec:
+  scope: Cluster
+  group: projectriff.io
+  version: v1
+  names:
+    kind: FunctionInvoker
```
markfisher: Since we are already within the scope of the riff namespace, I wonder if "invoker" would be sufficient. It would make things like completion much nicer than if we have 2 of our 3 resources beginning with "function".

scothis: In the initial proposal the type was `Invoker` but I changed it to `FunctionInvoker` since all of our repos are `*-function-invoker`. But, I'd be happy to switch back.

trisberg: Ae they function-invokers for java/node/go/... or are they invokers for java/node/go/... functions? I think the latter sounds better so "Invoker" would be fine.

kubernetes-crds/pkg/apis/projectriff.io/v1/function_invoker.go
```diff
@@ -0,0 +1,93 @@
...
+
+	// Files to generate for this invoker
+	Files []FunctionInvokerFile `json:"files"`
+}
+
+// Properties (invoker properties) for a function invoker
+type FunctionInvokerProperties struct {
+
+	// If true, the invoker requires a handler function name
+	// +optional
```
ericbottard: Any optional property in a CRD ought to be a pointer to primitive type

kubernetes-crds/pkg/apis/projectriff.io/v1/function_invoker.go
```diff
@@ -0,0 +1,93 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// Returned in list operations
+type FunctionInvokerList struct {
+	meta_v1.TypeMeta `json:",inline"`
+	meta_v1.ListMeta `json:"metadata"`
+	Items            []FunctionInvoker `json:"items"`
+}
+
+func SetDefaults_FunctionInvokerSpec(obj *FunctionInvokerSpec) {
```
ericbottard: Set default for eg `handler` here

kubernetes-crds/pkg/apis/projectriff.io/v1/function_invoker.go
```diff
@@ -0,0 +1,93 @@
...
+	// Default topic properties
+	// +optional
+	Topics TopicSpec `json:"topics,omitifempty"`
+}
+
+// File (generated files) for a function invoker
+type FunctionInvokerFile struct {
+
+	// File name to generate
+	Name string `json:"name"`
```
ericbottard: Maybe this would be better named `Path`. I believe it could also require dynamic parts in it, so could be a template itself

riff-cli/cmd/init.go
```diff
@@ -68,110 +69,38 @@ func Init() (*cobra.Command, *options.InitOptions) {
...
+			Use:   functionInvokerName,
+			Short: fmt.Sprintf("Initialize a %s function", functionInvokerName),
+			// Long:  "TODO",
+			RunE: func(cmd *cobra.Command, args []string) error {
+				initOptions.FunctionInvokerName = functionInvokerName
+				return initializer.Initialize(*initOptions)
+			},
+		}
+		if functionInvoker.Spec.Properties.HandlerRequired {
+			initFunctionInvokerCmd.PersistentFlags().StringVar(&initOptions.Handler, "handler", "", "the name of the function handler (required for some function invokers)")
```
ericbottard: This is not a persistent flag, just a regular flag

ericbottard: I believe there would be great value in having the last argument to this call be a property of our CRD (explains what exactly a "handler" is for each language)

riff-cli/cmd/init.go
```diff
@@ -68,110 +69,38 @@ func Init() (*cobra.Command, *options.InitOptions) {
...
+		var initFunctionInvokerCmd = &cobra.Command{
+			Use:   functionInvokerName,
+			Short: fmt.Sprintf("Initialize a %s function", functionInvokerName),
+			// Long:  "TODO",
+			RunE: func(cmd *cobra.Command, args []string) error {
+				initOptions.FunctionInvokerName = functionInvokerName
+				return initializer.Initialize(*initOptions)
+			},
+		}
+		if functionInvoker.Spec.Properties.HandlerRequired {
```
ericbottard: For some invokers, the "handler" could be supported, while not being required (derived from function name for example)

riff-cli/pkg/initializer/function_generator.go
```diff
@@ -0,0 +1,83 @@
...
+	"encoding/json"
+
+	"github.com/ghodss/yaml"
+	crds "github.com/projectriff/riff/kubernetes-crds/pkg/apis/projectriff.io/v1"
+	"github.com/projectriff/riff/riff-cli/pkg/options"
+	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func createFunction(opts options.InitOptions, functionSpec crds.FunctionSpec) ([]byte, error) {
+	functionCRD := crds.Function{
```
ericbottard: Technically, a `functionCR`, not `CRD`

riff-cli/pkg/initializer/initializers.go
```diff
@@ -0,0 +1,119 @@
...
+	err = generateFunctionArtifacts(functionInvoker, opts)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func LoadFunctionInvokers() ([]crds.FunctionInvoker, error) {
+	bytes, err := kubectl.ExecForBytes([]string{"get", "FunctionInvokers", "-o", "json"})
```
ericbottard: I really think it's time to use the go-client lib...

riff-cli/test_data/node/node-function-invoker.yaml
```diff
@@ -0,0 +1,28 @@
...
+      ENV FUNCTION_URI /functions/
+      COPY . ${FUNCTION_URI}
+      RUN (cd ${FUNCTION_URI} && npm install --production)
+      {{- else -}}
+      ENV FUNCTION_URI /functions/{{.Artifact}}
+      ADD {{.ArtifactBase}} ${FUNCTION_URI}
+      {{- end }}
+  - name: .dockerignore
+    template: |
+      node_modules
```
ericbottard: Missing newline

kubernetes-crds/config/function-invoker-resource.yaml
scothis: back to Invoker

kubernetes-crds/pkg/apis/projectriff.io/v1/function_invoker.go
scothis: added a handler default

kubernetes-crds/pkg/apis/projectriff.io/v1/function_invoker.go
scothis: now Path, holding off on making it templated.

riff-cli/cmd/init.go
scothis: done

riff-cli/cmd/init.go
scothis: this would best be handled by making the Handler.Default value templated. Holding off until we have a strong usecase.

ericbottard: The python and go invokers work this way

riff-cli/pkg/initializer/function_generator.go
scothis: done

riff-cli/test_data/node/node-function-invoker.yaml
scothis: done

kubernetes-crds/Makefile
```diff
@@ -55,4 +55,5 @@ hack/glide.lock: hack/glide.yaml
 
 kubectl-apply:
 	kubectl apply -f config/function-resource.yaml
+	kubectl apply -f config/function-invoker-resource.yaml
```
ericbottard: Now renamed `invoker-resource.yaml`

riff-cli/cmd/init.go
```diff
@@ -65,8 +67,8 @@ func Init() (*cobra.Command, *options.InitOptions) {
 	initCmd.PersistentFlags().BoolVar(&initOptions.DryRun, "dry-run", false, "print generated function artifacts content to stdout only")
 	initCmd.PersistentFlags().StringVarP(&initOptions.FilePath, "filepath", "f", "", "path or directory used for the function resources (defaults to the current directory)")
 	initCmd.PersistentFlags().StringVarP(&initOptions.FunctionName, "name", "n", "", "the name of the function (defaults to the name of the current directory)")
-	initCmd.PersistentFlags().StringVar(&initOptions.RiffVersion, "riff-version", utils.DefaultValues.RiffVersion, "the version of riff to use when building containers")
 	initCmd.PersistentFlags().StringVarP(&initOptions.Version, "version", "v", utils.DefaultValues.Version, "the version of the function image")
+	initCmd.Flags().StringVar(&initOptions.InvokerVersion, "invoker-version", "", "the version of invoker to use when building containers")
```
ericbottard: This is not really "when building containers" (ie not `docker build`), but rather initializing the function

trisberg: Since it is part of the `Dockerfile` it is only actually used during the build although it is specified here for the 'init' command that creates the `Dockerfile`

ericbottard: Ah, maybe I'm misunderstanding this then (now that I'm merging the other PR).

Is this the version of the invoker (ie what ends up in the `FROM` clause) or the version of the invoker custom resource? I now understand it is the former, but this is confusing

Really, what this PR deals with is not an *invoker* CRD, but rather an *initer*. Yes, this is a word.

scothis: `invoker-version` is the tag of the docker image in the `FROM` clause. Its default value is defined by `.spec.version` in the invoker.yaml.

kubernetes-crds/Makefile
scothis: fixed

scothis:
@ericbottard much has changed since your initial review. Can you give it another look? Most of the delta is from generated code and test artifacts. The key parts are in the `cmd` and `pkg/initializer` packages.

kubernetes-crds/pkg/apis/projectriff.io/v1/function.go
```diff
@@ -30,8 +30,8 @@ import (
 type Function struct {
 	meta_v1.TypeMeta   `json:",inline"`
 	meta_v1.ObjectMeta `json:"metadata"`
-	Spec               FunctionSpec   `json:"spec"`
-	Status             FunctionStatus `json:"status,omitempty"`
+	Spec               FunctionSpec    `json:"spec"`
+	Status             *FunctionStatus `json:"status,omitempty"`
```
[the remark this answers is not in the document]
scothis: I made this change. Previously, I had to manually nuke the `status` key from the generated function.yaml and topics.yaml. I'd like to leave it unless it causes issues down stream. Not a big deal either way.

ericbottard:
Let's keep it for now. I'm merging the beast

ericbottard:
Merged as 17ebb4dc

PR #398, opened 2018-03-16
Abstract docker interaction behind an interface.

ericbottard:
Capture return value.
Make dry-run use that same interface

This is the first step towards having our tests verify actual
interaction with docker (and maybe doing just that, internal state then
not mattering as much is witnessed side effects are correctly verified).

Next step is to generate a mock for the interface introduced in this PR.
Baby steps.

### DIFF
--- a/riff-cli/pkg/osutils/osutils.go
+++ b/riff-cli/pkg/osutils/osutils.go
@@ -28,7 +28,6 @@ import (
 	"time"
 
 	"github.com/projectriff/riff/riff-cli/pkg/ioutils"
-	"bufio"
 )
 
 func GetCWD() string {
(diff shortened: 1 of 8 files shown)


riff-cli/pkg/osutils/osutils.go
```diff
@@ -88,27 +87,6 @@ func Path(filename string) string {
 	return filepath.Join(strings.Split(path, "/")...)
 }
 
-func ExecWaitAndStreamOutput(cmdName string, cmdArgs []string) {
```
dturanski: This should be a common function for logs as well. That's what I did in the recently closed PR. But we can refactor later

dturanski:
LGTM;

PR #402, opened 2018-03-17
Set --namespace option for publish as deprecated

trisberg:
- print a note about the deprecation

- this will help with older demo scripts since they would fail if we didn't keep the option


ericbottard:
Merged as 79b17df074757a285932c7f12bd436c77940118d and parent (forgot to squash).

Used cobra native way of marking depreaction

trisberg:
Nice, didn't know that was available

Issue #321, opened 2018-03-01
add support for event source functions


ericbottard:
_From @markfisher on January 25, 2018 16:15_

an event source function would take lifecycle events as input, and emit events as output

this will require sidecar-generated metrics based on output activity and possibly other ways to indicate activity

this would also support "task" functions triggered by a scheduler (or potentially launched as cron-jobs) that terminate normally when complete (or when a "cancel" event is received on the input topic)

(idleTimeout could still apply)

one candidate for a first event source function would be one that listens/republishes Kubernetes events on its output topic

_Copied from original issue: projectriff/function-controller#8_

ericbottard:
_From @markfisher on February 14, 2018 22:41_

an initial PoC could just be a Function with a high value for `idleTimeout` that's triggered based on a lifecycle event or something like a query string (and/or could keep it alive by pinging within the `idleTimeout` period).

ericbottard:
Don't we want to introduce `minReplicas` anyway, I think that would make more sense.

Including towards making the `http-gametaway` a function itself

dturanski:
@markfisher "polling a source function could also be explicit. Send a message to the input topic as the trigger"

scothis:
This issue has gone stale, we're on Knative now! Please reopen if still relevant.

PR #403, opened 2018-03-19
Add "supplier" pattern to function.proto

dsyer:
User sends a Trigger, and server sends back a stream of messages.

See #321

### DIFF
--- a/function-proto/function.proto
+++ b/function-proto/function.proto
@@ -26,6 +26,13 @@ message Message {
 	map<string, HeaderValue> headers = 2;
 }
 
+message Trigger {
+}
+
 service MessageFunction {
   rpc Call(stream Message) returns (stream Message) {}
+}
+
+service MessageSource {
+  rpc Call(Trigger) returns (stream Message) {}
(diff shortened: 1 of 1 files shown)


scothis:
@dsyer how does the function-sidecar know which service to call the invoker with?

dsyer:
It doesn't. There's no support for this in the sidecar yet. Something we need to discuss.

function-proto/function.proto
```diff
@@ -26,6 +26,13 @@ message Message {
...
+message Trigger {
+}
+
 service MessageFunction {
   rpc Call(stream Message) returns (stream Message) {}
 }
+
+service MessageSource {
+  rpc Call(Trigger) returns (stream Message) {}
+}
```
dturanski: new line

dsyer: Fixed. Who cares though?

dsyer:
Closing because I don't think we need it (the RPC call itself is enough of a trigger for the supplier to start).

PR #412, opened 2018-03-19
Fix python samples links

### LINKED ISSUE #411: Python sample project links are broken
`sentiments` doesn't exist, and `concat` and `upper` are not mentioned

mcowger:
* sentiments doesn't exist, and concat and upper are not mentioned.  Removed sentiments and added concat/upper.  Fixes #411


trisberg:
LGTM, thanks!

PR #418, opened 2018-03-20
configurable sidecar image name

jldec:
function controller looks at environment variable RIFF_FUNCTION_SIDECAR_IMAGE 
uses that for the sidecar image config - still defaulting to "projectriff/function-sidecar"

also updated/tested helm yaml and non-helm config, and values.yaml.

### DIFF
--- a/function-controller/pkg/controller/deployer.go
+++ b/function-controller/pkg/controller/deployer.go
@@ -93,7 +93,11 @@ func (d *deployer) buildMainContainer(function *v1.Function) corev1.Container {
 
 func (d *deployer) buildSidecarContainer(function *v1.Function) corev1.Container {
 	c := corev1.Container{Name: "sidecar"}
-	c.Image = sidecarImage + ":" + os.Getenv("RIFF_FUNCTION_SIDECAR_TAG")
+	imageName := os.Getenv("RIFF_FUNCTION_SIDECAR_IMAGE")
+	if imageName == "" {
+		imageName = sidecarImage
+	}
+	c.Image = imageName + ":" + os.Getenv("RIFF_FUNCTION_SIDECAR_TAG")
 	outputDestination := function.Spec.Output
 	if outputDestination == "" {
 		outputDestination = "replies"
--- a/function-controller/pkg/controller/deployer.go
+++ b/function-controller/pkg/controller/deployer.go
@@ -93,7 +93,7 @@ func (d *deployer) buildMainContainer(function *v1.Function) corev1.Container {
 
 func (d *deployer) buildSidecarContainer(function *v1.Function) corev1.Container {
 	c := corev1.Container{Name: "sidecar"}
-	imageName := os.Getenv("RIFF_FUNCTION_SIDECAR_IMAGE")
+	imageName := os.Getenv("RIFF_FUNCTION_SIDECAR_REPOSITORY")
 	if imageName == "" {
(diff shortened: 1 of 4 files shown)


ericbottard:
What's the endgame here? Why would we want to override the sidecar with something not provided by us?

jldec:
this is necessary for non-local (GKE) installs for users without access to the prjectriff org on docker hub E.g. devlopers installing a forked build with their own docker id or using a docker registry like gcr.

jldec:
riff today can **only** be operated pulling the sidecar image from our docker hub repo.

ericbottard:
I see. Thanks for the clarification

jldec:
no problem @ericbottard, thanks for asking.
any objections to merging this?

trisberg:
we should add it to the helm chart values.yaml and the template as well

function-controller/pkg/controller/deployer.go
```diff
@@ -93,7 +93,11 @@ func (d *deployer) buildMainContainer(function *v1.Function) corev1.Container {
 
 func (d *deployer) buildSidecarContainer(function *v1.Function) corev1.Container {
 	c := corev1.Container{Name: "sidecar"}
-	c.Image = sidecarImage + ":" + os.Getenv("RIFF_FUNCTION_SIDECAR_TAG")
+	imageName := os.Getenv("RIFF_FUNCTION_SIDECAR_IMAGE")
+	if imageName == "" {
+		imageName = sidecarImage
+	}
+	c.Image = imageName + ":" + os.Getenv("RIFF_FUNCTION_SIDECAR_TAG")
```
trisberg: should this default to "latest" if the env var is empty/missing?

trisberg:
In Helm terms this should be called the `image.repository` which combined with the `image.tag` defines the image to use

function-controller/pkg/controller/deployer.go
```diff
@@ -93,7 +93,11 @@ func (d *deployer) buildMainContainer(function *v1.Function) corev1.Container {
 
 func (d *deployer) buildSidecarContainer(function *v1.Function) corev1.Container {
 	c := corev1.Container{Name: "sidecar"}
-	c.Image = sidecarImage + ":" + os.Getenv("RIFF_FUNCTION_SIDECAR_TAG")
+	imageName := os.Getenv("RIFF_FUNCTION_SIDECAR_IMAGE")
```
trisberg: I would call this  RIFF_FUNCTION_SIDECAR_REPOSITORY to be consistent with Helm usage

jldec: ok - done - consistency is good (although i always found the use of the "repository" term on dockerhub confusing)

function-controller/pkg/controller/deployer.go
jldec: I'm not a big fan of hardwiring "latest" with it's magic behavior in k8s into the code.
Do you know if there is a way to default to the same tag as the running (function-controller) image?
Since we build all the core images together it makes sense to default to the same version.

trisberg:
Squashed and merged as commit 9b83e69fabd2e7e44f1cb4f1215044bbee6b49f7

function-controller/pkg/controller/deployer.go
trisberg: We do that for the Helm chart, so let's just leave as is for now

PR #423, opened 2018-03-22
Fix rot13 sample

trisberg:


ericbottard:
OMG! Never roll your own crypto!

PR #424, opened 2018-03-22
Add notes on how to upgrade to new release

dsyer:


Getting-Started.adoc
```diff
@@ -152,6 +152,13 @@ The provided Helm chart installs the following components:
 
 * HTTP Gateway - the external facing gateway that listens for data posted to the various topics and forwards the data as a message for the corresponding Kafka topic
 
+=== To upgrade to a new version
+
+[source, bash]
+----
+helm upgade demo riffrepo/riff
```
trisberg: This belongs in  the Getting-Started doc. We also switched from naming the release "control" instead of "demo". We should include upgrade instructions once things stabilize a bit more between release, maybe once we release a "beta" version.

PR #438, opened 2018-03-22
Add Makefile targets for kafka, namespace and teardown

jchesterpivotal:
Intended to make the development experience more consistent.


Makefile
```diff
@@ -50,3 +66,4 @@ clean:
 	$(MAKE) -C topic-controller		clean
 	$(MAKE) -C riff-cli				clean
 
+dev-setup: namespace development-kafka clean build dockerize kubectl-apply
```
ericbottard: I'm not too keen on so many fine grained tasks that are only aliases for one kubectl command. What about moving the namespace and kafka parts inside kubectl-apply?

jchesterpivotal: I don't mind at all. I figured I'd start with a straight translation.

Would it be worth renaming `kubectl-apply` in that case? (To what I'm not sure, as I haven't had my 2nd coffee yet)

ericbottard: `dev-setup`? then it neatly mirrors `teardown`

jchesterpivotal: Works for me. I'll push it up in a few.

Makefile
```diff
@@ -30,13 +30,25 @@ debug-dockerize:
 	$(MAKE) -C http-gateway			debug-dockerize
 	$(MAKE) -C topic-controller		debug-dockerize
 
-kubectl-apply:
+dev-setup: clean dockerize
```
ericbottard: I wouldn't put clean dockerize as dependencies.

dev-setup and teardown need to be declared as PHONY

ericbottard:
Merged as 75bd4bb3

PR #453, opened 2018-03-26
Better artifact resolution for specified invoker

### LINKED ISSUE #448: CLI file matcher should respect explicit 'riff create <invoker>'
This is related to https://github.com/projectriff/command-function-invoker/issues/10 but applies to the riff CLI in general, not just the command invoker.

After installing the [command function invoker yaml](https://github.com/projectriff/command-function-invoker/blob/master/command-invoker.yaml), all matchers stop working when used with `riff create` or `riff init`.

**Suggestion:** Specifying an invoker explicitly with say `riff create node` or `riff init node` should force matching on just node-invoker file extensions, and is more consistent and easier than using `--artifact` which might still be ambiguous across invokers.

scothis:
When an invoker is specified, we only need to search for an artifact
matching the specified invoker rather than looking at all invokers.

Fixes #448

### DIFF
--- a/riff-cli/cmd/init_test.go
+++ b/riff-cli/cmd/init_test.go
@@ -85,6 +85,22 @@ var _ = Describe("The init command", func() {
 			Expect(".").NotTo(HaveUnstagedChanges())
 		})
 
+		It("should find an artifact based on an invoker", func() {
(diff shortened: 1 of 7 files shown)


ericbottard:
Merged as b5d44dc1

PR #456, opened 2018-03-26
Make it so that make dev-setup can be run multiple times

ericbottard:
Create namespace is the only command that is not idempotent. Ignore failures


trisberg:
LGTM

Issue #454, opened 2018-03-26
Add riff init and riff create docs for each invoker


scothis:
Now that invoker subcommands for `riff init` and `riff create` are decoupled from the riff cli core, the docs for these commands no longer appears in this repository. The docs for `riff init [invoker]` and `riff create [invoker]` should appear in the repository that defines the invoker.

We'll need the ability to:
- [x] provide a stub set of invokers for riff-cli, avoiding a kubectl lookup (#457)
- [x] generate docs for a single command (#464)

Add the command to generate docs to each invoker repo:
- [x] command-function-invoker (projectriff/command-function-invoker/pull/11)
- [x] go-function-invoker (https://github.com/projectriff/go-function-invoker/pull/6)
- [x] java-function-invoker (https://github.com/projectriff/java-function-invoker/pull/60)
- [x] node-function-invoker (https://github.com/projectriff/node-function-invoker/pull/43)
- [x] python2-function-invoker (https://github.com/projectriff/python2-function-invoker/pull/7)
- [x] python3-function-invoker (https://github.com/projectriff/python3-function-invoker/pull/5)

PR #464, opened 2018-03-26
Support generation of docs for a command subtree

scothis:
Invokers should be able to genreate docs for the `riff init [invoker]`
and `riff create [invoker]` commands without generating docs for *all*
riff commands.

    # generate docs for `riff init command` and `riff create command`
    $ RIFF_INVOKER_PATHS="command-invoker.yaml" riff docs -d docs --command "init command"
    $ RIFF_INVOKER_PATHS="command-invoker.yaml" riff docs -d docs --command "create command"

Refs #454

### DIFF
--- a/riff-cli/Makefile
+++ b/riff-cli/Makefile
@@ -1,5 +1,5 @@
 .PHONY: build clean test release verify-docs all gen-mocks
-OUTPUT = riff
+OUTPUT = ./riff
 OUTPUT_WINDOWS = $(OUTPUT).exe
 GO_SOURCES = $(shell find cmd pkg -type f -name '*.go' -not -name 'mock_*.go')
 VERSION ?= $(shell cat ../VERSION)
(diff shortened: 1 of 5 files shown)


ericbottard:
Merged as ca1bb9e2, which includes polish found in https://github.com/ericbottard/riff/commit/ad8b955d736d8bc01e47949b6f00b7bdc56361d2

I'm not sure I understand the test addition though? Where is this actually used? Is this gonna be used in invoker projects (and the test is just that: a test?)

scothis:
@ericbottard the intent is for invokers to create docs in their repo for the `init` and `create` commands using the local `*-invoker.yaml` file. The test in the Makefile felt more natural than using a go test since that's how the command is expected to be used.

ericbottard:
OK, that's what I understood then. Carry on :) 👍

PR #465, opened 2018-03-26
Disable glide on travis

scothis:
- Centralize the 'vendor' task into the root Makefile
- ifdef the vendor task to be empty on CI


scothis:
Will recreate this PR from a remote repo

PR #466, opened 2018-03-26
Remove glide on travis

scothis:
touch the vendor directory to avoid cases where make wanted to process
that target.


ericbottard:
Merged as 7b999cb6

Issue #470, opened 2018-03-27
Factor out autoscaler with current algorithm


glyn:
Wrap the Kafka offset measurement in a `QueueLength` API and package the current autoscaling algorithm using the autoscaler design from PR https://github.com/projectriff/riff/pull/399.

This will enable the behaviour of the current algorithm to be examined with a simulated workload which will provide a baseline for future autoscaler algorithm improvements.

glyn:
Since the current algorithm depends on partition offsets rather than just queue length, it won't be possible to reproduce the current behaviour exactly.

glyn:
Implemented in https://github.com/projectriff/riff/pull/505

PR #475, opened 2018-03-28
Metrics infrastructure

glyn:
Metrics need to be introduced in order to avoid depending on Kafka offsets in scaling algorithms. This PR does not impact external behaviour but reduces the size of the autoscaling branch.

Part of https://github.com/projectriff/riff/issues/470


ericbottard:
Merged as 3 separate commits ending in f69a0802

Issue #477, opened 2018-03-28
Add kafka as a conditional dependency requirement


trisberg:
This will avoid having to install a separate transport chart. We should also look into overriding the name of the kafka-broker/zookeeper services with a fixed DNS name like `kafka-broker.riff-system.svc.cluster.local`. We could add a `transport.enabled` or `kafka.enabled` switch with true as the default.

trisberg:
This relates to #449

PR #481, opened 2018-03-28
Add kafka chart as a conditional dependency to riff chart

trisberg:
- set projectriff/kafka dependency to version `0.0.2`

- new flag to turn on the kafka install: `--set kafka.create=true` (default is false, so this makes it opt-in as suggested)

- this eliminates the separate transport release

- changing the release name to be projectriff instead of control

- updating Getting-Started and Development-Helm-install docs

Fixes: #477

### DIFF
--- a/helm-charts/riff/Chart.yaml
+++ b/helm-charts/riff/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: riff is for functions - a FaaS for Kubernetes
 name: riff
-version: 0.0.6-snapshot
+version: 0.0.0
 appVersion: 0.0.0
 home: https://projectriff.io/
 icon: https://raw.githubusercontent.com/projectriff/projectriff.io/master/images/riff.png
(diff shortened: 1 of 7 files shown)


trisberg:
Yeah, I think I agree. Will amend the PR.

trisberg:
Rebased and addressed review comments

helm-charts/riff/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 description: riff is for functions - a FaaS for Kubernetes
 name: riff
-version: 0.0.0
+version: 0.0.6-snapshot
```
scothis: Do we need a non-0.0.0 version?

If so, can we use a version that doesn't need to be manually maintained and instead is replaced within the helm-publish script?

trisberg: Oops, I do that when testing locally, I guess we could use 9.9.9 or something similar

PR #491, opened 2018-03-29
Polish Invoker CRD

scothis:
- `spec.metadata.handler` -> `spec.handler`
- `spec.metadata.function` -> `spec.functionTemplate.spec`
- `spec.metadata.topics` -> `spec.topicTemplate.spec`

This makes it more clear that the function and topic key are templateing
other k8s resources. Other (non-spec) properties can now be included in
the template, like metadata.labels.


ericbottard:
LGTM

PR #495, opened 2018-03-30
Set the riff chart version to a non-zero value

trisberg:
- this will allow us to lint and install the chart from the local file system for testing


trisberg:
This was addressed in https://github.com/projectriff/riff/commit/7df5091

Issue #499, opened 2018-03-30
The Getting-Started guide doesn't work with current release (0.0.5)


trisberg:
If someone follows the instructions in master they will not get a working riff install since the docs depend on new features that will be in 0.0.6 while the `helm install` command will pull the 0.0.5 chart.

We can:

- [x] Revert the Getting-Started guide to a version that works with chart version 0.0.5 (current release)
- [x] Create a PR for updating the Getting-Started guide to the 0.0.6 instructions 
- [x] Add additional commits for the `Getting-Started.adoc` to the `trisberg-issue-499-doc-for-006` branch.
- [x] Merge that [PR](https://github.com/projectriff/riff/pull/501) just before the next release

scothis:
The getting started guide in the repo should always be accurate for the content of the current commit (or as close as practical). For the last stable release we should refer people to the website.

trisberg:
Since it relies on published Helm chart it shouldn't be updated until we are ready for the release and that chart is available, otherwise someone stumbling on our docs could have a sub-par first experience like we saw.

scothis:
We also publish snapshot helm charts where the new docs are relevent. In general, we need to be more clear about whether the docs are targeting a stable release or the current content of the repo.

trisberg:
The Gettting-Started guide only references released version and no snapshots (no --devel flag used). That part is in the Development-Helm-install doc which now works for 0.0.6-snapshots and is current.

PR #501, opened 2018-03-30
[DO NOT MERGE YET] Updating the Getting-Started guide to the 0.0.6 instructions

trisberg:
This should be merged just before the 0.0.6 release so that instructions in master matches the latest release (.0.0.5)

Add additional commits for the `Getting-Started.adoc` to this `trisberg-issue-499-doc-for-006` branch.

Task of #499

### DIFF
--- a/Getting-Started.adoc
+++ b/Getting-Started.adoc
@@ -61,29 +61,24 @@ helm search riff -l
 
 == [[current]]Install the current riff release
 
-We have separated the Kafka installation from the riff chart. This requires an extra step of installing Kafka in order to use riff. You can use the single-node Kafka chart provided by riff or the three-node kafka/zookeeper service provided by Kubeapps.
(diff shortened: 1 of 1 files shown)


Getting-Started.adoc
```diff
@@ -61,36 +61,40 @@ helm search riff -l
...
-Create the `riff-system` namespace:
+To create the `riff-system` namespace run:
 
 [source, bash]
 ----
 kubectl create namespace riff-system
 ----
 
-Install the kafka/zookeeper service provided by riff using:
+[TIP]
```
jldec: I think we should move this whole tip down to a separate section (not under `Create riff-system namespace`)
Most of our users won't run the multi-node kafka.

We should also avoid mixing wide code blocks with text in tips becuase they make the text scroll sideways. Use tips just for the text, and keep the code block outside of the tip.

screen shot 2018-04-06 at 1 56 13 pm

PR #505, opened 2018-04-03
Factor out current autoscaler

glyn:


glyn:
@ericbottard With @jldec's help, I ran the S1P demo agains this branch and it seemed to function as expected.

PR #512, opened 2018-04-04
Fail CLI in case of http error.

### LINKED ISSUE #373: `riff publish` silently ignores http errors
If the http-gateway returns an error, like a 404, `riff publish` will exit normally. This makes it impossible to know the difference between a function that failed to be invoked and a function that returned no output.

ericbottard:
Fixes #373

### DIFF
--- a/riff-cli/cmd/publish.go
+++ b/riff-cli/cmd/publish.go
@@ -167,7 +167,7 @@ func doPost(url string, publishOptions publishOptions) (string, error) {
 	if 200 <= resp.StatusCode && resp.StatusCode < 400 {
 		return string(body), nil
 	} else {
-		return "", fmt.Errorf("Gateway responded with code %v: %v", resp.StatusCode, string(body))
+		return "", fmt.Errorf("HTTP Gateway responded with code %v: %v", resp.StatusCode, string(body))
 	}
 
 }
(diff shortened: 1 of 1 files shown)


riff-cli/cmd/publish.go
```diff
@@ -61,6 +61,7 @@ will post '{"hello":"world"}' as json to the 'concat' topic and wait for a reply
 
 `,
 		RunE: func(cmd *cobra.Command, args []string) error {
+			cmd.SilenceUsage = true
```
scothis: Do we always want to silence usage, or only when we have a known http error?

ericbottard: As a matter of convention, we should often silence usage as soon as we enter `runE`.
Misconfiguration errors (that should then show usage) are normally handled in `preRunE`, but when we reach `runE` I think it's safe to assume we understood what the user wanted

riff-cli/cmd/publish.go
```diff
@@ -157,11 +158,16 @@ func doPost(url string, publishOptions publishOptions) (string, error) {
...
+
 	body, err := ioutil.ReadAll(resp.Body)
 	if err != nil {
 		return "", err
 	}
-	return string(body), nil
+	if 200 <= resp.StatusCode && resp.StatusCode < 400 {
+		return string(body), nil
+	} else {
+		return "", fmt.Errorf("Gateway responded with code %v: %v", resp.StatusCode, string(body))
```
glyn: "Gateway" could mean lots of things in a network. Perhaps change to "HTTP gateway".

ericbottard: Will do.

OK to merge afterwards?

glyn:
merged as e57aec4482c5f7ed6bf2b9254b30dd3908e89111

PR #513, opened 2018-04-04
Show 404s for missing topics

### LINKED ISSUE #410: Gateway should return 404 when topic does not exist
```
$ curl localhost:30365/messages/foo -i -d"hello"
HTTP/1.1 200 OK
Date: Mon, 19 Mar 2018 20:07:24 GMT
Content-Length: 32
Content-Type: text/plain; charset=utf-8

Message published to topic: foo

$ kubectl get topic foo
Error from server (NotFound): topics.projectriff.io "foo" not found
```

jchesterpivotal:
Intended to resolve #410.

These changes add topic-awareness to the `http-gateway`, so that it returns 404 errors for messages or requests that refer to non-existent topics.

### DIFF
--- a/http-gateway/cmd/http-gateway.go
+++ b/http-gateway/cmd/http-gateway.go
@@ -53,7 +53,12 @@ func main() {
 	}
 	defer consumer.Close()
 
-	gw := server.New(8080, producer, consumer, 60*time.Second)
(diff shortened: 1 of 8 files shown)


ericbottard:
Merged as b8deeee7

PR #515, opened 2018-04-04
riff is for lowercase

markfisher:


ericbottard:
Merged as 2be17d72 then deleted this PR branch which was on `projectriff`

markfisher:
sorry for that branch... I'm still getting used to having "origin" for upstream instead of my fork

ericbottard:
You can always override after the fact. Speaking of that, seems like this org's repo is a branch graveyard. Maybe we want to clean things up a bit and only keep things release-related?

PR #518, opened 2018-04-05
Override init and create usage template so that it uses "invoker" instead of "command".

### LINKED ISSUE #502: Require invoker name for 'riff create' and 'riff init' and clarify usage in help
The CLI UX would be simpler if an invoker parameter were required for all `riff create <invoker>` and `riff init <invoker>` commands. If no invoker is specified, display a usage error and the list of registered invokers.

this is a followup to #448 

1. Auto-detecting invokers from file extensions introduces non-determinism. E.g. adding new invokers could break old workflows by stealing the extension. This is true even when pointing to a source file with --artifact.

2. The current command-invoker already makes it necessary _always_ to disambiguate, because it matches *. 

3. It's often eaiser to specify an invoker than point to an "artifact" file (invoker names are short).

4. with this approach we could even auto-generate the source file when there are no files at all in the directory. E.g. `riff create node --stream`

ericbottard:
Fixes #502 

Fixes https://github.com/projectriff/riff/pull/518

### DIFF
--- a/riff-cli/Makefile
+++ b/riff-cli/Makefile
@@ -27,7 +27,7 @@ release: $(GO_SOURCES) ../vendor
 	GOOS=windows  GOARCH=amd64 go build -ldflags "-X main.version=$(VERSION)" -o $(OUTPUT_WINDOWS) ../cli.go && zip -mq riff-windows-amd64.zip $(OUTPUT_WINDOWS)
 
 docs: $(OUTPUT)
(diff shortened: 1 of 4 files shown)


markfisher:
should we move the command invoker into the core repo now?

ericbottard:
I did too. But the markdown docs don't use the same templates apparently

PR #532, opened 2018-04-11
Change scaler interval to 97ms

jchesterpivotal:
By inspiration from "Systems Performance"[0] by Brendan Gregg and by performance-sampling tools that book describes (eg lockstat[1]). A 97ms interval won't lockstep with processes running on 10/100/1000ms
boundaries, such as some OS schedulers.

[0] Page 219.
[1] https://docs.oracle.com/cd/E19253-01/816-5166/lockstat-1m/index.html


glyn:
Merged as 66bb94ae40adee320467d

Issue #488, opened 2018-03-29
init requires -f <dir> - inconsistent with create 


dturanski:
Currently `init` requires `-f <dir>` and does not interpret the arg as a directory path as create does.  Need to add `utils.AliasFlagToSoleArg("filepath")` for init.  Also test that the path does not collide with an invoker name.

scothis:
It's safer to always include the flag. While commands are greedy and will be matched before arguments, if your path happens to collide with an invoker that's installed you could get a surprise when the command starts to fail.

ericbottard:
Then we need to remove that behavior from `create` IIUC

ericbottard:
Note that if we do https://github.com/projectriff/riff/issues/502, then there won't be any ambiguity as `riff init` and `riff create` (w/o an invoker) will be gone

scothis:
`riff create` is picking up this behavior from the `build` command argument. We can add the same first-arg-to-filepath mapping for the `init` command to make the behavior consistent between `init` and `create`.

PR #535, opened 2018-04-11
Align init and create command args

scothis:
`riff create <invoker> <filepath>` is also handled as
`riff create <invoker> --filepath <filepath>`. Previously, the equivlent
with `riff init` was also not true. Now the first argument is treated
uniformly as the filepath flag for both the init and create commands.

Issue: #488

### DIFF
--- a/riff-cli/cmd/create.go
+++ b/riff-cli/cmd/create.go
@@ -31,6 +31,16 @@ func Create(initCmd *cobra.Command, buildCmd *cobra.Command, applyCmd *cobra.Com
 	createChainCmd.Long = utils.CreateCmdLong()
 	createChainCmd.SetUsageTemplate(utils.CustomInvokerUsageTemplate)
 
+	// ignore all validation
+	//
+	// This command will fail since no invoker is specified. The init command
+	// is able to provide a more meaningful error message. We still want to
+	// use a chained command to establish persistent flags
+	createChainCmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error { return nil }
+	createChainCmd.PersistentPreRun = func(cmd *cobra.Command, args []string) {}
+	createChainCmd.PreRunE = func(cmd *cobra.Command, args []string) error { return nil }
+	createChainCmd.PreRun = func(cmd *cobra.Command, args []string) {}
+
 	return createChainCmd
 }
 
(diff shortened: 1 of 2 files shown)


riff-cli/cmd/create.go
```diff
@@ -31,6 +31,16 @@ func Create(initCmd *cobra.Command, buildCmd *cobra.Command, applyCmd *cobra.Com
 	createChainCmd.Long = utils.CreateCmdLong()
 	createChainCmd.SetUsageTemplate(utils.CustomInvokerUsageTemplate)
 
+	// ignore all validation
+	//
+	// This command will fail since no invoker is specified. The init command
+	// is able to provide a more meaningful error message. We still want to
+	// use a chained command to establish persistant flags
+	createChainCmd.PersistentPreRunE = func(cmd *cobra.Command, args []string) error { return nil }
```
ericbottard: Not sure I understand this comment.

Who's "this" in the sentence "this command will fail"? Is it `createChainCmd`? If we have a command that will fail, should we have a command in the first place?


Persistence of flags has to do with "inheritance" of commands. Chaining is more akin to "composition" of commands.
I'm more and more thinking that our implementation of chaining brings more complexity than actually needed.

scothis: `createChainCmd` is the command that will fail, since it doesn't have an invoker specified. The purpose of the command is twofold:
- provide a good error message when a user executes the command, either by omitting an invoker or specifying an invoker that is not installed
- generate command docs that aggregate behavior across invokers

I agree it's a bit funny to define a command that we know will never do real work. If there's another way to accomplish these goals, I'm in.

ericbottard:
Well, for first item, I'd argue that it's cobra's job to provide an error when the user invokes a command that doesn't exist (because it's just that, given our usage of 1 command per invoker: `riff create bogus` is *not* a command). I see what you mean by "good" message, but it should be consistent with other wrong usages.


Again, I feel that we're more and more abusing the tool, for very small benefits. If I understand correctly, the only 2 reasons we're not having a (single) `riff create` command that accepts an arg being the invoker name are
- `--handler` is optional
- documentation of particular invokers (in particular, "meaning" of handler)

I get it, but seems the cost is getting higher, while the benefits are rather thin...

scothis:
@ericbottard I agree with your comments. If we run into further turbulence, we should rethink the approach.

PR #543, opened 2018-04-16
Link to invoker repos for samples

glyn:
@dsyer suggested that it would be helpful to link to the samples. This is a step in that direction while still acknowledging that 3rd parties can write invokers which are just as valid as those provided in the riff project.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -82,7 +82,8 @@ make teardown
 
 == [[samples]]Try Some Samples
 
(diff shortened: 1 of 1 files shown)


dsyer:
LGTM. Presumably `https://github.com/projectriff?q=-invoker` would work as well?

glyn:
Yes, cleaned up the URL.

Issue #323, opened 2018-03-01
Sidecar panics if function invoker not available


ericbottard:
_From @dsyer on February 19, 2018 15:43_

Not sure if this is an issue, but it's easy to reproduce: just launch a sidecar and a function invoker on localhost, and then kill the invoker. The sidecar panics and never recovers. Seems like there might be a similar failure mode even in a healthy pod in k8s.

_Copied from original issue: projectriff/function-sidecar#53_

jldec:
invoker healthchecks?

scothis:
This issue has gone stale, we're on Knative now! Please reopen if still relevant.

PR #544, opened 2018-04-17
Add backoff loop

dsyer:
Really unsophisticated solution to #323, just backs off for 1 second and
makes the RPC call again. If the function invoker dies
(or is replaced with a telepresence) then the sidecar will retry until it
succeeds.

...unless the command line flag --exitOnComplete=true, in which case it
simply exits when the RPC call completes. This is super useful for event
source and sink use cases (Consumer and Supplier in Java terms).

@ericbottard: I know this probably interferes with work you might be doing
on windowing. So please just close this PR if you have something more 
sophisticated.


glyn:
I agreed with @dsyer that this PR should not be merged until @ericbottard has looked at it. I wanted to emphasise this in case others review it and might be tempted to merge.

dturanski:
see https://github.com/dsyer/riff/pull/1

dsyer:
@ericbottard says he is not working on this, so it's OK to merge from my point of view. I don't think it addresses all the issues with connecting/failing to connect to grpc, but it does add the option to shutdown when the grpc call ends (which is important for event sources).

function-sidecar/cmd/function-sidecar.go
```diff
@@ -108,31 +111,75 @@ func main() {
...
+
+		carrier.Run(consumer, producer, dispatcher, output)
+
+		select {
+		case <-signals:
+			log.Println("Shutting Down...")
+			break LOOP
+		case <-dispatcher.Closed():
+			log.Println("End of Stream...")
+			if !backoff() {
```
glyn: Why back off after end of stream and potentially create another dispatcher? That doesn't seem related to https://github.com/projectriff/riff/issues/323.

dsyer: It's a possibly feeble attempt to address #323. The problem there is that when the invoker either finishes the RPC or just goes away the sidecar sits and does nothing. At least if you try to create another dispatcher, it might be able to have another crack at it. I can't say it's very successful, yet. Hopefully it makes it easier to see how to fix it properly for someone who knows more than I do.

function-sidecar/cmd/function-sidecar.go
```diff
@@ -108,31 +111,75 @@ func main() {
...
+}
+
+func backoff() bool {
+	if exitOnComplete {
+		return false
+	} else {
+		// Back off a bit to give the invoker time to come back
+		// (if we support windowing or polling this logic will be more complex)
+		time.Sleep(1000 * time.Millisecond)
+		return true
```
glyn: Shouldn't the back-off eventually give up? Otherwise we might be stuck in this cycle indefinitely.

dturanski:
Close in favor of #563, since it was easier for me to add a commit to this and issue a new PR.

PR #546, opened 2018-04-17
Make k8s resource checks in http gateway optional

dsyer:
You only need to be paranoid about topic existence if you are running
in an actual cluster. Running locally (like it says you can do in the
docs) there is no need to check for topic existence in k8s because
there is no k8s.

### DIFF
--- a/http-gateway/cmd/http-gateway.go
+++ b/http-gateway/cmd/http-gateway.go
@@ -56,15 +56,15 @@ func main() {
 	defer consumer.Close()
 
 	restConf, err := rest.InClusterConfig()
-	if err != nil {
-		panic(err)
+	var riffClient *versioned.Clientset
+	if err == nil {
+		riffClient, err = versioned.NewForConfig(restConf)
+		if err != nil {
+			panic(err)
+		}
+	} else {
+		riffClient = nil
 	}
-
-	riffClient, err := versioned.NewForConfig(restConf)
-	if err != nil {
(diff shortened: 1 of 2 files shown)


jchesterpivotal:
I guess two comments after a brief skim.

First is we should be able to add a testcase that doesn't rely on providing a stub that fits the `RiffTopicExistenceChecker` interface to allow the tests to run outside k8s. Which is nice, it's a slightly more realistic test.

Second is that I'd prefer that we provide a second implementation of the interface that always gives "true" than to rely on `nil`s to signal different policies. So instead of the alternatives being the current checker and `nil`, it'd be the current checker and `AlwaysTrueExistenceChecker`.

jchesterpivotal:
So it's less about detecting the k8s/non-k8s profile by reading the `err` of `client.InClusterConfig()`, which I think would remain as-is. It's that in the non-k8s case, setting the `riffClient` to `nil`, passing it forward and then `nil`-checking on the far side might be replaced by moving the creation of `RiffTopicExistenceChecker` into the k8s case and introducing a `AlwaysTrueExistenceChecker` for the non-k8s case.

Does that make more sense?

jchesterpivotal:
On second thoughts, the code looks like it will work and could be updated in a later PR, I'm going to pull it in.

dsyer:
Thanks. I'll refactor to `AlwaysTrueExistenceChecker` as an exercise.

PR #550, opened 2018-04-20
Adopt a common way of deferring close

glyn:
For optional io.Closer implementations, it's cleaner to test for io.Closer
before deferring the close.


glyn:
Thanks @jchesterpivotal!

Issue #553, opened 2018-04-21
Topic checker reduces http-gateway throughput


jldec:
A test was performed by calling the square function using apache bench, with and without https://github.com/projectriff/riff/pull/552 which disables the topic checker.

```
 ab -H "Content-Type: text/plain" -n 1000 -c 10 -p 11.txt http://127.0.0.1:31561/requests/sq

 # topic checker enabled:
Requests per second:    5.06 [#/sec] (mean)
Time per request:       1977.061 [ms] (mean)
Time per request:       197.706 [ms] (mean, across all concurrent requests)

# topic cheker disabled
Requests per second:    537.80 [#/sec] (mean)
Time per request:       18.594 [ms] (mean)
Time per request:       1.859 [ms] (mean, across all concurrent requests)
```

jchesterpivotal:
Oh good, didn't clean up my squash message.

PR #558, opened 2018-04-25
Repair 404 performance regression by using a shared TopicInformer

jchesterpivotal:
## Background

The issue reported at projectriff/riff#553 showed that topic checking introduced a performance regression of 2 orders of magnitude. This renders the s1p sample unusable. The commits in this PR change the implementation to use a callback-with-cache mechanism provided by Kubernetes APIs (shared informers).

To validate performance improvement, I compared the `alwaysTrueTopicExistenceChecker` and `RiffTopicExistenceChecker` using `ab2`. Not very scientific, but a useful sanity check. My test case shows the two checking strategies now perform approximately identically, suggesting they are hitting some bottleneck unrelated to topic lookup.

## alwaysTrueTopicExistenceChecker

```
ab -H "Content-Type: text/plain" -n 100000 -c 100 -p in.txt http://192.168.39.8:30788/requests/numbers
...
Concurrency Level:      100
Time taken for tests:   19.363 seconds
Complete requests:      100000
Failed requests:        0
Total transferred:      10400000 bytes
Total body sent:        17700000
HTML transferred:       300000 bytes
Requests per second:    5164.55 [#/sec] (mean)
Time per request:       19.363 [ms] (mean)
Time per request:       0.194 [ms] (mean, across all concurrent requests)
Transfer rate:          524.53 [Kbytes/sec] received
                        892.70 kb/s sent
                        1417.23 kb/s total

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.2      0       5
Processing:     5   19  38.6     17    1268
Waiting:        5   19  38.6     17    1268
Total:          5   19  38.7     18    1270

Percentage of the requests served within a certain time (ms)
  50%     18
  66%     19
  75%     20
  80%     21
  90%     24
  95%     26
  98%     29
  99%     32
 100%   1270 (longest request)
```

## RiffTopicExistenceChecker using a TopicInformer

```
ab -H "Content-Type: text/plain" -n 100000 -c 100 -p in.txt http://192.168.39.8:32408/requests/numbers
...
Concurrency Level:      100
Time taken for tests:   19.131 seconds
Complete requests:      100000
Failed requests:        0
Total transferred:      10400000 bytes
Total body sent:        17700000
HTML transferred:       300000 bytes
Requests per second:    5227.05 [#/sec] (mean)
Time per request:       19.131 [ms] (mean)
Time per request:       0.191 [ms] (mean, across all concurrent requests)
Transfer rate:          530.87 [Kbytes/sec] received
                        903.50 kb/s sent
                        1434.38 kb/s total

Connection Times (ms)
              min  mean[+/-sd] median   max
Connect:        0    0   0.2      0       5
Processing:     5   19  38.1     17    1247
Waiting:        5   19  38.1     17    1247
Total:          5   19  38.2     17    1250

Percentage of the requests served within a certain time (ms)
  50%     17
  66%     19
  75%     20
  80%     21
  90%     24
  95%     26
  98%     30
  99%     32
 100%   1250 (longest request)
```

### DIFF
--- a/http-gateway/cmd/http-gateway.go
+++ b/http-gateway/cmd/http-gateway.go
@@ -54,29 +54,34 @@ func main() {
 	defer consumer.Close()
 
 	var topicExistenceChecker server.TopicExistenceChecker
+	var informerDone chan struct{}
 	restConf, err := rest.InClusterConfig()
 	if err == nil {
 		riffClient, err := versioned.NewForConfig(restConf)
 		if err != nil {
 			panic(err)
 		}
-		topicExistenceChecker = server.NewRiffTopicExistenceChecker(riffClient)
+		informerDone = make(chan struct{})
+		topicExistenceChecker = server.NewRiffTopicExistenceChecker(riffClient, informerDone)
 	} else {
 		topicExistenceChecker = server.NewAlwaysTrueTopicExistenceChecker()
 	}
 
 	gw := server.New(8080, producer, consumer, 60*time.Second, topicExistenceChecker)
 
-	done := make(chan struct{})
-	gw.Run(done)
+	serverDone := make(chan struct{})
+
+	gw.Run(serverDone)
 
 	// Wait for shutdown
 	signals := make(chan os.Signal, 1)
 	signal.Notify(signals, os.Interrupt, syscall.SIGTERM, os.Kill)
 	<-signals
 	log.Println("Shutting Down...")
-	close(done)
-
+	close(serverDone)
+	if informerDone != nil {
+		close(informerDone)
+	}
 }
 
 func brokers() []string {
--- a/http-gateway/cmd/http-gateway.go
+++ b/http-gateway/cmd/http-gateway.go
@@ -54,6 +54,7 @@ func main() {
 	defer consumer.Close()
 
 	done := make(chan struct{})
+	defer close(done)
 
 	var topicExistenceChecker server.TopicExistenceChecker
 	restConf, err := rest.InClusterConfig()
@@ -76,7 +77,6 @@ func main() {
 	signal.Notify(signals, os.Interrupt, syscall.SIGTERM, os.Kill)
 	<-signals
 	log.Println("Shutting Down...")
-	close(done)
 }
 
 func brokers() []string {
--- a/http-gateway/cmd/http-gateway.go
+++ b/http-gateway/cmd/http-gateway.go
@@ -53,35 +53,30 @@ func main() {
 	}
 	defer consumer.Close()
 
+	done := make(chan struct{})
+
 	var topicExistenceChecker server.TopicExistenceChecker
-	var informerDone chan struct{}
 	restConf, err := rest.InClusterConfig()
 	if err == nil {
 		riffClient, err := versioned.NewForConfig(restConf)
 		if err != nil {
 			panic(err)
 		}
-		informerDone = make(chan struct{})
-		topicExistenceChecker = server.NewRiffTopicExistenceChecker(riffClient, informerDone)
+		topicExistenceChecker = server.NewRiffTopicExistenceChecker(riffClient, done)
 	} else {
 		topicExistenceChecker = server.NewAlwaysTrueTopicExistenceChecker()
 	}
 
 	gw := server.New(8080, producer, consumer, 60*time.Second, topicExistenceChecker)
 
-	serverDone := make(chan struct{})
-
-	gw.Run(serverDone)
+	gw.Run(done)
 
 	// Wait for shutdown
 	signals := make(chan os.Signal, 1)
 	signal.Notify(signals, os.Interrupt, syscall.SIGTERM, os.Kill)
 	<-signals
 	log.Println("Shutting Down...")
-	close(serverDone)
-	if informerDone != nil {
-		close(informerDone)
-	}
+	close(done)
 }
 
 func brokers() []string {
--- a/http-gateway/pkg/server/topic_existence_checker.go
+++ b/http-gateway/pkg/server/topic_existence_checker.go
@@ -12,6 +12,7 @@ import (
 
 	informers "github.com/projectriff/riff/kubernetes-crds/pkg/client/informers/externalversions"
 	"github.com/projectriff/riff/kubernetes-crds/pkg/client/informers/externalversions/projectriff/v1alpha1"
+	"sync"
 )
 
 const (
@@ -27,6 +28,7 @@ type TopicExistenceChecker interface {
 type riffTopicExistenceChecker struct {
 	topicInformer v1alpha1.TopicInformer
 	knownTopics   map[string]ignoredValue
+	mutex *sync.Mutex
 }
 
 type ignoredValue struct {
@@ -45,11 +47,14 @@ func NewAlwaysTrueTopicExistenceChecker() TopicExistenceChecker {
 func NewRiffTopicExistenceChecker(clientSet *versioned.Clientset) TopicExistenceChecker {
 	riffInformerFactory := informers.NewSharedInformerFactory(clientSet, time.Second*30)
 	topicInformer := riffInformerFactory.Projectriff().V1alpha1().Topics()
-
 	knownTopics := make(map[string]ignoredValue)
+	mutex := &sync.Mutex{}
 
 	topicInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
 		AddFunc: func(obj interface{}) {
+			mutex.Lock()
+			defer mutex.Unlock()
+
 			key, err := cache.MetaNamespaceKeyFunc(obj)
 			if err != nil {
 				log.Printf("AddFunc had an error for key '%s': %+v", err)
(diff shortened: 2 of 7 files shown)


jchesterpivotal:
For this PR, I'd argue yes.

http-gateway/pkg/server/topic_existence_checker.go
```diff
@@ -33,29 +43,48 @@ func NewAlwaysTrueTopicExistenceChecker() TopicExistenceChecker {
...
-		if k8serrors.IsNotFound(err) {
-			return false, nil
-		}
+// TopicExists checks to see if Kubernetes is aware of a riff Topic in a namespace.
+func (tec *riffTopicExistenceChecker) TopicExists(namespace string, topicName string) bool {
+	topicKey := fmt.Sprintf("%s/%s", namespace, topicName)
 
-		return false, err
-	}
+	_, exists := tec.knownTopics[topicKey]
```
glyn: Not thread safe. Use a mutex.

jchesterpivotal: Updated. I also kicked around using `RWLock`, but it was measurably slower (approx 20%). Also looked at `sync.Map`, but heeded the documentation's guidance not to use it unless strictly necessary.

http-gateway/pkg/server/topic_existence_checker.go
```diff
@@ -33,29 +45,57 @@ func NewAlwaysTrueTopicExistenceChecker() TopicExistenceChecker {
...
+			defer mutex.Unlock()
 
-func (tec *alwaysTrueTopicExistenceChecker) TopicExists(namespace string, topicName string) (bool, error) {
-	return true, nil
+			key, err := cache.MetaNamespaceKeyFunc(obj)
+			if err != nil {
+				log.Printf("AddFunc had an error for key '%s': %+v", err)
+			}
+
+			knownTopics[key] = ignoredValue{}
```
glyn: According to its docs, `MetaNamespaceKeyFunc` returns a string of the form `namespace/name` or of the form `name` if the namespace is empty (whatever that means - default? unknown?). This opens up the possibility of a bug in this implementation. If the namespace can be empty in situations when it is not the default, then the same topic name in two distinct namespaces could end up being represented by the same key and the map would lose track of the existence of both topics.

http-gateway/pkg/server/topic_existence_checker.go
```diff
@@ -33,29 +45,57 @@ func NewAlwaysTrueTopicExistenceChecker() TopicExistenceChecker {
...
+			log.Printf("Added topic to internal map: %+v", key)
+		},
+
+		DeleteFunc: func(obj interface{}) {
+			mutex.Lock()
+			defer mutex.Unlock()
+
+			key, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+			if err != nil {
+				log.Printf("DeleteFunc had an error for key '%s': %+v", err)
```
glyn: See comments about the corresponding line in `AddFunc`.

http-gateway/pkg/server/topic_existence_checker.go
```diff
@@ -33,29 +45,57 @@ func NewAlwaysTrueTopicExistenceChecker() TopicExistenceChecker {
...
+			}
+
+			delete(knownTopics, key)
+			log.Printf("Removed topic from internal map: %+v", key)
```
glyn: See comments about the corresponding line in `AddFunc`.

http-gateway/pkg/server/topic_existence_checker.go
```diff
@@ -33,29 +45,57 @@ func NewAlwaysTrueTopicExistenceChecker() TopicExistenceChecker {
...
-// error that was encountered.
-func (tec *riffTopicExistenceChecker) TopicExists(namespace string, topicName string) (bool, error) {
+func (tec *alwaysTrueTopicExistenceChecker) TopicExists(namespace string, topicName string) bool {
+	return true
+}
 
-	_, err := tec.client.ProjectriffV1alpha1().Topics(namespace).Get(topicName, v1.GetOptions{})
+// TopicExists checks to see if Kubernetes is aware of a riff Topic in a namespace.
+func (tec *riffTopicExistenceChecker) TopicExists(namespace string, topicName string) bool {
+	tec.mutex.Lock()
```
glyn: Well done (if you are new to mutexes)! A common mistake is to lock for updates and not bother locking for reads, which ruins thread safety.

http-gateway/pkg/server/topic_existence_checker.go
glyn: Did you try `RWMutex`? I don't think the benefits, if any, will be significant, but this would reduce contention between goroutines handling concurrent HTTP requests

http-gateway/pkg/server/topic_existence_checker.go
```diff
@@ -33,29 +45,57 @@ func NewAlwaysTrueTopicExistenceChecker() TopicExistenceChecker {
...
+	defer tec.mutex.Unlock()
 
-	if err != nil {
-		if k8serrors.IsNotFound(err) {
-			return false, nil
-		}
+	topicKey := fmt.Sprintf("%s/%s", namespace, topicName)
```
glyn: This won't work if the key that came back from k8s was of the form `name` because the namespace is empty (whatever that means).

http-gateway/pkg/server/topic_existence_checker.go
jchesterpivotal: My mistake -- I wrote `RWLock` but meant `RWMutex`. It showed up as _slower_, which was a counterintuitive outcome.

Doing an eyeball comparison of `Mutex` and `RWMutex` code shows what appears to be an important difference. `Mutex.Lock()` has [a fast-path that uses a compare-and-swap operation](https://golang.org/src/sync/mutex.go#L73) to acquire the lock, whereas `RWMutex.RLock()` does not have such a fast path ([it atomically increments an int](https://golang.org/src/sync/rwmutex.go#L48)). My hunch is that the `Mutex.Lock()` version is compiling down to take advantage of CPU compare-and-swap instructions, which would be very fast in a lightly-contended lock. Why the `RWMutex` isn't being done with a CAS is a bit beyond my level of understanding.

And having thought further, the lock here _should_ be lightly-contended. The write operations should happen as fast as topics are added or removed, whereas the read operations will be per-message. So it seems as though the fast-path implementation of plain `Mutex` dominates the lock-contention advantage of `RWMutex`.

http-gateway/pkg/server/topic_existence_checker.go
jchesterpivotal: I'm just going to say that _Java Concurrency in Practice_ is worth its weight in gold, even when outside of Java.

http-gateway/pkg/server/topic_existence_checker.go
jchesterpivotal: For the moment I'm inclined to kick the missing-namespace case down the road, as my reckoning is that topics will always have namespaces.

Looking at the `cache` package where `MetaNamespaceKeyFunc` lives shows that it's the default implementation for `KeyFunc` function type. That might actually be the ideal extension point for us to implement checking across all namespaces instead of just the default namespace. It would also let us sidestep the missing-namespace issue you've raised here.

The `DeletionHandlingMetaNamespaceKeyFunc` (which implements `KeyFunc`) has a hardcoded dependency on `MetaNamespaceKeyFunc`, so we would need to replace that also.

I'll add a TODO.

http-gateway/pkg/server/topic_existence_checker.go
```diff
@@ -26,9 +26,9 @@ type TopicExistenceChecker interface {
 }
 
 type riffTopicExistenceChecker struct {
+	mutex         *sync.Mutex
```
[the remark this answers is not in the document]
jchesterpivotal: Ah, you're right. I'd have preferred to name it something like `topicInformerMutex`, but that does not seem to be an idiomatic naming decision for Go source.

http-gateway/pkg/server/topic_existence_checker.go
```diff
@@ -33,29 +43,67 @@ func NewAlwaysTrueTopicExistenceChecker() TopicExistenceChecker {
...
+				log.Printf("DeleteFunc failed during a topic lookup: %#v", err)
+				return
+			}
+
+			delete(knownTopics, key)
+			log.Printf("A topic was removed: %s", key)
+		},
+	})
+
+	done := make(chan struct{}) //TODO: ideally, this would be the same channel used by the gateway itself
```
glyn: We don't seem to close `done` anywhere.

http-gateway/pkg/server/topic_existence_checker.go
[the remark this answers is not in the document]
jchesterpivotal: er, `knownTopicsMutex`.

http-gateway/pkg/server/topic_existence_checker.go
jchesterpivotal: Time to act on my TODO. I had forgotten to come back and work out if there was a way to do it cleanly without widening the `NewRiffTopicExistenceChecker` signature.

http-gateway/cmd/http-gateway.go
```diff
@@ -54,29 +54,34 @@ func main() {
 	defer consumer.Close()
 
 	var topicExistenceChecker server.TopicExistenceChecker
+	var informerDone chan struct{}
```
glyn: Why not create a single `done` channel and immediately defer a close operation on it. Then use that channel for both the server and informer?

Issue #536, opened 2018-04-12
riff CLI should check its version and component versions


trisberg:
The versions should be compatible, CLI version should be >= controller versions

jldec:
I also suggest that if we can't get the version stamped into the build, we should discourage the use of `go get` to install the CLI. (see [main readme](https://github.com/projectriff/riff#get-the-main-riff-repo-and-install-the-cli) and [docs](https://projectriff.io/docs/getting-started-on-minikube/#install-the-current-riff-cli-tool)). 

Would it be possible to add a Makefile target for the cli which also copies the build to the go/bin ?

The [cli folder readme](https://github.com/projectriff/riff/tree/master/riff-cli) needs cleanup also - the instructions there build the old cli in projectriff-archive.

ericbottard:
> Would it be possible to add a Makefile target for the cli which also copies the build to go/bin ?

Yes, that's what `go install <package>` does (instead of just `go build`)

jldec:
How do we set the VERSION of the build like the [Makefile](https://github.com/projectriff/riff/blob/master/riff-cli/Makefile) does?

scothis:
This issue has gone stale, we're on Knative now! Please reopen if still relevant.

PR #562, opened 2018-04-27
Include riff component version in `riff version`

scothis:
```bash
$ riff version
riff CLI version: 0.0.7-snapshot

COMPONENT             IMAGE
function-controller   projectriff/function-controller:0.0.7-snapshot
http-gateway          projectriff/http-gateway:0.0.7-snapshot
topic-controller      projectriff/topic-controller:0.0.7-snapshot
```

Refs #536

### DIFF
--- a/riff-cli/cmd/invokers.go
+++ b/riff-cli/cmd/invokers.go
@@ -18,16 +18,10 @@ package cmd
 
 import (
 	"fmt"
-	"io/ioutil"
-	"net/http"
-	"net/url"
-	"path/filepath"
 
-	"github.com/ghodss/yaml"
-	projectriff_v1 "github.com/projectriff/riff/kubernetes-crds/pkg/apis/projectriff.io/v1alpha1"
 	"github.com/projectriff/riff/riff-cli/cmd/utils"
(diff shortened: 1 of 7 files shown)


trisberg:
Very nice. Should we also add the invoker versions currently registered? That could be another PR.

scothis:
@trisberg `riff invokers list` already does that I'm not opposed to duplicating the functionality, but it's redundant.

jldec:
actually i had the same questions as @trisberg - why not include the invokers list
i think there is value in having a single command for users to run to see what's running in their setup.

in addition to the invoker list, i would suggest adding the output of `kc config current-context`
i think it will help for debugging

scothis:
@jldec @trisberg 

now:
```bash
$ riff version
riff CLI version: 0.0.7-snapshot

COMPONENT             IMAGE
function-controller   projectriff/function-controller:0.0.7-snapshot
http-gateway          projectriff/http-gateway:0.0.7-snapshot
topic-controller      projectriff/topic-controller:0.0.7-snapshot

INVOKER   VERSION
java      0.0.6
node      0.0.9-snapshot
```

Issue #485, opened 2018-03-29
Are topics cluster scoped?


scothis:
The Topic CRD is namespaces scoped, but topics in Kafka and function inputs/outputs are essentially cluster scoped. The k8s resource should either become cluster scoped, or the k8s namespace should be reflected in the topic name in the broker and function bindings.

markfisher:
I think Topics should be cluster-scoped, at least for now. They should only remain namespaced if we intend to enforce scoped bindings.

trisberg:
Then we would need to separate the management of the topics resources from functions more, was mentioned here in #459 as well

ericbottard:
In effect they are, but I think this is yet another topic (ah ah) for which we'll need to sit at the virtual table and re-discuss:
- if topics are clusterscoped, given that several functions can be bound to a topic, and assuming namespacing eventually implies access rules, how can I inject an event for my function F_eric and make sure I won't mess with F_mark (which I don't even konw exists) that is bound on the same topic?
- if topics are namespaced, then we must at the minimum (if we want to keep a single Kafka and a single GW)
  1. change the GW paths to include something that maps to the namespace eg `http://server/{namespace}/{topic}`
  2. internally post to kafka topics named eg `{namespace}/{topic}`

scothis:
This issue has gone stale, we're on Knative now! Please reopen if still relevant.

Issue #459, opened 2018-03-26
manage topics separately from functions  


jldec:
riff create/init makes it too easy to create duplicate topics and assumes 1-to-1 relationship between function and topics.

topics should either be explicitly created, or auto-created by function controller only when the topic(s) do not exist yet.

scothis:
This issue has gone stale, we're on Knative now! Please reopen if still relevant.

PR #566, opened 2018-04-30
Check for topic in all namespaces

### LINKED ISSUE #529: "riff publish" should check all namespaces
Looks like we currently only check for topics in "default" namespace, we should check all namespaces

https://github.com/projectriff/riff/commit/aabc03d57c09ba49778bcd392b67bc97234e6705#r28488084

trisberg:
- this change ignores the namespace and just checks for topic name

- riff currently only works for deploying a function or topic with a specific name into a single namespace, deploying the same resource with the same name to two different namespaces is not supported.

- the same function and topic can be created in two namespaces but which one will be used is unknown

- if a topic that exists in two namespaces is removed from one of the namespaces used the gateway will not be able to find the one from the other namespace

- this is more similar to the behavior that was in place before we started checking for the existence of the topic

See #485 and #459 

Fixes #529

### DIFF
--- a/http-gateway/pkg/server/topic_existence_checker.go
+++ b/http-gateway/pkg/server/topic_existence_checker.go
@@ -12,24 +12,21 @@ import (
 
 	informers "github.com/projectriff/riff/kubernetes-crds/pkg/client/informers/externalversions"
 	"github.com/projectriff/riff/kubernetes-crds/pkg/client/informers/externalversions/projectriff/v1alpha1"
+	"strings"
 	"sync"
-)
-
-const (
-	defaultNamespace = "default" // expected to be used by consumers of TopicExists
 )
 
 // TopicExistenceChecker allows the http-gateway to check for the existence of
 // a Topic before attempting to send a message to that topic.
 type TopicExistenceChecker interface {
-	TopicExists(namespace string, topicName string) bool
+	TopicExists(topicName string) bool
 }
 
 type riffTopicExistenceChecker struct {
 	topicInformer v1alpha1.TopicInformer
 
-	mutex         *sync.Mutex
-	knownTopics   map[string]ignoredValue
+	mutex       *sync.Mutex
+	knownTopics map[string]ignoredValue
 }
 
 type ignoredValue struct{}
@@ -64,8 +61,8 @@ func NewRiffTopicExistenceChecker(clientSet *versioned.Clientset, stop <-chan st
 				return
 			}
 
-			knownTopics[key] = ignoredValue{}
-			log.Printf("New topic has been added: %s", key)
+			knownTopics[topicName(key)] = ignoredValue{}
+			log.Printf("Topic has been added: %s", key)
 		},
 
 		DeleteFunc: func(obj interface{}) {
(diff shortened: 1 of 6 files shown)


glyn:
I'd be more comfortable with this PR if we also introduce a check so that deploying a topic with the same name to two different namespaces would result in the second deployment failing. Then topic name would unambiguously identify a topic.

http-gateway/pkg/server/topic_existence_checker.go
```diff
@@ -91,16 +88,25 @@ func NewRiffTopicExistenceChecker(clientSet *versioned.Clientset, stop <-chan st
 	return &riffTopicExistenceChecker{topicInformer: topicInformer, mutex: mutex, knownTopics: knownTopics}
 }
 
-func (tec *alwaysTrueTopicExistenceChecker) TopicExists(namespace string, topicName string) bool {
+func topicName(key string) string {
```
jchesterpivotal: Turns out there's a provided method for this: `cache.SplitMetaNamespaceKey`. Same package as the `MetaNamespaceKeyFunc` used in the `AddFunc` informer callback.

trisberg:
ATM we don't have a way to enforce that the topics are essentially cluster scoped, scope of this PR is to restore functionality we had before we started the 404 checks

markfisher:
Why not just designate in the CRD itself that Topics are cluster scoped?

trisberg:
so we'll fix #485 at the same time then

trisberg:
wan't sure that we had decided whether to support namespaced topics or not

trisberg:
We might still have the issue of two functions in different namespaces using the same topic. What happens when we delete one of those functions?

glyn:
@trisberg I love the direction of your latest commit. Would you care to summarise the intended behaviour of riff with respect to namespaced topics in this PR? that would be incredibly valuable (especially when it comes to reviewing this PR!). Probably best in a README, e.g. the topic controller README.

trisberg:
The behavior of the topic-controller is outside of the scope for this PR which just deals with looking up that a topic resource exists. The issues with different properties for a duplicate topic is a known issue and should be part of #485.

ericbottard:
> Why not just designate in the CRD itself that Topics are cluster scoped?

Amen. We must be ok with the fact that a function in namespace n1 may write to topic T, while a function in namespace n2 may be downstream T then (while chances are that the "owners" of n1 and n2 have no idea)...

trisberg:
Then we would need to change to not delete a topic that is used by any other function

trisberg:
My current modifications do not change anything other than the ability to post messages to a topic that isn't defined in the default namespace. 

Changing CRD to be cluster scoped requires additional changes in other components and affect the way the functions/topics relate to each other. It's a bigger change that need more care when we do implement it.

markfisher:
> Then we would need to change to not delete a topic resource that is used by any other function

This is exactly the direction we need to go, as described in https://github.com/projectriff/riff/issues/459

glyn:
Merged as e297544c6d0e6b6deaf86ffc6c2244a57c8f7b27

PR #570, opened 2018-05-02
[DNM] Make topics cluster scoped

scothis:
When upgrading a cluster, the previous Topic CRD will need to be delted
before this version can be applied.

Refs #485

### DIFF
--- a/helm-charts/riff/templates/topic-resource.yaml
+++ b/helm-charts/riff/templates/topic-resource.yaml
@@ -3,7 +3,7 @@ kind: CustomResourceDefinition
 metadata:
   name: topics.projectriff.io
 spec:
-  scope: Namespaced
+  scope: Cluster
   group: projectriff.io
   version: v1alpha1
   names:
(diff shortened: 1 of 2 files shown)


trisberg:
We need to change more than this - need to decouple the topics from the functions as well. I don't see any real point to making just this change, plus it will impact the namespace tracking we just implemented for the cli when looking up topics.

ericbottard:
I agree we need to carefully consider this. Maybe a topic (ah ah) for F2F ?

scothis:
I agree this is something we should consider carefully before applying. Added a `DNM` flag to the subject.

> We need to change more than this - need to decouple the topics from the functions as well.

Decoupling Functions and Topics is a much broader concern, I don't see cluster scoped topics as helping or hindering that cause.

> it will impact the namespace tracking we just implemented for the cli when looking up topics.

I tested this change with the latest http-gateway logic, it works. Granted that new code could become much simpler as we can go back to a plain map.

trisberg:
It gets more complicated if you have two functions in different namespaces sharing an input topic - with cluster scoped topics removing one function with `--all` deletes the topic for the other. If the topics are namespace scoped then the other function still has its input topic intact and we can publish messages to it.

ericbottard:
This is obsolete via the move to Knative. Knative serving has Bus and ClusterBus

PR #573, opened 2018-05-02
Parameterize ports between invoker and sidecar

scothis:
The port used for the function sidecar to communicate with the function
invoker is currently shared knowledge. Each invoker needs to know to
listen on 8080 for http or 10382 for grpc. This knowledge is hard coded
into the sidecar and each function invoker.

Instead the ports should be determined by the function-controller and
set into the function's deployment. For the sidecar there is a new
`--port` argument, and for the invoker there are environment variables
`HTTP_PORT` and `GRPC_PORT`.

There are two port variables for the invoker because there are cases
where an invoker may need to support http and grpc.

These ports should no longer be hard coded in invokers. Although the
values have not changed.

### DIFF
--- a/function-controller/pkg/controller/deployer.go
+++ b/function-controller/pkg/controller/deployer.go
@@ -36,7 +36,11 @@ const (
 )
 
 var (
-	zero = int32(0)
+	zero  = int32(0)
+	ports = map[string]string{
+		"http": "8080",
+		"grpc": "10382",
+	}
 )
 
 // Deployer allows the realisation of a function on k8s and its subsequent scaling to accommodate more/less load.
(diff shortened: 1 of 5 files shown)


ericbottard:
LGTM, but I need to ask: why? What's wrong with hardcoded knowledge of the ports? All parties involved are riff specific, and this can be part of the contract. Now the contract is "there is a GRPC_PORT and a HTTP_PORT env var". We have to draw the line somewhere, so why here?

scothis:
Decoupling the ports makes our function invokers more flexible. The ports are currently specified as part of the invoker contract, but now they are specified as environment vars instead of hard coded values.

Which port used is not very interesting, but that the sidecar and invoker agree on the port is critical. With this change the function-controller is responsible for making sure the sidecar and invoker agree on which port to use.

Issue #572, opened 2018-05-02
Sidecar errors when running Java functions


trisberg:
The new timeout features for the grpc protocol times out quickly and errors before the Java function has time to start up which could take 10s or more depending on the function.

We need a way to tweak the timeout settings. Maybe introduce an `initialDelay` property.

Once the sidecar container errors it gets restarted but that could cause the back-off-loop to delay the availability fo the functions. Also ugly to see error status and restart count when running the function:

```
$ kubectl get pod -w
NAME                         READY     STATUS    RESTARTS   AGE
uppercase-8545f6fb95-wdqwz   0/2       Pending   0          0s
uppercase-8545f6fb95-wdqwz   0/2       Pending   0         0s
uppercase-8545f6fb95-wdqwz   0/2       ContainerCreating   0         0s
uppercase-8545f6fb95-wdqwz   2/2       Running   0         2s
uppercase-8545f6fb95-wdqwz   1/2       Error     0         10s
uppercase-8545f6fb95-wdqwz   2/2       Running   1         11s
uppercase-8545f6fb95-wdqwz   1/2       Error     1         20s
uppercase-8545f6fb95-wdqwz   1/2       CrashLoopBackOff   1         32s
uppercase-8545f6fb95-wdqwz   2/2       Running   2         33s
uppercase-8545f6fb95-wdqwz   2/2       Terminating   2         47s
uppercase-8545f6fb95-wdqwz   0/2       Terminating   2         50s
```

dturanski:
@trisberg I'm reconsidering the initial delay in favor of making the grpc wait time configurable.  Currently the grpc wait time is set to 100ms with an exponential backoff of 1sec - 4sec between retries. This causes the sidecar to crash if a java app takes > ~7.3 sec to start which is not that unlikely (the Java invoker `uppercase` sample takes about 5 sec.   Talking with @dsyer, he introduced the backoff loop and 100ms wait time. Originally it was set to 60sec. This would cause an unreasonable delay if the function container crashed and recovered, since a new connection is required after the function container recovers, a much shorter wait time eliminated this delay.  Dave's original change also used a fixed backoff which would retry forever, so no significant downside for the java happy path.

[...]

PR #576, opened 2018-05-03
DO NOT MERGE: Add InitalDelayMs to function spec and use in sidecar

dturanski:
Addresses #572

### DIFF
--- a/function-controller/pkg/controller/deployer.go
+++ b/function-controller/pkg/controller/deployer.go
@@ -18,6 +18,7 @@ package controller
 
 import (
 	"log"
+	"strconv"
 
 	"os"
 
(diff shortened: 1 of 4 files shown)


function-controller/pkg/controller/deployer.go
```diff
@@ -121,12 +122,18 @@ func (d *deployer) buildSidecarContainer(function *v1.Function) corev1.Container
 	if outputDestination == "" {
 		outputDestination = "replies"
 	}
+	initialDelayVal := "0"
+	if function.Spec.InitialDelayMs != nil {
```
ericbottard: Sorry, I was wrong in the call. If the Function went through settings defaults, `0` should already be there.
Better not to encode the logic of that default in several places...

function-controller/pkg/controller/deployer.go
```diff
@@ -121,12 +122,18 @@ func (d *deployer) buildSidecarContainer(function *v1.Function) corev1.Container
...
+		initialDelayVal = strconv.Itoa(int(*function.Spec.InitialDelayMs))
+	}
+	//Check for nil and convert int32 to int and then to str.
```
ericbottard: Can remove

function-sidecar/cmd/function-sidecar.go
```diff
@@ -94,7 +95,7 @@ func main() {
 		output = ""
 	}
 
-	log.Printf("Sidecar for function '%v' (%v->%v) using %v dispatcher starting\n", group, input, output, protocol)
+	log.Printf("Sidecar for function '%v' (%v->%v) using %v on port %v dispatcher starting\n", group, input, output, protocol, port)
```
ericbottard: The sentence now doesn't make sense (word "dispatcher" needs to be next to the protocol)

dturanski:
Closing in favor of #577

PR #579, opened 2018-05-04
Allow artifact to be absolute or relative to $CWD

### LINKED ISSUE #417: artifact should not be considered as relative to functionpath
It's ok for it to be considered relative once we're inside the cli code, but from a flags perspective, both artifact and function path should be just that: paths (hence absolute or relative to CWD).
One of the advantages is that bash completion could work on both.

dturanski:
Fixes #417

### DIFF
--- a/riff-cli/cmd/common_options.go
+++ b/riff-cli/cmd/common_options.go
@@ -36,7 +36,6 @@ func validateFunctionName(name *string, path string) error {
 
 func validateAndCleanArtifact(artifact *string, path string) error {
 	if *artifact != "" {
-		fmt.Printf("artifact: %v\n", *artifact)
 		absFilePath, err := filepath.Abs(path)
 		if err != nil {
 			return err
--- a/riff-cli/cmd/common_options.go
+++ b/riff-cli/cmd/common_options.go
@@ -43,8 +43,8 @@ func validateAndCleanArtifact(artifact *string, path string) error {
 		}
 
 		/*
-		 *	If artifact is relative to current directory or an absolute path then use the absolute path
-		 *  else make it relative to the file path directory. But it must be in the file path directory.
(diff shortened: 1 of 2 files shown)


riff-cli/cmd/common_options.go
```diff
@@ -35,23 +36,28 @@ func validateFunctionName(name *string, path string) error {
...
-		}
-
+		fmt.Printf("artifact: %v\n", *artifact)
 		absFilePath, err := filepath.Abs(path)
 		if err != nil {
 			return err
 		}
 
+		/*
+		 *	If artifact is relative to current directory or an absolute path then use the absolute path
```
dturanski: Not sure why this formatting is happening

ericbottard:
I think the code could be simplified even more (the logic doesn't require to always think in terms of "absolute" paths), but the real problem is downstream: what happens if the path is outside the directory of the function? There is an assumption (because of Docker context) that it is

dturanski:
I don't understand your comment. It will not allow the artifact to be outside the directory, but it may be given as a relative path or absolute path. https://github.com/dturanski/riff/blob/issue_417/riff-cli/cmd/common_options.go#L68

dturanski:
I'm all for simplifying it. Currently it checks if the artifact exists. If it's a relative path that does not exist, the assumption is it may be relative to the function directory.  It seems like a simple way to handle cases like `../myfunc/foo.js` (may or may not exist but must exist relative to the function dir)   vs `foo.js` (exists only relative to the function dir)  or `C:\work\myfunc\foo.js`  or `/work/myfunc/foo.js`  (must exist and be in the function dir)

trisberg:
Note: when testing with node you would need a fix to the invoker: https://github.com/projectriff/node-function-invoker/pull/59

PR #583, opened 2018-05-09
Add windowing capability at sidecar

ericbottard:
Works with https://github.com/projectriff/go-function-invoker/pull/10


function-sidecar/pkg/dispatcher/grpc/grpc.go
```diff
@@ -52,44 +59,100 @@ func (this *grpcDispatcher) Closed() <-chan struct{} {
...
+// A WindowingStrategyFactory is a function for creating a new instance of a (most certainly stateful) WindowingStrategy,
+// given a new, first message for a stream.
+type WindowingStrategyFactory func(message.Message) WindowingStrategy
+
+//
+type WindowingStrategy interface {
+	// ShouldClose notifies this strategy that a new message has just been sent to the function, resulting in the optional
+	// error passed as 2nd argument.
+	// ShouldClose should return true as its first result if this strategy decides that the current stream should end.
+	ShouldClose(in message.Message, err error) bool
```
glyn: As defined, windows cannot consist of a single message since `ShouldClose` is called _after_ the message has been processed by the function. Why not remove the `Message` parameter from the `WindowingStrategyFactory` and then `ShouldClose` can be called with the first message **and** optional error? This allows the strategy to deal with an error from the first message and also allows the strategy to produce a window of size 1.

function-sidecar/pkg/dispatcher/grpc/windowing_size.go
```diff
@@ -0,0 +1,36 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package grpc
+
+import "github.com/projectriff/riff/message-transport/pkg/message"
+
+// Type size is a WindowingStrategy based on the number of input messages seen: the stream is closed after N messages.
+type size struct {
```
glyn: Same nit as on the `ptimeWindow` `struct`: I don't think you need a `struct` here.

function-sidecar/pkg/dispatcher/grpc/windowing_processing_time_window.go
```diff
@@ -0,0 +1,37 @@
...
+
+package grpc
+
+import (
+	"github.com/projectriff/riff/message-transport/pkg/message"
+	"time"
+)
+
+// Type ptimeWindow is a WindowingStrategy that is based on elapsed wall clock time.
+type ptimeWindow struct {
```
glyn: Just a nit, but I'm not sure `ptimeWindow` needs to be a `struct`. I _think_ you could define:
```
type ptimeWindow int
```
and still define a method:
```
func (s *ptimeWindow) ShouldClose(...
```

function-sidecar/pkg/dispatcher/grpc/grpc.go
```diff
@@ -114,14 +177,23 @@ func NewGrpcDispatcher(port int, timeout time.Duration) (dispatcher.Dispatcher,
...
+		output:         make(chan message.Message, 100),
+		closed:         make(chan struct{}),
+		streams:        make(map[interface{}]ongoingStream),
+		windowingFactory: unmarshallFactory(),
+		correlation: func(m message.Message) interface{} {
+			// Correlate by correlationId header by default if present
+			if c, ok := m.Headers()[correlationId] ; ok {
+				return c[0]
+			} else {
+				return nil
```
glyn: I wonder if we should correlate by partition too to preserve kafka ordering guarantees.

function-sidecar/pkg/dispatcher/grpc/grpc.go
ericbottard: I don't understand your comment. What we cannot do is windows of size _zero_, not one. Or am I missing something?

While not 100% necessary, passing the message to the factory is for strategies that need info from the first message (think time windows with _event time_ semantics)

function-sidecar/pkg/dispatcher/grpc/windowing_processing_time_window.go
ericbottard: I think I tried (elsewhere) and failed. So did that for consistency.

Will revisit

function-sidecar/pkg/dispatcher/grpc/grpc.go
ericbottard: Not sure what you mean here. Broker partitioning has already happened upstream, so this partitioning is finer grained.
Imagine we have 2 kafka partitions and one pod gets even orderIds, while the other gets odd ones.

Then _this_ partitioning can be by, say, orderId itself. But this particular pod may only see odd orderIds anyway

function-sidecar/pkg/dispatcher/grpc/grpc.go
```diff
@@ -52,44 +59,100 @@ func (this *grpcDispatcher) Closed() <-chan struct{} {
...
-					if streamClosureDiagnosed(err) {
-						close(this.closed)
-						return
-					}
-
-					log.Printf("Error sending message to function: %v", err)
-				}
+// A CorrelationStrategy groups together messages that belong to the same "spatial" stream (the "temporal" aspect
+// of grouping is taken care of by a WindowStrategy). Several invocations for different spatial groups may happen concurrently.
+type CorrelationStrategy func(message.Message) interface{}
```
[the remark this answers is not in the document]
ericbottard: Sure. `nil == nil` so while it could clarify things, there is no real need: `nil` is not a special case. It's just one way to do no special partitioning

function-sidecar/pkg/dispatcher/grpc/grpc.go
glyn: Oh, is the protocol that the first message is passed both to the factory and also to `ShouldClose`?  It would be good to document that if it's the case. If so, that would enable a window size of 1.

function-sidecar/pkg/dispatcher/grpc/grpc.go
glyn: I don't understand how broker partitioning can already have happened. The sidecar consumer gets a message and sends it to the gRPC dispatcher. But if multiple partitions are associated with the consumer, the message could be from any partition.

function-sidecar/pkg/dispatcher/grpc/grpc.go
ericbottard: Yes.

I thought you guys heard me say "window of size 1" enough times :) How could you even think it did not support that. I have recurring dreams of windows of size 1!

function-sidecar/pkg/dispatcher/grpc/grpc.go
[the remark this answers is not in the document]
glyn: Well, `nil` might indicate a default partition, especially since [`nil` can be a map key](https://play.golang.org/p/eYu-YYKRprb).

function-sidecar/pkg/dispatcher/grpc/grpc.go
markfisher: yes, I agree we would ideally de-multiplex before invoking the RPC calls so that for any single RPC call, the messages are from a single partition, but perhaps that's something to address after this PR, since that same situation applies today.... at least we'd be making progress on the windowing as opposed to always sending unbounded streams

glyn: Ok, fair enough. Thanks.

function-sidecar/pkg/dispatcher/grpc/grpc.go
```diff
@@ -52,44 +59,100 @@ func (this *grpcDispatcher) Closed() <-chan struct{} {
...
+}
+
+const correlationId = "correlationId"
+var headersToForward []string
+func init() {
+	headersToForward = []string {correlationId}
+}
+
+
+type ongoingStream struct {
```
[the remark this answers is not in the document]
ericbottard: Not this is actually the window. So "window" could be the struct name

function-sidecar/pkg/dispatcher/grpc/grpc.go
```diff
@@ -114,14 +177,23 @@ func NewGrpcDispatcher(port int, timeout time.Duration) (dispatcher.Dispatcher,
...
+	result := &grpcDispatcher{
+		client:         function.NewMessageFunctionClient(conn),
+		input:          make(chan message.Message, 100),
+		output:         make(chan message.Message, 100),
+		closed:         make(chan struct{}),
+		streams:        make(map[interface{}]ongoingStream),
+		windowingFactory: unmarshallFactory(),
+		correlation: func(m message.Message) interface{} {
+			// Correlate by correlationId header by default if present
+			if c, ok := m.Headers()[correlationId] ; ok {
```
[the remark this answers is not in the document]
ericbottard: As explained in the `CorrelationStrategy` type doc, this is what I call the "spatial" correlation (similar to say, the grouping correlation strategy that an SI aggregator does).

I'm using the `correlationId` here by default because the sidecar is now responsible for forwarding the `correlationId` on the way back. If we did not correlate (ie if there was just 1 giant group, even though of size 1) it is possible that response for request B ends up as the response for request A

function-sidecar/pkg/dispatcher/grpc/grpc.go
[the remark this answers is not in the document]
markfisher: I would prefer "window" then.

function-sidecar/pkg/dispatcher/grpc/grpc.go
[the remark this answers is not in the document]
markfisher: I don't think we're ready to get into aggregation use-cases yet.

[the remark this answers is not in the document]
ericbottard: Now that I think of it, this should not even be necessary as if a window is already opened (waiting for its response) and a new request comes in, a new window is created and replaces the current one in the map (assuming no correlation whatsoever). But the evicted window should still "live" and wait for the reply (with some eventual reaping mechanism)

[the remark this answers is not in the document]
ericbottard: Well, my (unconcious at the time) reasoning is that if we move the windowing out of the function, potential correlation has to come with it, because I think in most cases you want correlation to happen _before_ windowing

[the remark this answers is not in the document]
markfisher: You are referring to what would normally go into a `groupBy(..)` on a `Flux` for example? That could happen before or after windowing, depending on the use-case. I think we should be careful not to conflate "correlation" with "grouping", at least not yet (we have a very specific role for correlation today, and it's not about grouping for processing).

function-sidecar/pkg/dispatcher/grpc/grpc.go
```diff
@@ -147,3 +219,27 @@ func toDispatcher(grpc *function.Message) message.Message {
...
 	return message.NewMessage(grpc.Payload, dHeaders)
 }
+
+func unmarshallFactory() WindowingStrategyFactory {
+	var w v1alpha1.Windowing
+	json.Unmarshal([]byte(os.Getenv("WINDOWING_STRATEGY")), &w)
+	// TODO: this can be done reflectively, once there are more alternatives.
+	// Also, check there is at most one key with a non-zero value
+	if w.None {
+		log.Printf("Will use 'infinite' windowing strategy\n")
```
markfisher: I would log that it will *not* use a windowing strategy (i.e. unbounded stream). I'd avoid the term 'infinite' in favor of consistency around bounded vs. unbounded terminology.

function-sidecar/pkg/dispatcher/grpc/windowing_none.go
```diff
@@ -0,0 +1,34 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package grpc
+
+import "github.com/projectriff/riff/message-transport/pkg/message"
+
+// Type none is a WindowingStrategy that does no windowing at all: the stream is everlasting.
```
markfisher: again: s/everlasting/unbounded/

kubernetes-crds/pkg/client/clientset/versioned/fake/register.go
```diff
@@ -29,7 +29,7 @@ var codecs = serializer.NewCodecFactory(scheme)
 var parameterCodec = runtime.NewParameterCodec(scheme)
 
 func init() {
-	v1.AddToGroupVersion(scheme, schema.GroupVersion{Version: "v1alpha1"})
+	v1.AddToGroupVersion(scheme, schema.GroupVersion{Version: "v1"})
```
markfisher: why has this (and below) changed to "v1"?

kubernetes-crds/pkg/apis/projectriff.io/v1alpha1/windowing.go
```diff
@@ -0,0 +1,28 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package v1alpha1
+
+type Windowing struct {
```
markfisher: these 3 are all mutually exclusive right? how is that handled. Would it be better to have a "Type" enum and a value?

trisberg:
I tested the three Java samples, all LGTM with the latest changes

function-sidecar/pkg/dispatcher/grpc/grpc.go
```diff
@@ -52,59 +58,157 @@ func (this *grpcDispatcher) Closed() <-chan struct{} {
...
+// A WindowingStrategyFactory is a function for creating a new instance of a (most certainly stateful) WindowingStrategy,
+// given a new, first message for a stream.
+type WindowingStrategyFactory func(message.Message) WindowingStrategy
+
+//
+type WindowingStrategy interface {
+	// ShouldClose notifies this strategy that a new message has just been sent to the function, resulting in the optional
+	// error passed as 2nd argument. This is a synchronous way of deciding whether or not a window should be closed.
+	// ShouldClose should return true if this strategy decides that the current stream should end.
+	ShouldClose(in message.Message, err error) bool
```
glyn: It's still not clear whether `ShouldClose` needs to be passed the message that was used to create the `WindowingStrategy`. I'd prefer the factory to take no parameters to avoid this ambiguity, if that makes sense. (Sorry if I'm repeating myself, but I couldn't find the previous thread on this point or remember the conclusion.)

function-sidecar/pkg/dispatcher/grpc/grpc.go
```diff
@@ -52,59 +58,157 @@ func (this *grpcDispatcher) Closed() <-chan struct{} {
...
+
+	// AsyncClose should return a non-nil channel that this strategy will close if it decides that the current window
+	// should end asynchronously
```
glyn: I'm being pedantic about the English, but shouldn't the comment say "... if it decides asynchronously that the current window should end."? It's like an async equivalent to `ShouldClose` returning `true` isn't it?

function-sidecar/pkg/dispatcher/grpc/grpc.go
```diff
@@ -52,59 +58,157 @@ func (this *grpcDispatcher) Closed() <-chan struct{} {
...
+	AsyncClose() <-chan struct{}
```
glyn: Wouldn't it be cleaner to return the channel from the factory since the channel and the `WindowingStrategy` are essentially peers of each other? A benefit of this approach would be that you could then build a `select` from a fixed number of cases which will make the loop below easier to read.

function-sidecar/pkg/dispatcher/grpc/grpc.go
```diff
@@ -147,3 +258,43 @@ func toDispatcher(grpc *function.Message) message.Message {
...
+		return ptimeFactoryFactory(d)
+	} else if w.Session != "" {
+		log.Printf("Will use session based windowing strategy of %v\n", w.Session)
+		d, err := time.ParseDuration(w.Session)
+		if err != nil {
+			panic(err) // Assume it's validated at fn registration time
+		}
+		return ptimeSessionFactoryFactory(d)
+	} else {
+		log.Printf("Will use no windowing strategy (unbounded stream)\n")
```
glyn: I'd prefer this to say "Will use unbounded windowing strategy ..." since we are using a strategy even though it's unbounded.

function-sidecar/pkg/dispatcher/grpc/grpc_test.go
```diff
@@ -17,30 +17,199 @@
...
+			m = <-d.Output()
+			received[string(m.Payload())] = struct{}{}
+			m = <-d.Output()
+			received[string(m.Payload())] = struct{}{}
+			m = <-d.Output()
+			received[string(m.Payload())] = struct{}{}
+
+			expected := map[string]struct{}{"100": {}, "150": {}, "170": {}, "54": {}, "98": {}}
+
+			Expect(received).To(Equal(expected))
```
glyn: Why is this doing a set-wise comparison since the ordering seems to be important?

kubernetes-crds/pkg/apis/projectriff.io/v1alpha1/windowing.go
glyn: @ericbottard For the record, I'm wondering the same thing as @markfisher. You are simulating a union here.  There is presumably quite a lot of overlap between the "branches" of the simulated union for the `Volume` spec, but here the data seems to be entirely disjoint.

function-sidecar/pkg/dispatcher/grpc/grpc.go
ericbottard: ShouldClose doesn't take the message that was passed to the strategy (ie the *first* one), but rather the last (ie most recent) one. They're only the same in the case of window-of-size-one.

The factory takes the first message to ease the case of remembering eg when a stream was created

function-sidecar/pkg/dispatcher/grpc/grpc.go
ericbottard: Absolutely, will change

function-sidecar/pkg/dispatcher/grpc/grpc.go
ericbottard: Could do.

The select will not be out of fixed number of cases though, because streams come and go at runtime, so will always need to deal with that.

But optional channel as second return value is indeed more sensible

function-sidecar/pkg/dispatcher/grpc/grpc.go
ericbottard: Will change

function-sidecar/pkg/dispatcher/grpc/grpc_test.go
ericbottard: The ordering *is* indeed important, but not that much. What's more important to notice is that results of 2 streams can be received in parallel, so I chose a setup that can cope with that easily at the detriment of order checking (per stream). Ordering is checked in first tests for what it's worth

function-sidecar/pkg/dispatcher/grpc/grpc.go
glyn: Ok. May I suggest a small improvement to the comment? The phrase "a new message" might be taken to imply "a message different from that which was passed to the factory". I suggest something along these lines (which also clarified the _requirement_ to call this method - for example, it's not acceptable simply to call this method from time to time):

> This method should be called with each message that is sent to the function, including the 
> message that was passed to the `WindowingStrategyFactory` to construct the 
> `WindowingStrategy`.

kubernetes-crds/pkg/client/clientset/versioned/fake/register.go
ericbottard: After second inspection, there is nothing here that should refer to these CRDs (riff's) version. So maybe it was a bug in the generator previously (or we changed too much by hand)

function-sidecar/pkg/dispatcher/grpc/grpc.go
ericbottard: Keeping the channel as the way it is now because:
- it needs to be looked at more than once, not just at construction time
- feels weird to have an XFactory return an X and something else, which really belongs to the logic of X..

glyn:
Merged as 8efed627f7c2b68e325f8ecb885585f5ba38bfd1

PR #591, opened 2018-05-18
[DNM] Prepare for deploying riff with istio enabled

trisberg:
- add opt-out annotations to kafka and zk

### DIFF
--- a/config/kafka/kafka.yaml
+++ b/config/kafka/kafka.yaml
@@ -45,6 +45,8 @@ spec:
   replicas: 1
   template:
     metadata:
+      annotations:
+        sidecar.istio.io/inject: "false"
       labels:
         app: kafka
         component: zookeeper
@@ -72,6 +74,8 @@ spec:
   replicas: 1
   template:
     metadata:
+      annotations:
+        sidecar.istio.io/inject: "false"
       labels:
         app: kafka
         component: kafka-broker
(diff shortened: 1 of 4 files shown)


config/kafka/kafka.yaml
```diff
@@ -72,6 +74,8 @@ spec:
   replicas: 1
   template:
     metadata:
+      annotations:
+        sidecar.istio.io/inject: "false"
```
glyn: What's the rationale for not injecting istio into kafka? How will this play out if istio is injected into the pods which need to use kafka?

trisberg: It was more of an experiment, I thought that the Envoy proxy was interfering somehow, but that wasn't the case. I had to create an egress rule whether Kafka was fronted with a proxy or not. On the other hand the proxy isn't adding anything for Kafka, we're not going to use Istio to control anything Kafka related until Envoy supports the Kafka protocol, so why add the Envoy proxy sidecar?

glyn: Ok, thanks. We can always inject istio into kafka later if need be.

glyn: I don't mind either way. Perhaps others will. If we don't merge, please can we close this PR and re-open it later if needed? You see any open PR creates a cognitive burden for the team.

trisberg:
Holding off on merging this

PR #594, opened 2018-05-22
Modify the docker command environment to work with minikube

### LINKED ISSUE #593: riff build no longer works with eval $(minikube docker-env)
It's no longer possible to build any samples for minikube by using `eval $(minikube docker-env)` without using the `--push` option. The images now end up in the local docker daemon and not the minikube one

trisberg:
- adding the env vars for `minikube docker-env`

Fixes #593

### DIFF
--- a/riff-cli/pkg/docker/docker_utils_test.go
+++ b/riff-cli/pkg/docker/docker_utils_test.go
@@ -57,10 +57,15 @@ var _ = Describe("DockerUtils", func() {
 
 		Describe("Environment variable propagation", func() {
 			var (
-				homeWasSet bool
-				oldHome    string
-				pathWasSet bool
-				oldPath string
+				homeWasSet    bool
+				oldHome       string
+				pathWasSet    bool
+				oldPath       string
+				dockerWasSet  bool
+				oldDockerHost string
+				oldDockerCert string
+				oldDockerTLS  string
+				oldDockerAPI  string
 			)
 
 			BeforeEach(func() {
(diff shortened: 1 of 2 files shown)


riff-cli/pkg/docker/docker_utils_test.go
```diff
@@ -69,6 +74,15 @@ var _ = Describe("DockerUtils", func() {
 
 				oldPath, pathWasSet = os.LookupEnv("PATH")
 				Expect(os.Unsetenv("PATH")).To(Succeed())
+
+				oldDockerHost, dockerWasSet = os.LookupEnv("DOCKER_HOST")
+				oldDockerCert, _ = os.LookupEnv("DOCKER_CERT_PATH")
```
glyn: This should record `dockerCertWasSet` so that the `AfterEach` can correctly reset or unset the value. As this stands, if `DOCKER_CERT_PATH` is set but `DOCKER_HOST` is not set, the test will corrupt the value of `DOCKER_CERT_PATH` (by unsetting it rather than resetting it to its original value).

I guess it's unlikely that `DOCKER_CERT_PATH` would be set without `DOCKER_HOST`, but for the sake a few lines of code, this assumption can be removed and the code made watertight.

ericbottard:
Discussed this with @glyn :

Seems like we've made this more complicated than it needs to be really.

if `Command.Env` is `nil`, then the parent process env is inherited (which makes sense, because it was globally working previously).

I think we should rather revert @glyn 's PR and make sure `Command.Env` is `nil`

glyn:
I agree that the parent process environment should be inherited by default, but that wasn't what I was seeing. I'm happy to revert my change for now.

glyn:
I reverted my change. Closing this PR for now.

PR #596, opened 2018-05-23
Prevent re-dockerization when it is not needed.

ericbottard:


glyn:
How does this change cope with switching between branches? Will it always know to rebuild docker images based on the branch contents?

ericbottard:
The same way `make` rebuilds a different binary when you switch branches: `git` touches files when you switch branches and file contents change

PR #597, opened 2018-05-23
Add advertised DNS name for kafka broker

trisberg:
- this change is required in order to seamlessly work with Istio 0.8.0

- add headless Kafka brokers service to enable DNS name for broker

- set hostname and subdomain for kafka broker pod

- switch to use the DNS name for advertised host

- since we aren't using a StatefulSet, this only works when using 1 broker replica which is all we need

- bump chart version

### DIFF
--- a/config/kafka/kafka.yaml
+++ b/config/kafka/kafka.yaml
@@ -34,6 +34,23 @@ spec:
     component: zookeeper
 
 ---
+apiVersion: v1
+kind: Service
+metadata:
+  name: projectriff-brokers
+  labels:
+    app: kafka
+    component: kafka-broker
+spec:
+  clusterIP: None
+  ports:
+  - port: 9092
+    name: broker # Port is not used
+  selector:
+    app: kafka
+    component: kafka-broker
+
+---
 apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
@@ -69,13 +86,15 @@ metadata:
     component: kafka-broker
   name: projectriff-kafka
 spec:
-  replicas: 1
+  replicas: 1 # The headless service/hostname config only supports 1 replica!
   template:
     metadata:
       labels:
         app: kafka
         component: kafka-broker
     spec:
+      hostname: broker-1
+      subdomain: projectriff-brokers
       containers:
       - name: kafka
         image: wurstmeister/kafka:0.11.0.1
@@ -87,8 +106,6 @@ spec:
           - name: KAFKA_ADVERTISED_PORT
             value: "9092"
           - name: KAFKA_ADVERTISED_HOST_NAME
-            valueFrom:
-              fieldRef:
-                fieldPath: status.podIP
+            value: broker-1.projectriff-brokers.riff-system.svc.cluster.local
           - name: KAFKA_ZOOKEEPER_CONNECT
             value: projectriff-zookeeper:2181
--- a/helm-charts/kafka/Chart.yaml
+++ b/helm-charts/kafka/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 description: kafka for riff
 name: kafka
-version: 0.0.2
+version: 0.0.3
 appVersion: 0.11.0.1
 home: https://kafka.apache.org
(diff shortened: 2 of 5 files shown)


config/kafka/kafka.yaml
```diff
@@ -69,13 +86,15 @@ metadata:
...
 spec:
-  replicas: 1
+  replicas: 1 # The headless service/hostname config only supports 1 replica!
   template:
     metadata:
       labels:
         app: kafka
         component: kafka-broker
     spec:
+      hostname: broker-1
```
ericbottard: I don't understand this. Why force a hostname?

config/kafka/kafka.yaml
```diff
@@ -87,8 +106,6 @@ spec:
           - name: KAFKA_ADVERTISED_PORT
             value: "9092"
           - name: KAFKA_ADVERTISED_HOST_NAME
-            valueFrom:
-              fieldRef:
-                fieldPath: status.podIP
+            value: broker-1.projectriff-brokers.riff-system.svc.cluster.local
```
ericbottard: why not just put the service here? Or even maybe a `valueFrom` extracting the service?

config/kafka/kafka.yaml
```diff
@@ -33,6 +33,23 @@ spec:
...
+kind: Service
+metadata:
+  name: projectriff-brokers
+  labels:
+    app: kafka
+    component: kafka-broker
+spec:
+  clusterIP: None
+  ports:
+  - name: dummy # Actually, no port is needed.
```
ericbottard: So might as well use 9092

config/kafka/kafka.yaml
trisberg: No, that doesn't work. Each broker should advertise it's hostname which should be resolvable by any clients. The service is only used initially to obtain a list of broker hosts either IP addresses or DNS names. At least that's my understanding how the Kafka connections work.

config/kafka/kafka.yaml
trisberg: I guess, will change

config/kafka/kafka.yaml
ericbottard: True, but as stated elsewhere there is only one replica. So service address == pod address

config/kafka/kafka.yaml
ericbottard: See my comment below about having only one replica. This should make this unnecessary.

Also, how would this play out if Envoy was sitting in the middle?

config/kafka/kafka.yaml
trisberg: That's not necessarily true either, the service and the pod could have different IP addresses

```
/ # nslookup broker-1.projectriff-brokers.riff-system
Server:    10.96.0.10
Address 1: 10.96.0.10 kube-dns.kube-system.svc.cluster.local

Name:      broker-1.projectriff-brokers.riff-system
Address 1: 172.17.0.12 broker-1.projectriff-brokers.riff-system.svc.cluster.local
```

```
/ # nslookup projectriff-kafka.riff-system
Server:    10.96.0.10
Address 1: 10.96.0.10 kube-dns.kube-system.svc.cluster.local

Name:      projectriff-kafka.riff-system
Address 1: 10.98.53.134 projectriff-kafka.riff-system.svc.cluster.local
```

ericbottard: But what's important is that kafka is reachable by the service address (from the "outside").

This is basically what we did last week together in Boston, and it works (including with Envoy in the middle)

trisberg: There is no outside, these are cluster IPs, not reachable from the outside except for the NodePort one that could be reached. The kafka clients have to run inside the k8s cluster or we would have to expose the broker node as well. I think what worked for the experiments was a fluke where the pod and service shared an IP address. I have not been able to replicate this. None of this makes a difference for istio, we still need an egress rule for the pod IP range. I will try some route rules to see if that can be avoided.

ericbottard: Sorry, I wasn't clear. When I wrote "outside", I meant outside of this pod.

My belief is that this should work, even with istio present, and even with no particular route/egress. And also if the IPs are different (as long as clients to kafka connect to it via the service)

trisberg:
Rebased and squashed

ericbottard:
This is obsolete with the move to Knative. The sample kafka bus provides a similar setup

PR #598, opened 2018-05-24
Improve README

glyn:


README.adoc
```diff
@@ -75,6 +75,13 @@ Once you're done playing with riff (see samples below), you can destroy everythi
 make teardown
 ----
 
+To unset the environment variables for building images with docker in minikube, issue:
```
ericbottard: This could read like "to build in minikube, some vars need to be unset, here's how".

Should we use "To unset the environment variables *used* for building images ..."

glyn: Yes we should. I'll fix that.

glyn: @ericbottard Fixed.

PR #601, opened 2018-05-29
[DNM] Support stream gateway in function controller

glyn:
Support is provided by enabling a feature flag BEFORE the function controller is
deployed. The feature flag is enabled by setting the property
stream-gateway=enabled in the riff-feature-flags configmap in the riff-system
namespace.

For example, the feature flag may be enabled as follows:

```
kubectl create configmap riff-feature-flags --from-literal=stream-gateway=enabled --namespace=riff-system
```

When the feature flag is enabled:
* A service is created for each link which fronts the gRPC servers of the link's
  function pods
* The function pod consists of the main container only (with no function sidecar)

### DIFF
--- a/function-controller/pkg/controller/deployer.go
+++ b/function-controller/pkg/controller/deployer.go
@@ -31,6 +31,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
+	"strconv"
 )
 
 const (
@@ -61,8 +62,9 @@ type Deployer interface {
 }
 
 type deployer struct {
-	clientset *kubernetes.Clientset
-	brokers   []string
+	clientset                *kubernetes.Clientset
+	brokers                  []string
+	streamGatewayFeatureFlag bool
 }
 
 func (d *deployer) Deploy(link *v1.Link, function *v1.Function) error {
@@ -70,6 +72,13 @@ func (d *deployer) Deploy(link *v1.Link, function *v1.Function) error {
 	_, err := d.clientset.ExtensionsV1beta1().Deployments(link.Namespace).Create(&deployment)
 	if err != nil {
 		return err
+	}
+	if d.streamGatewayFeatureFlag {
+		service := d.buildService(link)
+		_, err = d.clientset.CoreV1().Services(link.Namespace).Create(&service)
+		if err != nil {
+			return err
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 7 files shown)


function-controller/pkg/controller/deployer.go
```diff
@@ -143,6 +159,29 @@ func (d *deployer) buildSidecarContainer(link *v1.Link, protocol string) corev1.
...
+			Namespace: link.Namespace,
+			OwnerReferences: []metav1.OwnerReference{
+				*metav1.NewControllerRef(link, v1.SchemeGroupVersion.WithKind("Link")),
+			},
+		},
+		Spec: corev1.ServiceSpec{
+			Ports: []corev1.ServicePort{corev1.ServicePort{Protocol: corev1.ProtocolTCP,
+				Port: int32(grpcPort),
+			}},
+			Selector: map[string]string{"link": link.Name, "namespace": link.Namespace}, // do not assume that labels are scoped by namespace
```
ericbottard: TODO see above

function-controller/pkg/controller/deployer.go
```diff
@@ -143,6 +159,29 @@ func (d *deployer) buildSidecarContainer(link *v1.Link, protocol string) corev1.
...
+	return corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      link.Name,
+			Namespace: link.Namespace,
+			OwnerReferences: []metav1.OwnerReference{
+				*metav1.NewControllerRef(link, v1.SchemeGroupVersion.WithKind("Link")),
+			},
+		},
+		Spec: corev1.ServiceSpec{
+			Ports: []corev1.ServicePort{corev1.ServicePort{Protocol: corev1.ProtocolTCP,
```
ericbottard: Could we name the port "grpc"

function-controller/pkg/controller/deployer.go
```diff
@@ -86,16 +95,23 @@ func (d *deployer) buildDeployment(link *v1.Link, function *v1.Function) v1beta1
 		Spec: v1beta1.DeploymentSpec{
 			Replicas: &zero,
 			Template: corev1.PodTemplateSpec{
-				ObjectMeta: metav1.ObjectMeta{Name: link.Name, Labels: map[string]string{"function": function.Name, "link": link.Name}},
+				ObjectMeta: metav1.ObjectMeta{Name: link.Name, Labels: map[string]string{"function": function.Name, "link": link.Name, "namespace": link.Namespace}},
```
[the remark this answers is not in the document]
glyn: As discussed I will add a TODO to the code and investigate in parallel with this PR being merged.

function-controller/pkg/controller/deployer.go
glyn: Done. I will force push the change.

PR #605, opened 2018-05-29
Tighten up the RBAC configuration

### LINKED ISSUE #602: Tighten up the RBAC configuration
Remove any unnecessary rules and also see if we can remove the Role and only use a ClusterRole

Using a minimal set of rules and roles makes it easier to make necessary changes going forward.

For more background see: https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-and-clusterrole

trisberg:
- remove namespaced Role and RoleBinding resources

- remove unused verbs like "patch"

- remove replicasets, we only touch the deployments

- remove RBAC configs on `make teardown`

Fixes #602


trisberg:
Rebased on the stream gateway changes

config/rbac/riff-clusterrole.yaml
```diff
@@ -8,10 +8,10 @@ rules:
     verbs: ["get", "list", "watch", "create", "delete", "update"]
   - apiGroups: ["apiextensions.k8s.io"]
```
ericbottard: I'm not even sure we need anything on CR *definitions* do we? We don't handle those programmatically; only instances of the definitions.

ericbottard:
Merged as 985e35c1, which includes additional polish as seen in https://github.com/ericbottard/riff/commit/2e09222731d47f407fb76e1bfd1399e36f7cc8ef

PR #611, opened 2018-06-05
Force re-dockerization if the last built image is no longer available

ericbottard:


ericbottard:
`docker rmi <the-image>` then `make dockerize` again (without deleting the `.dockerize` file) to validate

trisberg:
Having to use `-q` changes the developer experience, not sure this change is worth it for that, the images will get pulled from docker hub if they aren't built locally

ericbottard:
Closing w/o merging for now

PR #612, opened 2018-06-05
Update client-go to kubernetes 1.10

ericbottard:


ericbottard:
The new Knative compatible CLI already uses go-client 1.10

PR #613, opened 2018-06-05
Update build and install instructions for CLI

trisberg:


riff-cli/Makefile
```diff
@@ -21,6 +21,9 @@ test: build
 $(OUTPUT): $(GO_SOURCES) ../vendor
 	go build -ldflags "-X main.version=$(VERSION)" -o $(OUTPUT) ../cli.go
 
+install: build
+	cp $(OUTPUT) $(GOPATH)/bin
```
ericbottard: Should use `$GOBIN`

README.adoc
```diff
@@ -34,23 +34,30 @@ To build images with docker in minikube, issue:
...
-NOTE: This installs the CLI in the `bin` subdirectory of the first directory listed in `$GOPATH`, but doesn't
-set the CLI version. To set the version, see the link:riff-cli/README.adoc[riff CLI README].
 
 === Build the riff core components
 
 [source, bash]
 ----
 cd $(go env GOPATH)/src/github.com/projectriff/riff
-make dockerize
+make clean dockerize
```
ericbottard: `clean` wouldn't be necessary with https://github.com/projectriff/riff/pull/611

riff-cli/Makefile
jldec: I assume you're suggesting changing this to `cp $(OUTPUT) $(GOBIN)` for consistency with `go install`.
Unfortunately i expect that many users will not have GOBIN set, so if we do this, i suggest also adding the following at the top of the Makefile.
```
GOBIN ?= $(shell go env GOPATH)/bin
``` 
I realize that this will break when there are multiple directories in GOPATH, however that feels a lot less likely.

riff-cli/README.adoc
```diff
@@ -64,8 +64,9 @@ make -C riff-cli/ build
 == Add riff binary to PATH
 ```
 cd $(go env GOPATH)/src/github.com/projectriff/riff
-cp riff-cli/riff /usr/local/bin/riff
+make -C riff-cli/ install
 ```
+NOTE: This installs the CLI in the `bin` subdirectory of the first directory listed in `$GOPATH`.
```
jldec: I would suggest saying "This installs the CLI in $GOBIN or $GOPATH/bin"
Talking about multi-directory GOPATH is a bit confusing and not something we should recommend IMO.

riff-cli/Makefile
ericbottard: We should use `go env GOBIN` then even.

jldec: @ericbottard afict `go env GOBIN` returns nothing when GOBIN is unset, which is the case for most users. Am i missing something?

trisberg: I think we could add 
`GOBIN ?= $(GOPATH)/bin`

and use $(GOBIN) in the cp command

jldec: `GOBIN ?= $(shell go env GOPATH)/bin` will work even when GOPATH is not explicitly set, defaulting to ~/go.

trisberg: Nice, fixed in last commit

riff-cli/README.adoc
jldec: @trisberg did you want to make this NOTE consistent with the [line](#diff-7bd4a925c695c2eb0eced3872b9b965fR60) in the main README?

trisberg: Good point, fixing

PR #614, opened 2018-06-22
[WIP] Containerized `riff init`

scothis:
As an alternative to using riff-cli to initialize a riff function, we can wrap the init functionality inside a container images. The images can operate on an attached disk with the function source.

Configuration via environment:
- *RIFF_INVOKER_PATHS*: Path to the riff invoker.yaml, may be a local path or http
- *FUNCTION_ARTIFACT*: Path to the function artifact, source code or jar file (attempts detection if not specified)
- *FUNCTION_HANDLER*: Name of method or class to invoke (see specific invoker for detail)
- *FUNCTION_FILE_PATH*: Path or directory used for the function (defaults to the current directory)
- *FUNCTION_NAME*: The name of the function

### DIFF
--- a/riff-cli/cmd/initcontainer/hack.yaml
+++ b/riff-cli/cmd/initcontainer/hack.yaml
@@ -1,0 +1,1 @@
+image: github.com/projectriff/riff/riff-cli/cmd/initcontainer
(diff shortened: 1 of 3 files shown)


ericbottard:
This idea has been implemented as a "build template" in the new Knative compatible riff

PR #633, opened 2018-07-28
Add "system uninstall" command and update Knative/Istio

### LINKED ISSUE #625: Provide "riff system uninstall" command

trisberg:
- update to use latest Istio and Knative resources for 0.1.1

- don't overwrite existing Istio installation

- add new checks for install/uninstall to make sure namespaces aren't terminating

- check that namespaces exist before uninstalling

- make Istio uninstall optional

- add --force flag to install/uninstall to avoid prompts

- labeling namespaces no longer needed

- change to use fmt.Printf instead of print

Fixes #625

### DIFF
--- a/docs/riff_system_uninstall.md
+++ b/docs/riff_system_uninstall.md
@@ -1,10 +1,13 @@
 ## riff system uninstall
 
-Remove all riff and Knative system components
+Remove riff and Knative system components
 
 ### Synopsis
 
-Remove all riff and Knative system components
+Remove riff and Knative system components
+
+Use the '--istio' flag to also remove Istio components.'
+
 
 ```
 riff system uninstall [flags]
@@ -19,7 +22,7 @@ riff system uninstall [flags]
 ### Options
 
 ```
-      --force   force the removal of system components
+      --force   force the removal of components without getting any prompts
   -h, --help    help for uninstall
       --istio   include Istio and the istio-system namespace in the removal
 ```
--- a/docs/riff_system_uninstall.md
+++ b/docs/riff_system_uninstall.md
@@ -1,27 +1,27 @@
-## riff system install
+## riff system uninstall
 
-Install the riff and Knative system components
+Remove all riff and Knative system components
 
 ### Synopsis
 
-Install the riff and Knative system components
+Remove all riff and Knative system components
 
 ```
-riff system install [flags]
+riff system uninstall [flags]
 ```
 
 ### Examples
 
 ```
-  riff system install
+  riff system uninstall
 ```
 
 ### Options
 
 ```
-      --force       force the install of Knative system components for riff
-  -h, --help        help for install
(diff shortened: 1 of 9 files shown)


trisberg:
There is an option of installing riff/Knative on a cluster that already has Istio installed. The new changes makes this possible by treating Istio as a separate dependency.

We are now skipping installation of Istio if the namespace already exists and we only remove Istio and istio-system namespace when using the `--istio` flag or user confirms the removal

trisberg:
Rebased and squashed changes

jldec:
Thanks for doing this @trisberg - it will be really handy for testing or to cleanup between demos without blowing the entire cluster away.

docs/riff_system_uninstall.md
```diff
@@ -0,0 +1,37 @@
+## riff system uninstall
+
+Remove all riff and Knative system components
```
scothis: "all" is a bit misleading. We make a good faith attempt to remove everything, but cannot promise to get everything. Particularly as knative releases evolve with different names. 

This is one area where helm has an advantage by sticking an annotation on all resources it creates.

pkg/core/system.go
```diff
@@ -143,6 +255,128 @@ func waitForIstioSidecarInjector(kc *kubectlClient) error {
...
+	res, err := reader.ReadString('\n')
+	if err != nil {
+		return false, err
+	}
+	if len(res) < 2 {
+		return false, nil
+	}
+	answer := strings.ToLower(strings.TrimSpace(res))[0] == 'y'
+	return answer, nil
+}
```
scothis: nit: add a trailing newline

pkg/core/namespace.go
```diff
@@ -37,11 +39,11 @@ func (c *client) explicitOrConfigNamespace(namespaced Namespaced) string {
 
 func (kc *kubectlClient) NamespaceInit(options NamespaceInitOptions) error {
 
-	buildtemplateRelease := "https://storage.googleapis.com/riff-releases/riff-buildtemplate-0.1.0.yaml"
+	riffBuildRelease := "https://storage.googleapis.com/riff-releases/riff-build-0.1.1.yaml"
```
scothis: where did this file come from? There is no such release in the riff-build repo, nor would the [publish script](https://github.com/projectriff/riff-build/blob/master/hack/publish.sh) create a release at this path.

trisberg: It is the curated and versioned artifact based on `riff-releases/latest/riff-build.yaml` which I think is produced by the CI.

scothis: Let's use tagged releases rather than a random snapshot build. I'll cut a release.

scothis: https://storage.googleapis.com/riff-releases/previous/riff-build/riff-build-0.1.0.yaml is available for your pleasure and is from the riff-build v0.1.0 tag.

PR #636, opened 2018-07-30
Add unit tests to most commands

ericbottard:
This brings the coverage of the commands package to about 60%.

I left out the commands that interact with kubectl for later.
Need to move their writing to stdout out of the core package, and need to look into
mocking the kubectl access.


scothis:
merged manually after resolving the conflict

PR #638, opened 2018-07-30
Make cli go get'able

scothis:
go get github.com/projectriff/riff

Generally, users should use the released binaries, but having something
that can be installed via `go get` is useful for archetectures we don't
provide releases for, or CI integration.


ericbottard:
I don't get this. Why not move main.go to the top level directory (of this repo) instead?

markfisher:
I agree `go get github.com/projectriff/riff` would be nicer.

scothis:
moved

trisberg:
Fixed #638

PR #639, opened 2018-07-30
Add quotes around -w args for "riff service invoke"

### LINKED ISSUE #637: Add quotes around  -w args for "riff service invoke"
The args for curl's  -w or --write-out flag needs quoting if they contain escaped characters like `-w "\n"`

trisberg:
- quote args for curl's -w or --write-out flag if they contain `\`

- also supports arg passed in as `-w"\n"`

Fixes #637


ericbottard:
I'd like to add tests to the `service invoke` command first, but what about this proposal:

Why isn't `curlPrintArgs` *just* the shell* escaped version of `curlCmd`? Why should the `-w` flag be particular really? (I understand it has more chances of having bogus chars, but I think there is no particular reason for it to be special).

\* What exactly "shell" means here is TBD. Also, `"github.com/frioux/shellquote"` looks like it would do the job

trisberg:
Closing in favor of #641

PR #647, opened 2018-08-06
Add the option to provide env vars to services and functions

trisberg:
- the env vars are provided in a `key=value` format

- the value part can be a valueFrom expression based of k8s valueFrom field in EnvVar
  https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#envvarsource-v1-core

  example for secretKeyRef:

    --env MYSQL_PASSWORD=valueFrom:secretKeyRef:petclinic-db-mysql:mysql-root-password

### DIFF
--- a/cmd/commands/function.go
+++ b/cmd/commands/function.go
@@ -159,5 +159,7 @@ func FunctionCreate(fcTool *core.Client) *cobra.Command {
 	command.Flags().StringVar(&createFunctionOptions.Handler, "handler", "", "the name of the `method or class` to invoke, depending on the invoker used")
 	command.Flags().StringVar(&createFunctionOptions.Artifact, "artifact", "", "`path` to the function source code or jar file; auto-detected if not specified")
 
+	command.Flags().StringArrayVar(&createFunctionOptions.Env, "env", []string{""}, "environment variable expressed in a 'key=value' format. The value can be a 'valueFrom' expression")
+
 	return command
 }
--- a/cmd/commands/service.go
+++ b/cmd/commands/service.go
@@ -172,6 +172,8 @@ If an input channel and bus are specified, create the channel in the bus and sub
 
 	command.Flags().StringVar(&createServiceOptions.Image, "image", "", "the `name[:tag]` reference of an image containing the application/function")
 
+	command.Flags().StringArrayVar(&createServiceOptions.Env, "env", []string{""}, "environment variable expressed in a 'key=value' format. The value can be a 'valueFrom' expression")
+
 	return command
 }
 
--- a/pkg/core/service.go
+++ b/pkg/core/service.go
@@ -24,6 +24,7 @@ import (
 	"github.com/knative/serving/pkg/apis/serving/v1alpha1"
 	core_v1 "k8s.io/api/core/v1"
 	meta_v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 )
 
 const (
@@ -46,13 +47,18 @@ type CreateServiceOptions struct {
 
 	Image string
 
+	Env []string
+
 	DryRun bool
 }
 
 func (c *client) CreateService(options CreateServiceOptions) (*v1alpha1.Service, error) {
 	ns := c.explicitOrConfigNamespace(options.Namespaced)
 
-	s := newService(options)
+	s, err := newService(options)
+	if err != nil {
+		return nil, err
+	}
 
 	if !options.DryRun {
 		_, err := c.serving.ServingV1alpha1().Services(ns).Create(&s)
@@ -63,8 +69,9 @@ func (c *client) CreateService(options CreateServiceOptions) (*v1alpha1.Service,
 
 }
 
-func newService(options CreateServiceOptions) v1alpha1.Service {
-	return v1alpha1.Service{
+func newService(options CreateServiceOptions) (v1alpha1.Service, error) {
+
+	s := v1alpha1.Service{
 		TypeMeta: meta_v1.TypeMeta{
 			APIVersion: "serving.knative.dev/v1alpha1",
 			Kind:       "Service",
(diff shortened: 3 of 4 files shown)


pkg/core/service.go
```diff
@@ -86,6 +93,105 @@ func newService(options CreateServiceOptions) v1alpha1.Service {
...
+	if len(options.Env) > 0 {
+		var envVars []core_v1.EnvVar
+		for _, env := range options.Env {
+			envEntry := strings.SplitN(env, "=", 2)
+			if len(envEntry) != 2 {
+				return v1alpha1.Service{}, errors.New(fmt.Sprintf("unable to parse '%v', environment variables must provided as 'name=value'", env))
+			}
+			envKey := strings.TrimSpace(envEntry[0])
+			envValue := strings.TrimSpace(envEntry[1])
+			if strings.HasPrefix(envValue, "valueFrom:") {
```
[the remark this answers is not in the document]
trisberg: there is another EnvFromSource that would be confused with `env-from`, maybe use 

```
--env-var-source MYSQL_PASSWORD=secretKeyRef:petclinic-db-mysql:mysql-root-password
```

together with

```
--env-var MYSQL_USERNAME=root
```

and also add

```
--env-from configMapRef:petclinic-config
```

pkg/core/service.go
```diff
@@ -86,6 +93,105 @@ func newService(options CreateServiceOptions) v1alpha1.Service {
...
+				source := strings.Split(envValue, ":")
+				if len(source) < 3 {
+					return v1alpha1.Service{}, errors.New(fmt.Sprintf("unable to parse 'valueFrom' entry '%v', it should be provided as a secretKeyRef, configMapKeyRef, fieldRef or resourceFieldRef", env))
+				}
+				sourceType := strings.TrimSpace(source[1])
+				switch sourceType {
```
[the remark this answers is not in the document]
trisberg: should add that to what is in https://github.com/projectriff/riff/pull/636

pkg/core/service.go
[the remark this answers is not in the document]
scothis: I think `--env-var-source` and `--env-from-source` can be safely combined as the type (key ref vs map ref) cannot collide and are also invalid key names (`:` is disallowed).

single key-value:

```
--env MYSQL_USERNAME=root
```

single value from secret (or config):

```
--env-from MYSQL_PASSWORD=secretKeyRef:petclinic-db-mysql:mysql-root-password
```

all values from config map (or secret):

```
--env-from configMapRef:petclinic-config
```

all values from config map (or secret), with a prefix:

```
--env-from PETCLINIC_=configMapRef:petclinic-config
```

cmd/commands/function.go
```diff
@@ -159,5 +159,7 @@ func FunctionCreate(fcTool *core.Client) *cobra.Command {
 	command.Flags().StringVar(&createFunctionOptions.Handler, "handler", "", "the name of the `method or class` to invoke, depending on the invoker used")
 	command.Flags().StringVar(&createFunctionOptions.Artifact, "artifact", "", "`path` to the function source code or jar file; auto-detected if not specified")
 
+	command.Flags().StringArrayVar(&createFunctionOptions.Env, "env", []string{""}, "environment variable expressed in a 'key=value' format. The value can be a 'valueFrom' expression")
```
glyn: Should "environment variable" be plural?

cmd/commands/service.go
```diff
@@ -172,6 +172,8 @@ If an input channel and bus are specified, create the channel in the bus and sub
 
 	command.Flags().StringVar(&createServiceOptions.Image, "image", "", "the `name[:tag]` reference of an image containing the application/function")
 
+	command.Flags().StringArrayVar(&createServiceOptions.Env, "env", []string{""}, "environment variable expressed in a 'key=value' format. The value can be a 'valueFrom' expression")
```
glyn: Should "environment variable" be plural?

pkg/core/service.go
```diff
@@ -86,6 +93,105 @@ func newService(options CreateServiceOptions) v1alpha1.Service {
 			},
 		},
 	}
+
+	if len(options.Env) > 0 {
```
glyn: Why special case the empty case since this code isn't performance sensitive? The rest of the code should be able to cope smoothly with that case.

pkg/core/service.go
```diff
@@ -86,6 +93,105 @@ func newService(options CreateServiceOptions) v1alpha1.Service {
...
+		var envVars []core_v1.EnvVar
+		for _, env := range options.Env {
+			envEntry := strings.SplitN(env, "=", 2)
+			if len(envEntry) != 2 {
+				return v1alpha1.Service{}, errors.New(fmt.Sprintf("unable to parse '%v', environment variables must provided as 'name=value'", env))
+			}
+			envKey := strings.TrimSpace(envEntry[0])
```
glyn: I'm not keen on trimming whitespace here as it makes the documentation of the syntax (when we write it) more fussy than necessary. I don't really see why users can't learn to get the strict syntax correct from day 1.

pkg/core/service.go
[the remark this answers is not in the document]
glyn: I'd be happier if this `switch`, at least, was pulled out into a separately unit tested helper package.

pkg/core/service.go
```diff
@@ -86,6 +93,105 @@ func newService(options CreateServiceOptions) v1alpha1.Service {
...
+			envValue := strings.TrimSpace(envEntry[1])
+			if strings.HasPrefix(envValue, "valueFrom:") {
+				source := strings.Split(envValue, ":")
+				if len(source) < 3 {
+					return v1alpha1.Service{}, errors.New(fmt.Sprintf("unable to parse 'valueFrom' entry '%v', it should be provided as a secretKeyRef, configMapKeyRef, fieldRef or resourceFieldRef", env))
+				}
+				sourceType := strings.TrimSpace(source[1])
+				switch sourceType {
+				case "secretKeyRef":
```
glyn: I can't see any documentation for these values. Am I looking in the right place? If we don't document them, users won't use them and we don't need to support them. ;-)

pkg/core/service.go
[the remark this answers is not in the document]
trisberg: Nice, I like it, simplifies things a bit form a user perspective

cmd/commands/function.go
trisberg: each `--env` flag specifies a single env var although you can specify multiple `--env` flags

pkg/core/service.go
trisberg: Where would the docs go? In the `Long:` description for the command?

glyn: Yes please!

trisberg:
Closing in favor of PR #650 

I rewrote most of it based on suggestions in this review

PR #659, opened 2018-08-09
Add checks to make sure Istio components are running

### LINKED ISSUE #658: `riff system install` needs to wait for all istio pods to start
```
Installing Istio Components
Applying resources defined in: https://storage.googleapis.com/knative-releases/serving/previous/v20180809-6b01d8e/istio.yaml
Istio for riff installed
Waiting for istio-sidecar-injector to start .....'Running'
Installing Knative Components
Applying resources defined in: https://storage.googleapis.com/knative-releases/serving/previous/v20180809-6b01d8e/release-no-mon.yaml
Error from server (InternalError): error when creating "STDIN": Internal error occurred: failed calling admission webhook "pilot.validation.istio.io": Post https://istio-galley.istio-system.svc:443/admitpilot?timeout=30s: no endpoints available for service "istio-galley"
Error: exit status 1
```

trisberg:
Fixes #658

### DIFF
--- a/pkg/core/system.go
+++ b/pkg/core/system.go
@@ -27,6 +27,7 @@ import (
 	"bufio"
 	"os"
 	"strings"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
@@ -89,7 +90,7 @@ func (kc *kubectlClient) SystemInstall(options SystemInstallOptions) error {
 		}
 	}
 
-	err = waitForIstioSidecarInjector(kc)
+	err = waitForIstioComponents(kc)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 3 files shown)


pkg/core/system.go
```diff
@@ -235,28 +236,42 @@ func loadRelease(release string) ([]byte, error) {
...
+			if !strings.HasPrefix(pod.Name, "istio-") {
+				continue
+			}
+			if pod.Status.Phase != "Running" && pod.Status.Phase != "Succeeded" {
+				waitLonger = true
+				break
+			} else {
+				if pod.Status.Phase == "Running" {
+					containers := pod.Status.ContainerStatuses
+					for _, cont := range containers {
```
scothis: unless the Pod is for a Job, in which case the container will not be ready, but completed successfully

trisberg: we check for 'Succeeded' which covers the job pods, see no reason to check the containers at that point. So anything 'Running' has to have all containers 'Ready', I guess a job pod could temporarily be in that state but I don't think that would cause issues for our purposes

PR #661, opened 2018-08-09
Use a riff Istio yaml file for installs

trisberg:


scothis:
Let's have fats take it for a spin

Issue #648, opened 2018-08-07
Install manifest


scothis:
Currently the installable resources are hard coded into the riff cli binary. This makes it very difficult to test new releases. Rather than expose the URL for each resource as a cli switch, we could define a manifest file that can then reference individual yaml files. A structured manifest will make it easier to pause for validation after installing Istio, or any other component that takes time to start.

A straw man proposal:
```
version: 0.1
istio: https://path/to/istio.yaml
system:
- https://path/to/knative-build.yaml
- https://path/to/knative-serving.yaml
- https://path/to/knative-eventing.yaml
- https://path/to/stub-clusterbus.yaml
namespace:
- https://path/to/riff-buildtemplate.yaml
```

cc @trisberg

trisberg:
I agree we need to externalize this and a manifest seems better than overriding individual UTLs for the YAML artifacts

glyn:
- [ ] Switch to istio and knative arrays in the manifest
- [ ] Implement 5 second delay between installs of items in the istio array
- [ ] Implement more specific global changes for items in the istio and knative arrays (`s/type: LoadBalancer/type: NodePort/`)
- [ ] Provide a sample manifest which is equivalent to the hard-coded behaviour (in the future we can sync these up using code generation)

PR #670, opened 2018-08-10
[DNM] Add --monitoring --tracing --all --latest flags to system install/uninstall

### LINKED ISSUE #644: Provide options for tracing and monitoring for "system install"
We should make it possible to install Knative monitoring components and also Zipkin for tracing, maybe add two new flags:
```
riff system install --monitoring --tracing
```

trisberg:
Fixes #644


trisberg:
The monitoring namespace will be renamed to knative-monitoring, see: https://github.com/knative/serving/pull/1851

When that is merged we can go ahead with this PR

trisberg:
No longer needed, will be supported with the manifest support in #648

PR #678, opened 2018-08-14
Add `function build` command

### LINKED ISSUE #671: riff CLI support for function re-builds and revisions
After creating a function, i would like to modify the code, and then use the riff CLI to trigger a rebuild, and point to the new revision.

trisberg:
- add label for function name to kservice

- add annotation with build number to kservice

- support the --verbose flag

Fixes #671


trisberg:
I added some `riff.projectriff.io` metadata to the service/runLatest/revisionTemplate to keep track of things:
```
apiVersion: serving.knative.dev/v1alpha1
kind: Revision
metadata:
  annotations:
    riff.projectriff.io/build: "3"
    serving.knative.dev/configurationGeneration: "3"
  clusterName: ""
  creationTimestamp: 2018-08-14T17:00:09Z
  generation: 1
  labels:
    riff.projectriff.io/function: greeter
    serving.knative.dev/configuration: greeter
  name: greeter-00003
  namespace: default
```

trisberg:
Rebased and squashed

pkg/core/function.go
```diff
@@ -53,6 +58,19 @@ func (c *client) CreateFunction(options CreateFunctionOptions, log io.Writer) (*
...
+	if labels == nil {
+		labels = map[string]string{}
+	}
+	labels[functionLabel] = options.Name
+	s.Spec.RunLatest.Configuration.RevisionTemplate.SetLabels(labels)
+	annotations := s.Spec.RunLatest.Configuration.RevisionTemplate.Annotations
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+	annotations[buildAnnotation] = "1"
```
scothis: it looks like you're attempting to reimplement the generation id. All we really care is that the value is different. I'd rather make it clear that this label has no semantic meaning.

markfisher:
LGTM (but I know @scothis has some feedback)

pkg/core/function.go
trisberg: the way we use it now, it is a sequence, assuming no one other than the CLI is changing this annotation value

PR #679, opened 2018-08-14
add support for output channel

### LINKED ISSUE #672: CLI support for subscription replyTo
Function developers should be able to use the riff CLI to chain the output of a function to another channel for dispatching to other functions.

see https://github.com/knative/eventing/pull/325

markfisher:
fixes #672


ericbottard:
Opened https://github.com/markfisher/riff/pull/1 to only allow `--output` when `--input` is set.

Also, the same logic will need to be applied to `riff function create`

glyn:
Merged as 1fbe382394584bba08ccd9ad29031a229211865d

Issue #623, opened 2018-07-25
--verbose mode for`riff function create`


ericbottard:
_From @jldec on July 23, 2018 14:6_

Users should have observability during function builds, and error notification for `riff function create`. 

I have already bumped up against this for two different situations

1. misconfigured push secret (probably the most common gotcha making our quickstart not quick and not start)

2. when network is not great kaniko will take 5 minutes or longer to upack the node invoker image layers - it does not use cached images.

### proposed solution.
- if --verbose is set, block return from `riff function create...`
- lookup the init containers in the build pod
- while each is busy, stream the logs for that container on stdout
- bail out if one of the containers exits with error code
- or finish with with happy success message.

I have been doing the above manually mostly for the build-and-push container which is where both the above issues will bite. E.g.
```
kubectl logs -f square-00001-hl5k4 -c build-step-build-and-push
```

trisberg:
unfortunately it takes for ever to pull the build creds-init and git-init images (they are ~2.4 GB) and therefore there could be no visual feedback for several minutes, should we add a loop adding dots to a `waiting for image pull ...` message while we are waiting for the actual build to start ...

scothis:
The build image size is being tracked at https://github.com/knative/build/issues/179. The image pull delay is annoying in GKE, it's intolerable in minikube.

glyn:
# Implementation notes

- See kubectl's [logging code](https://github.com/kubernetes/kubernetes/blob/94c2c6c8423d722f436305cd67ef515a8800d723/pkg/kubectl/polymorphichelpers/logsforobject.go)

glyn:
Feedback from @jldec:
> i was expecting just build container logs - not sure if including autoscaler and other sidecars is quite right

PR #681, opened 2018-08-16
--verbose switch for function create

glyn:
Fixes https://github.com/projectriff/riff/issues/623


trisberg:
so, first attempt testing and I get an error

```
Waiting for LatestCreatedRevisionName:..................Error: function create failed: RevisionMissing: Configuration "upper" does not have any ready Revision.
```

maybe we could at that point show some info regarding that error, why did the kservice/configuration not create a revision? 

checking the `configuration.serving.knative.dev/upper` status: 
```
Status:
  Conditions:
    Last Transition Time:  2018-08-16T15:13:13Z
    Message:               Revision creation failed with message: "Internal error occurred: admission webhook \"webhook.build.knative.dev\" denied the request: mutation failed: serviceaccounts \"riff-build\" not found".
    Reason:                RevisionFailed
    Status:                False
    Type:                  Ready
```

so, it was user error, I didn't run `riff namespace init` with the secret

glyn:
@trisberg Fixed with an extra commit. The output in the above case now looks like this:
```
Waiting for LatestCreatedRevisionName:Error: function create failed: RevisionMissing: Configuration "square" does not have any ready Revision.; Revision creation failed with message: "Internal error occurred: admission webhook \"webhook.build.knative.dev\" denied the request: mutation failed: serviceaccounts \"riff-build\" not found".
```

cmd/commands/shared.go
```diff
@@ -27,4 +27,5 @@ const (
 or 'secretKeyRef' to select a key from a Secret. The following formats are supported:
   --env-from configMapKeyRef:{config-map-name}:{key-to-select}
   --env-from secretKeyRef:{secret-name}:{key-to-select}`
+	verboseUsage = "print details of command progress"
```
trisberg: should this '=' line up with the ones above? Seems that is the style used here and in other places. Not sure what the typical Go formatting for this is.

trisberg:
LGTM

This works nicely. I was able to rebase my `function build` command on this PR to add the `--verbose` option, works great.

cmd/commands/shared.go
glyn: @trisberg It's working as designed! There is one standard formatting rule, defined by `go fmt`, but sometimes it's a bit too subtle and looks wrong. See [this comment](https://github.com/golang/go/issues/7819#issuecomment-66093841)  about a similar issue to get a flavour...

glyn: @trisberg It's working as designed! There is one standard formatting rule, defined by `go fmt`, but sometimes it's a bit too subtle and looks wrong. See [this comment](https://github.com/golang/go/issues/7819#issuecomment-66093841)  about a similar issue to get a flavour...

trisberg: go figure :) thanks for the link

glyn:
Merged as 15993c36573334edf62d5207352a9999cd4dd592.

Issue #682, opened 2018-08-17
riff function create should optionally wait until the service reaches a success or error state


glyn:
By default, `riff function create` logs success after creating the ksvc:
```
$ riff function create ...
...
riff function create completed successfully
```
Subsequently, problems may occur and the ksvc may never reach a good state. The status may be checked using `riff service status`.

So that the user does not need to check the service status manually, a `--wait` switch is introduced which causes the command to block until the service reaches either a successful or an error state.

An example of success:
```
$ riff function create node square ... --wait
...
riff function create completed successfully
```
An example of an error:
```
$ riff function create node square ... --namespace badnamespace --wait
...
Error: function create failed: RevisionMissing: Configuration "square" does not have any ready Revision.; Revision "square-00001" failed with message: "build step \"build-step-build-and-push\" exited with code 1 (image: \"docker-pullable://gcr.io/kaniko-project/executor@sha256:4ba457d5c73a085358d2062152dc739ff2e96f4290412136a1efae8dca1a7501\"); for logs run: kubectl -n default logs square-00001-d2vxs -c build-step-build-and-push".
```

The wording of the completion message when `--wait` and `--verbose` are not specified, should also be adjusted, e.g.:
```
$ riff function create node square ...
...
riff function create completed successfully
Issue `riff service status square` to see the status of the function
```

PR #689, opened 2018-08-23
Add --wait switch on riff function create

glyn:
Fixes https://github.com/projectriff/riff/issues/682


glyn:
If this works out, we can do the same for other places where `--verbose` is supported.

cmd/commands/function.go
```diff
@@ -188,6 +189,13 @@ From then on you can use the sub-commands for the 'service' command to interact
 		"verbose", "v", verboseUsage,
 	).NoOptDefVal = "true"
 
+	command.Flags().VarPF(
+		BroadcastBoolValue(false,
```
ericbottard: A broadcast value only makes sense if it is applied to several variables. This can be simplified to a more traditional cobra flag

cmd/commands/function.go
```diff
@@ -137,6 +137,7 @@ From then on you can use the sub-commands for the 'service' command to interact
 				}
 			} else {
 				printSuccessfulCompletion(cmd)
+				fmt.Fprintf(cmd.OutOrStdout(), "Issue `riff service status %s` to see the status of the function\n", fnName)
```
ericbottard: Should this be printed *only* if not using `--wait/--verbose`?

cmd/commands/function.go
glyn: @ericbottard Fair enough, but I was copying the pattern employed elsewhere. I would prefer to raise an issue to apply the simplification throughout rather than "step out of line" in this PR.

cmd/commands/function.go
glyn: @ericbottard Yes. Good catch. Fixed via force push.

ericbottard:
Merged as 491d1094

PR #691, opened 2018-08-24
Support manifest on riff system install

glyn:
Fixes https://github.com/projectriff/riff/issues/648


cmd/commands/system.go
```diff
@@ -40,6 +40,16 @@ func SystemInstall(kc *core.KubectlClient) *cobra.Command {
 If an 'istio-system' namespace isn't found then the it will be created and Istio components will be installed.
 
 Use the '--node-port' flag when installing on Minikube and other clusters that don't support an external load balancer.'
+
+Use the '--manifest' flag to specify the path of a manifest file which provides the URLs of the YAML definitions of the
+components to be installed. The manifest file contents should be of the following form:
+
+version: 0.1
+istio-crds: https://path/to/istio-crds.yaml
```
ericbottard: Shouldn't this be an underscore?

glyn: Indeed. Well spotted!

docs/riff_system_install.md
```diff
@@ -10,6 +10,19 @@ If an 'istio-system' namespace isn't found then the it will be created and Istio
 
 Use the '--node-port' flag when installing on Minikube and other clusters that don't support an external load balancer.'
 
+Use the '--manifest' flag to specify the path of a manifest file which provides the URLs of the YAML files to be applied.
+The manifest file contents should be of the following form:
+
+version: 0.1
```
trisberg: the format here is different than what I see in the `cmd/commands/system.go` file.

I kind of like the grouping here, so if we could have the following:
```
istio: 
- https://path/to/istio.yaml
knative:
- https://path/to/knative-build.yaml
- https://path/to/knative-serving.yaml
- https://path/to/knative-eventing.yaml
- https://path/to/stub-clusterbus.yaml
namespace:
- https://path/to/riff-buildtemplate.yaml
```
then we would have more flexibility in including one, two ore more files for each group

pkg/core/manifest.go
```diff
@@ -0,0 +1,63 @@
...
+	"gopkg.in/yaml.v2"
+	"io/ioutil"
+)
+
+const MANIFEST_VERSION = "0.1"
+
+// Manifest defines the location of YAML files for system components.
+type Manifest struct {
+	Version         string
+	IstioCRDs       string `yaml:"istio_crds"`
```
scothis: The convention in yaml is camelCase, but I like [Thomas' suggestion](
https://github.com/projectriff/riff/pull/691/files#r212765973) better.

docs/riff_system_install.md
glyn: I was considering a grouping along those lines, especially since @scothis suggested it in the issue. However, there is some special processing which doesn't lend itself to this approach.

For istio, there is a 5 second delay between installing the CRDs and installing the other istio resources. Apart from the obvious race involved when deploying to a very sluggish system, I was trying not to break the current behaviour. I guess I could introduce a 5 second delay between all the items in the istio group, but that feels like a bit of a hack.

Similarly, there is a change applied to the serving YAML if and only if `--node-port` is specified. I am a bit nervous about applying this change to all the YAML files in the `knative` group because we can't tell what YAML files users will add to each group and whether such a global change would be correct or not.

pkg/core/manifest.go
glyn: Thanks. I'll fix this for now regardless of whether we adopt the grouping suggestion.

docs/riff_system_install.md
glyn: @trisberg As per our chat, I'll rework in terms of an istio array with 5 second intervals between each pair and a knative array with a more specific global change. Thanks for helping me work that through. /cc @scothis

scothis: The current Istio config is a hack to work around a bug that has been fixed. There are still potential issues when applying the new istio yaml, but that can be solved by a retry on failure. A retry is not a bad idea for any applied yaml.

The NodePort/LoadBalancer swap should be "relatively" safe, we can make the match more selective if there are false positives.

trisberg: how do we know when to retry? we don't really know what's in the yaml files, orwhat namespaces they target except for the istio install

scothis: > how do we know when to retry?

When the command you're running exits with a non-zero code.

glyn: The 5 second delay between istio YAMLs may be necessary if someone uses a manifest to install istio earlier than v1.0.1. I'd prefer to defer retries to a separate issue to avoid scope creep in this issue.

scothis: I'll open a separate issue for retry

pkg/core/fixtures/valid.yaml
```diff
@@ -0,0 +1,8 @@
+version: 0.1
+istio:
+  - istio-crds
```
scothis: nit: is this really valid, or just good enough pass the basic validation

pkg/core/system.go
```diff
@@ -236,7 +253,7 @@ func resolveReleaseURLs(filename string) (url.URL, error) {
 	if u.Scheme == "http" || u.Scheme == "https" {
 		return *u, nil
 	}
-	return *u, fmt.Errorf("filename must be file, http or https, got %s", u.Scheme)
+	return *u, fmt.Errorf("filename must be file, http or https, got %s", u.Scheme) // TODO: support file scheme
```
scothis: if we don't support file, remove it from the error message

pkg/core/system.go
```diff
@@ -31,7 +31,8 @@ import (
 )
 
 const (
-	istioNamespace  = "istio-system"
+	istioNamespace = "istio-system"
+	// NOTE: KEEP THE FOLLOWING IN STEP WITH manifest.yaml IN THE ROOT OF THIS REPOSITORY
```
scothis: If this is important, add a test

pkg/core/fixtures/invalid.yaml
```diff
@@ -0,0 +1 @@
+.
```
glyn: Fixed.

pkg/core/system.go
glyn: Yeah, beats codegen. Thanks.

pkg/core/system.go
glyn: Done.

pkg/core/fixtures/valid.yaml
glyn: I suspect it's valid, but the intention was that it was just enough to test the manifest parsing.

PR #692, opened 2018-08-24
Restore a Knative provided Istio

scothis:
This is a pre-release 1.0.1 version of Istio. We'll want to switch the url to a knative nightly before merging.

### DIFF
--- a/pkg/core/system.go
+++ b/pkg/core/system.go
@@ -22,18 +22,18 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"net/http"
 	"net/url"
 	"os"
 	"strings"
 	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
 	istioNamespace  = "istio-system"
-	istioCrds       = "https://storage.googleapis.com/riff-releases/istio/istio-1.0.0-riff-crds.yaml"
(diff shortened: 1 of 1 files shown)


trisberg:
Does the `istio-lean.yaml` work?

Seeing some issues like these:
```
  Warning  FailedCreate  1m (x16 over 4m)  replicaset-controller  Error creating: Internal error occurred: failed calling admission webhook "sidecar-injector.istio.io": Post https://istio-sidecar-injector.istio-system.svc:443/inject?timeout=30s: service "istio-sidecar-injector" not found
```

pkg/core/system.go
```diff
@@ -22,18 +22,18 @@ import (
...
 	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
 	istioNamespace  = "istio-system"
-	istioCrds       = "https://storage.googleapis.com/riff-releases/istio/istio-1.0.0-riff-crds.yaml"
-	istioRelease    = "https://storage.googleapis.com/riff-releases/istio/istio-1.0.0-riff-main.yaml"
+	istioRelease    = "https://github.com/knative/serving/raw/master/third_party/istio-1.0.1/istio-lean.yaml"
```
trisberg: not sure we should point to a git file that doesn't have a tag or isn't from the release bucket, it could be gone when they decide to use istio 1.0.2 instead

trisberg: and I assume we'd be better off using `istio.yaml` instead of the lean one

trisberg:
This can be closed now that we have a PR that updates all YAML files to recent releases, including Istio 1.0.1 - see #705

PR #698, opened 2018-08-28
Don't use BroadcastXXValue when not necessary

### LINKED ISSUE #694: Don't use BroadcastXXValue when not necessary
BroadcastStringValue/BroadcastBoolValue are only necessary when we want a single cobra flag (say `--namespace`) to impact several variables at once (say `&createFunctionOptions.Namespace, &createInputChannelOptions.Namespace, &createOutputChannelOptions.Namespace and &createSubscriptionOptions.Namespace`)

It is not needed when setting only a single variable. The current code can be simplified

ericbottard:
Fixes #694


trisberg:
LGTM

PR #708, opened 2018-08-29
Create a 'latest' manifest

scothis:
Allow named manifests to be defined and references via the CLI. If the
value provided by the user is not predefined, it is resolved by URL.

A new 'latest' manifest tracks the Knative nightly releases.

### DIFF
--- a/cmd/commands/system.go
+++ b/cmd/commands/system.go
@@ -18,6 +18,7 @@ package commands
 
 import (
 	"errors"
+
 	"github.com/projectriff/riff/pkg/core"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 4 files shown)


cmd/commands/system.go
```diff
@@ -79,7 +80,7 @@ knative:
 		},
 	}
 
-	command.Flags().StringVarP(&options.Manifest, "manifest", "m", "", "file path of a manifest file referring to the YAML files to be applied")
+	command.Flags().StringVarP(&options.Manifest, "manifest", "m", "default", "file path of a manifest file referring to the YAML files to be applied")
```
trisberg: should we document 'latest' as an option?

scothis: Eventually. My thought was to keep it slightly hidden for now until we feel comfortable with Istio's ability to be applied without issues.

PR #709, opened 2018-08-29
Add namespace specific files to manifest

scothis:
YAML files applied to a specific namespace should also be included in
the manifest. In the future, ideally we'd save the installed manifest
into the cluster and use that rather than forcing/allowing the user to
specify.


pkg/core/manifest.go
```diff
@@ -21,19 +21,84 @@ import (
 	"fmt"
 	"io/ioutil"
 
-	"gopkg.in/yaml.v2"
+	"github.com/ghodss/yaml"
 )
 
-const manifestVersion_0_1 = "0.1"
+const ManifestVersion = "0.1"
```
glyn: @ericbottard chose the name earlier today. This poses some questions:
1. Does the constant now need to be exported?
2. Should the constant name include the version?

I guess we can claim YAGNI for question 2 and add in the version if and only if we end up supporting multiple manifest versions in a single CLI.

cmd/commands/system.go
```diff
@@ -45,13 +45,17 @@ Use the '--node-port' flag when installing on Minikube and other clusters that d
...
-  - https://path/to/istio-release.yaml
+- https://path/to/istio-release.yaml
 knative:
-  - https://path/to/serving-release.yaml
-  - https://path/to/eventing-release.yaml
-  - https://path/to/stub-bus-release.yaml
+- https://path/to/serving-release.yaml
+- https://path/to/eventing-release.yaml
+- https://path/to/stub-bus-release.yaml
+namespace:
```
glyn: Why is this called `namespace`?

pkg/core/manifest.go
scothis: I can change the const name back.

We don't need to know the version in order to unmarshal since unknown values are ignored. Since there's an affirmative check for a valid version, we are covered. How we handle multiple manifest versions is unknown, but we can cross that bridge when we get to it.

PR #711, opened 2018-08-30
Don't use ALL_CAPS for constant.

ericbottard:
Include version as part of constant name.
Don't export it until it's actually needed.


glyn:
Merged as 602c11e94f38f7ef5b119ce0c4eda6df0afed510. Closing.

PR #713, opened 2018-08-30
Update stable manifest with latest eventing files

trisberg:
- files as of 2018-08-30


scothis:
lgtm, but will conflict with #709

PR #722, opened 2018-08-31
Create the push-credentials Secret on behalf of the user

### LINKED ISSUE #718: Create the push-credentials secret for me
While `riff namespace init` goes a long way towards making setup up a new namespace easier, it's still painful to correctly setup push credentials for your registry of choice and very error prone.

As an alternative to specifying a custom `--secret`, we should create a secret for the user from credentials they provide.

For GCR:
```bash
$ riff namespace init default --gcr ./path/to/credentials.json
```

For Docker Hub prompt for the password:
```bash
$ riff namespace init default --dockerhub username
Docker Hub password: *******
```

For non-interactive Docker Hub use accept password via stdin:
```bash
$ pbpaste | riff namespace init default --dockerhub username
```

Any user that isn't comfortable giving the riff cli their credentials can still create the secret manually. We can extend support to other registry types in the future.

### LINKED ISSUE #721: riff init namespace doesn't apply files in the correct namespace

ericbottard:
Fixes #718
Fixes #721


pkg/core/namespace.go
```diff
@@ -48,47 +64,109 @@ func (kc *kubectlClient) NamespaceInit(options NamespaceInitOptions) error {
...
 		fmt.Printf("%s\n", log)
 		if err != nil {
 			return err
 		}
 	}
-
 	return nil
 }
+
+func randomName() string {
```
trisberg: doesn't seem very random, always generates "push-credentials-fpllngz" for me

ericbottard: Ah, go random has a fixed seed by default. Need to use timestamp

scothis: use generateName https://stackoverflow.com/a/48023670

pkg/core/namespace.go
```diff
@@ -17,7 +17,20 @@
...
+	"bufio"
+	"fmt"
+	"io/ioutil"
+	"math/rand"
+	"os"
+	"syscall"
+
+	"golang.org/x/crypto/ssh/terminal"
+
+	v12 "k8s.io/api/core/v1"
```
scothis: `corev1` or something like that is a more idiomatic name

PR #725, opened 2018-09-03
Improve diagnostics for incomplete manifests

glyn:


pkg/core/manifest_test.go
```diff
@@ -75,7 +75,7 @@ var _ = Describe("Manifest", func() {
 
 			It("should return a suitable error", func() {
 				Expect(err).To(HaveOccurred())
-				Expect(err.Error()).To(HavePrefix("Manifest is incomplete:"))
+				Expect(err.Error()).To(HavePrefix("Manifest is incomplete: istio array missing: "))
```
ericbottard: Why not use `MatchError()` on the `err` variable directly?

Otherwise, LGTM

ericbottard:
You can pass a Matcher to MatchError (instead of just a string)

Le lun. 3 sept. 2018 à 14:23, Glyn Normington <notifications@github.com> a
écrit :

ericbottard:
Merged as 9893df87

PR #726, opened 2018-09-03
Avoid tail recursion

glyn:
Tail recursion can blow the stack, which is not what was intended.


glyn:
Ah, of course, sorry! I was surprised by just one retry, but that's certainly safer than a tail recursion. Closing.

PR #731, opened 2018-09-03
Image mapping

### LINKED ISSUE #702: Introduce `riff image relocate` to map the image coordinates to a user's docker registry
See https://github.com/projectriff/riff/issues/688#issuecomment-415800900 for background.

glyn:
Fixes #702.


glyn:
@scothis If we transform the manifest once, I think we'd need to store it in the cluster so that other users can use it. Otherwise, the user who ran the transformation will have to share the transformed manifest with other users.

scothis:
I don't think storing the manifest in the cluster is required here. I expect an operator to transform the manifest once when they bring a new release into their environment. The transformed manifest can be used multiple times against multiple distinct clusters.

glyn:
@scothis Let's defer the decision about where to store the transformed manifest (actually a manifest pointing to YAML files which have had their image names mapped) as well as the registry information until we need that information somewhere else (such as in `riff namespace init`).

As for whether `riff system install` is the place to do the mapping, that's my proposal for now as I don't want to complicate the UX unnecessarily. We can split out a separate operation to do mapping if we need to, but I'd like to understand the user workflows before complicating the UX.

I anticipate that when we add registry support to `riff namespace init`, we'll want to reuse the same registry parameters (hostname and user) which were used to map the system installation images.

glyn:
@jldec Yes, I can see that splitting out a separate utility will complicate the UX for the administrator in the on-prem workflow while improving the UX for other workflows and users by reducing the number of options in the `riff` CLI.

The gaps in this approach seem to be:
1. giving `riff namespace init` access to the transformed YAML
2. recording the registry parameters so that they can be applied to images, such as invoker images, which were not part of the riff system installation.

We could make the transformation step store the following things in the cluster for subsequent use: the transformed YAML for use by `riff namespace init` _and_ the registry parameters for mapping images which are not part of the riff system install.

Storing those things in the cluster would also mean we could reduce the number of parameters needed by `riff namespace init` because we would no longer have to pass it a manifest or exploded directory.

trisberg:
What about installing a new bus or a different buildtemplate? If there are no riff commands that fit, there would need to be a shared location where the operator can store transformed yaml files for everyone to use with kubectl. Seems that the simplest solution would be to have the operator store transformed yaml files for all the different kinds of yaml that are needed (invokers, buildtemplates, buses etc.), that way the files only need to be transformed once.

glyn:
It's a good point about YAML files that don't have corresponding riff commands, but how could an operator know about all these files in advance?

glyn:
Also, I was anticipating that invoker YAML files would be generated from templates and so would need transforming "on the fly".

scothis:
> I think what @scothis is suggesting is a workflow with an additional preparation step prior to running the actual install which applies config to the cluster.

Yes, I'm suggesting that remapping the manifest for a private image registry is 100% independent of the cluster. Images need to be pushed to the registry at the same time the yaml files are rewritten to use to use the updated paths. Once the operator has published the blessed images and has the amended yaml, only then should they apply that manifest to a cluster. This way that can take that same manifest and apply it to others clusters (dev, test, prod, whatever).

The decoupling of the manifest transform and applying the transformed manifest to the cluster is important.

> The only drawback i can see at first glance, is the issue @glyn mentioned with riff namespace init needing access to the transformed yaml, which would limit use of that commad to users who ran the initial prep step, and then run the actual install.

This is something that needs to happen regardless of which manifest is used to install the cluster. If a cluster is installed with riff 0.1.2 and the user has the 0.1.3 CLI installed, when initing a namespace the 0.1.2 manifest should be used by default to apply namespace specific resources.

> What about installing a new bus or a different buildtemplate? If there are no riff commands that fit, there would need to be a shared location where the operator can store transformed yaml files for everyone to use with kubectl.

We either need to create a release for every optional component of the system we support, or we need to bake those components into the monolith release but not install them by default. The process for transforming those components should identical to that of the core system.

glyn:
Closing since the requirements seem to be changing.

Issue #736, opened 2018-09-05
riff service unsubscribe


jldec:
riff CLI users need a way to undo [`riff service subscribe`](https://github.com/projectriff/riff/blob/master/docs/riff_service_subscribe.md) without resorting to kubectl.

fbiville:
Closing this issue as a new approach has been decided (see #771 and #772).

PR #739, opened 2018-09-07
Implement service unsubscribe command

fbiville:
Fixes issue #736

### DIFF
--- a/cmd/commands/service.go
+++ b/cmd/commands/service.go
@@ -55,6 +55,11 @@ const (
 )
 
 const (
+	serviceUnsubscribeSubscriptionNameIndex = iota
+	serviceUnsubscribeNumberOfArgs
+)
+
+const (
 	serviceDeleteServiceNameIndex = iota
 	serviceDeleteNumberOfArgs
 )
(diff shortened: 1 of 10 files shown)


cmd/commands/service.go
```diff
@@ -372,7 +377,7 @@ func ServiceSubscribe(fcClient *core.Client) *cobra.Command {
 
 	command := &cobra.Command{
 		Use:     "subscribe",
-		Short:   "Subscribe a service to an existing input channel",
+		Short:   "Subscribe a service to an existing input or output channel",
```
ericbottard: Actually, not an "or", as the input is required AFAIU.

More like "and an optional output channel"

cmd/commands/service.go
```diff
@@ -446,6 +451,34 @@ func ServiceDelete(fcClient *core.Client) *cobra.Command {
 	return command
 }
 
+func ServiceUnsubscribe(fcClient *core.Client) *cobra.Command {
+	deleteSubscriptionOptions := core.DeleteSubscriptionOptions{}
+
+	command := &cobra.Command{
+		Use:     "unsubscribe",
+		Short:   "Unsubscribe a service from an existing subscription",
+		Example: "  riff service unsubscribe my_subscription --namespace joseph-ns",
```
ericbottard: Can't have underscores in k8s resource names

cmd/commands/service_test.go
```diff
@@ -527,3 +527,86 @@ var _ = Describe("The riff service delete command", func() {
...
+	})
+
+	AfterEach(func() {
+		clientMock.AssertExpectations(GinkgoT())
+	})
+
+	It("should be documented", func() {
+		Expect(unsubscribeCommand.Name()).To(Equal("unsubscribe"))
+		Expect(unsubscribeCommand.Short).To(Equal(
+			"Unsubscribe a service from an existing subscription"))
```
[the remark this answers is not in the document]
fbiville: I was not 100% sure. It certainly provides little value for future changes. 
I'm going to change that as you suggested.

cmd/commands/service.go
fbiville: Right! Let me fix that!

cmd/commands/service.go
fbiville: Fixing it now.

fbiville:
Closing as a new approach has been decided (see https://github.com/projectriff/riff/issues/771 and https://github.com/projectriff/riff/issues/772).

PR #744, opened 2018-09-07
Bump Go version to 1.11

### LINKED ISSUE #719: Upgrade Go to 1.11
1.11 is the latest stable release of Go and since several of us are running this on our development machines, the [riff version of Go](https://github.com/projectriff/riff/blob/master/.travis.yml#L3) should be bumped to match.

fbiville:
Fixes #719

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-- '1.10'
+- '1.11'
 install: "go get -u github.com/golang/dep/cmd/dep"
 matrix:
   include:


.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: go
+go_import_path: github.com/projectriff/riff
```
scothis: this is the default value, do we need to override it?

fbiville: that's the only way I could build `fbiville/riff` in Travis CI.
Without `go_import_path`: https://travis-ci.org/fbiville/riff/builds/425791314
With it: https://travis-ci.org/fbiville/riff/builds/425795890

ericbottard: > forks won't have the correct env var to allow the build to pass anyway. They should provide their own CI
What do you mean @scothis ?

In any case, forks would compile in a package that is indeed `github.com/projectriff/riff`, so I think this setting is sound

scothis: Forks that are expected to merge back upstream should use the main travis config and build via PRs. Permanent forks should find their own CI tools.

The `.travis.yaml`file is not expected to work when run from any repo other than `projectriff/riff`. There are a number of environment variables that are setup within our config that a fork will not have and should never have.

I don't want to support CI for forks of riff. PR builds work great, use them. It's fine to open a PR and tag it as DNM or Do Not Merge for in progress work.

ericbottard: Ok, so as you say, to even test a [future] PR against fats, one then needs to _actually_ open a PR and make changes, push, observe, rinse, repeat. That's fine by me.

Let's drop this line then

ericbottard:
Merged as 6886d9ab

PR #745, opened 2018-09-09
Add `riff service revise` command

### LINKED ISSUE #738: Add the ability to generate new revision for a service
If the image for a service is updated we need a way to generate a new revision for the configuration created for the kservice. We should also provide the option of updating any of the following options:

*  --env stringArray        environment variable expressed in a 'key=value' format
*  --env-from stringArray   environment variable created from a source reference; see command help for supported formats
*  --image name[:tag]       the name[:tag] reference of an image containing the application/function

Maybe call the command "service revise" as in `riff service revise square --image trisberg/square:v2` since it creates a new revision

ericbottard:
Fixes #738

### DIFF
--- a/cmd/commands/service.go
+++ b/cmd/commands/service.go
@@ -36,6 +36,11 @@ const (
 )
 
 const (
+	serviceReviseServiceNameIndex = iota
+	serviceReviseNumberOfArgs
+)
+
+const (
 	serviceStatusServiceNameIndex = iota
 	serviceStatusNumberOfArgs
 )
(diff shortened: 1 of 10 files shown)


trisberg:
The only issue is that if the image doesn't use a tag that changes, there is no way to force a new version since the kservice yaml won't change but we know the image that it uses has changed in the registry. We could add an annotation like we do for `function build` to force a new revision:

```
      revisionTemplate:
        metadata:
          annotations:
            riff.projectriff.io/nonce: "2"
```

ericbottard:
Should we do that, or should we make the user realize that this maybe isn't a blessed pattern (and eg force them to set a dummy env var, or something)?

trisberg:
I think we should trust the user, if they say `service revise` we should do it without requiring a new tag or fake env var. Same as we do for `function build`, we assume that they have a new commit in the git repo and we just bump the annotation.

ericbottard:
Ping @trisberg. You may want to take a new look

PR #746, opened 2018-09-10
Don't use too obvious markdown in help that could end up printed in a terminal.

### LINKED ISSUE #730: markdown appears in online help
For example:
```
$ riff help system install
````
prints out
```    
Install riff and Knative system components

<etc.> The manifest file contents should be of the following form:

```yaml
manifestVersion: 0.1
istio:
- https://path/to/istio-release.yaml
<etc.>
````

### PR DESCRIPTION
**Pull request by ericbottard**
Use consistent formatting. Use markers that will end up being interpreted as markdown where appropriate, but still look fine in a terminal.

Fixes #730

I've ended up changing a thing or two as well. The rationale is:
- make it legible both on terminal and html
- we don't control the width of the user terminal
- be consistent

### DIFF
--- a/cmd/commands/function.go
+++ b/cmd/commands/function.go
@@ -60,8 +60,7 @@ func FunctionCreate(fcTool *core.Client) *cobra.Command {
 		Short: "Create a new function resource, with optional input and output channels",
 		Long: `Create a new function resource from the content of the provided Git repo/revision.
 
-The INVOKER arg defines the language invoker that is added to the function code in the build step. The resulting image is 
-then used to create a Knative Service (service.serving.knative.dev) instance of the name specified for the function. 
+The INVOKER arg defines the language invoker that is added to the function code in the build step. The resulting image is then used to create a Knative Service (service.serving.knative.dev) instance of the name specified for the function. 
 From then on you can use the sub-commands for the 'service' command to interact with the service created for the function. 
 
 ` + channelLongDesc + `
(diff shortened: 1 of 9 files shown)

### CODE REVIEW

File: cmd/commands/function.go [hunk 1]
```diff
@@ -60,8 +60,7 @@ func FunctionCreate(fcTool *core.Client) *cobra.Command {
 		Short: "Create a new function resource, with optional input and output channels",
 		Long: `Create a new function resource from the content of the provided Git repo/revision.
 
-The INVOKER arg defines the language invoker that is added to the function code in the build step. The resulting image is 
-then used to create a Knative Service (service.serving.knative.dev) instance of the name specified for the function. 
+The INVOKER arg defines the language invoker that is added to the function code in the build step. The resulting image is then used to create a Knative Service (service.serving.knative.dev) instance of the name specified for the function.
```
glyn: "service.serving.knative.dev" should be back-tick quoted for consistency.

ericbottard: Damn, there are more instances of this than there are pokemons :)
Will fix

cmd/commands/function.go
```diff
@@ -60,8 +60,7 @@ func FunctionCreate(fcTool *core.Client) *cobra.Command {
 		Short: "Create a new function resource, with optional input and output channels",
 		Long: `Create a new function resource from the content of the provided Git repo/revision.
 
-The INVOKER arg defines the language invoker that is added to the function code in the build step. The resulting image is 
-then used to create a Knative Service (service.serving.knative.dev) instance of the name specified for the function. 
+The INVOKER arg defines the language invoker that is added to the function code in the build step. The resulting image is then used to create a Knative Service (service.serving.knative.dev) instance of the name specified for the function. 
 From then on you can use the sub-commands for the 'service' command to interact with the service created for the function.
```
glyn: Should "service" be back-tick quoted?

glyn:
Merged as cdf25c12a5b0945c00ef56302d2f52084feaab97.

PR #749, opened 2018-09-10
Restore bash completion for riff

### LINKED ISSUE #747: Add bash completion support
When discovering a new CLI, it's very handy to use tab completion at a command prompt to list valid values that can be entered. The 0.0.x riff CLI had completion support, we should restore it for the new CLI.

scothis:
Code was lifted from riff v0.0.7.

Fixes #747

### DIFF
--- a/cmd/commands/completion.go
+++ b/cmd/commands/completion.go
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2018 the original author or authors.
+ *
+ *   Licensed under the Apache License, Version 2.0 (the "License");
+ *   you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


cmd/commands/completion.go
```diff
@@ -0,0 +1,70 @@
...
+			}
+			panic("Can't be reached")
+		},
+	}
+
+	return completionCmd
+
+}
+
+func and(functions ...cobra.PositionalArgs) cobra.PositionalArgs {
```
ericbottard: We have something else for that now, will fix

ericbottard:
Adapted and merged as bd063c03

PR #753, opened 2018-09-11
Install mockery as part of travis

ericbottard:
Fixes the current build condition (not sure how it did not fail before)
and also prevent the makefile from messing with a user's installation
without her consent

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,9 @@
 language: go
 go:
 - '1.11'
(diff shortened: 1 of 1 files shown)


ericbottard:
@glyn It's not supposed to work if you remove *your* mockery.
Quite the contrary, it assumes uses install their own (per the principle of "don't install stuff on the user station without their consent").

This fixes **travis**

PR #754, opened 2018-09-11
Add missing back-tickery

glyn:


ericbottard:
Merged as ecf312bc by also removing `\n` used as line breaks, because we shouldn't impose them on the user (let their terminal app handle the wrapping)

PR #762, opened 2018-09-12
Support path optional argument when invoking service

### LINKED ISSUE #757: `riff service invoke` should allow to address path other than /

fbiville:
Fixes #757.

### DIFF
--- a/cmd/commands/service.go
+++ b/cmd/commands/service.go
@@ -327,19 +327,19 @@ Additional curl arguments and flags may be specified after a double dash (--).`,
 		Example: `  riff service invoke square --namespace joseph-ns
   riff service invoke square /foo -- --data 42`,
 		Args: UpToDashDash(ArgValidationConjunction(
-			cobra.MinimumNArgs(serviceInvokeMaxNumberOfArgs - 1),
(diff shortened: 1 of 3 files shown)


cmd/commands/service.go
```diff
@@ -324,19 +325,21 @@ The curl command is printed so it can be copied and extended.
...
+  riff service invoke square /foo -- --data 42`,
 		Args: UpToDashDash(ArgValidationConjunction(
-			cobra.ExactArgs(serviceInvokeNumberOfArgs),
+			cobra.MinimumNArgs(serviceInvokeMaxNumberOfArgs - 1),
+			cobra.MaximumNArgs(serviceInvokeMaxNumberOfArgs),
 			AtPosition(serviceInvokeServiceNameIndex, ValidName()),
 		)),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			serviceInvokeOptions.Name = args[serviceInvokeServiceNameIndex]
+			path := args[serviceInvokeServicePathIndex]
```
ericbottard: This doesn't support the case where the arg is not set, does it?

fbiville: I'll check that with the test I should have added 😓

ericbottard:
Merged as 8c897d5c

Issue #765, opened 2018-09-13
Change image manifest to contain a map rather than an array


glyn:
The image manifest should contain a map from image name to image digest rather than simply being an array of image names.

Note that maps are unordered, but this shouldn't affect the function.

PR #766, opened 2018-09-13
Change image manifest to contain a map rather than an array

glyn:
Fixes https://github.com/projectriff/riff/issues/765

### DIFF
--- a/pkg/core/image_manifest.go
+++ b/pkg/core/image_manifest.go
@@ -29,7 +29,7 @@ const imageManifestVersion_0_1 = "0.1"
 // ImageManifest defines the image names found in YAML files of system components.
 type ImageManifest struct {
 	ManifestVersion string `json:"manifestVersion"`
-	Images          []string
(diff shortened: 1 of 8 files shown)


pkg/core/image_manifest.go
```diff
@@ -29,7 +29,7 @@ const imageManifestVersion_0_1 = "0.1"
 // ImageManifest defines the image names found in YAML files of system components.
 type ImageManifest struct {
 	ManifestVersion string `json:"manifestVersion"`
-	Images          []string
+	Images          map[string]string
```
ericbottard: Could it be useful to introduce a type alias for image coordinates and image sha, or do you think it's gonna be cumbersome?

glyn: No, I think that could disambiguate the types nicely. Thanks! I suggest `imageName` and `imageDigest` for type names.

ericbottard:
Merged with a2a69894 on top

Issue #740, opened 2018-09-07
riff system install support for manifests containing file paths


glyn:
Currently `riff system install` only supports manifests containing http(s) URLs.

PR #769, opened 2018-09-13
riff system install support for manifests containing file paths

glyn:
Fixes https://github.com/projectriff/riff/issues/740

### DIFF
--- a/pkg/core/manifest.go
+++ b/pkg/core/manifest.go
@@ -20,6 +20,8 @@ package core
 import (
 	"fmt"
 	"io/ioutil"
+	"net/url"
+	"path/filepath"
 
 	"github.com/ghodss/yaml"
 )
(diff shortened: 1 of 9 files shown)


pkg/core/manifest.go
```diff
@@ -149,3 +160,19 @@ func checkCompleteness(m Manifest) error {
...
+func checkResource(resource string) error {
+	u, err := url.Parse(resource)
+	if err != nil {
+		return err
+	}
+	if u.Scheme == "http" || u.Scheme == "https" || (u.Scheme == "" && !filepath.IsAbs(u.Path)) {
+		return nil
+	}
+
+	if u.Scheme == "" {
```
fbiville: this can be removed, the last return will then cover file paths and unsupported schemes

fbiville: although the error message will be a bit less specific

PR #773, opened 2018-09-14
Add `riff image push` command

### LINKED ISSUE #760: riff image upload
This command tags and uploads the binary images listed by an image manifest (and present on the machine where `riff` is running) to a (private) docker registry.

It should reuse `docker` either by shelling out to the `docker` binary or by programmatically driving public APIs in the implementation of `docker`. If we shell out, we could support all `docker` flags after `--` (like `riff service invoke` does for `curl`). If we drive APIs, we need to check their stability in case they aren't really designed for public consumption (Golang doesn't help here).

ericbottard:
Fixes #760

### DIFF
--- a/pkg/core/image_client.go
+++ b/pkg/core/image_client.go
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2018 The original author or authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


pkg/docker/docker.go
```diff
@@ -0,0 +1,62 @@
...
+
+// processDocker interacts with docker by spawning a process and running the `docker`
+// command line tool.
+type processDocker struct {
+	stdin  io.Reader
+	stdout io.Writer
+	stderr io.Writer
+}
+
+func (kc *processDocker) PushImage(name string, digest string, file string) error {
```
glyn: `kc` looks vestigial. How about `pd`?

pkg/core/image_client.go
```diff
@@ -0,0 +1,56 @@
...
+
+type PushImagesOptions struct {
+	Images string
+}
+
+type imageClient struct {
+	docker docker.Docker
+}
+
+func (c *imageClient) PushImages(options PushImagesOptions) error {
```
glyn: Please could you add a test (with mock docker) since we are planning on merging this to master?

glyn:
Merged as 6be4b583b2defa0a3569c748bb000aa4a3f648bc.

PR #774, opened 2018-09-14
Relocate image manifest

glyn:


glyn:
@ericbottard Regarding the final commit, please see https://github.com/projectriff/riff/issues/776#issuecomment-421415744.

cmd/commands/image.go
```diff
@@ -60,8 +61,8 @@ func ImageRelocate(c *core.Client) *cobra.Command {
     ... 
 
 `,
-		Example: `  riff image relocate --manifest=/path/to/manifest --registry=hostname --user=username --images=/path/to/image/manifest --output=/path/to/output/dir
-  riff image relocate --file=/path/to/file --registry=hostname --user=username --images=/path/to/image/manifest --output=/path/to/output`,
+		Example: `  riff image relocate --manifest=/path/to/manifest.yaml --registry=hostname --user=username --images=/path/to/image-manifest.yaml --output=/path/to/output/dir
```
ericbottard: I'd change those examples to relative paths, even though both are supported, because this seems like the most typical use

ericbottard:
Merged with no modification (but squashed) as bfccf2c3

PR #777, opened 2018-09-14
Add `riff-distro image pull` command

### LINKED ISSUE #758: riffutil image download
This command downloads binary images listed by an image manifest to the machine where riff is running.

It should reuse `docker` either by shelling out to the `docker` binary or by programmatically driving public APIs in the implementation of `docker`. If we shell out, we could support all `docker` flags after `--` (like `riff service invoke` does for `curl`). If we drive APIs, we need to check their stability in case they aren't really designed for public consumption (Golang doesn't help here).

This command is a step in building an archive which a user can use to run riff with a private registry.

Note that the downloaded images will be (docker) _proprietary_ in format. An alternative, which may be more appropriate in the future, is to download OCI standard images using `skopeo`.

ericbottard:
Fixes #758

### DIFF
--- a/cmd/commands/channel.go
+++ b/cmd/commands/channel.go
@@ -73,7 +73,7 @@ func ChannelCreate(fcTool *core.Client) *cobra.Command {
 					return err
 				}
 			} else {
-				printSuccessfulCompletion(cmd)
+				PrintSuccessfulCompletion(cmd)
 			}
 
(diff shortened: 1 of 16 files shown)


ericbottard:
Merged as d7a329ed2adb2271c9734db3f2203ff72e8f1e2a

Issue #770, opened 2018-09-13
Allow https URL for namespace init --manifest flag


trisberg:
The `riff namespace init`  command only supports filepaths. We should support http(s) style URLs as well.

PR #781, opened 2018-09-17
Allow manifest on namespace init and other commands to be URL

glyn:
This applies to the following commands:

* namespace init
* image relocate
* system install

Fixes https://github.com/projectriff/riff/issues/770

### DIFF
--- a/pkg/fileutils/read.go
+++ b/pkg/fileutils/read.go
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2018 The original author or authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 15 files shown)


pkg/fileutils/read.go
```diff
@@ -0,0 +1,106 @@
...
+	b, err := url.Parse(base)
+	if err != nil {
+		return "", err
+	}
+	if b.IsAbs() {
+		b.Path = path.Join(b.Path, file)
+		return b.String(), nil
+	}
+
+	if filepath.IsAbs(base) {
```
ericbottard: Can this be true while line 58 returned false?

glyn: No it can't. This was defensive programming in case `fileutils` grows and someone accidentally calls `readAbsFile` with a relative path.

ericbottard: I'm not sure I follow. This is `absFile`, and it doesn't interact with `readAbsFile` (`Read` does).

glyn: Apologies for answering a question you weren't asking!

`IsAbs` from the `url` package returns `true` if and only if the scheme is non-empty. So `IsAbs` will return `false` if we pass a non-URL path, regardless of whether that path is absolute or relative.

PR #784, opened 2018-09-17
Cache go modules between travis builds

scothis:


scothis:
Adding the cache cuts a full minute off the build time.

PR #783, opened 2018-09-17
Switch to go modules

ericbottard:
Fixes #752

PR #786, opened 2018-09-17
Restore go module support

scothis:
This contains the content of #783


glyn:
Should this be closed until such time as we decide to switch to go modules?

PR #787, opened 2018-09-17
Add --text and --json shortcuts to set Content-Type

scothis:
`riff service invoke` now accepts `--text` to set the request's content
type to `text/plain`, and likewise `--json` sets `application/json`.

These are shortcuts to the more formal curl syntax like
`-H "Content-Type: text/plain`. Any content type may still be set using
the formal syntax.

### DIFF
--- a/cmd/commands/service.go
+++ b/cmd/commands/service.go
@@ -377,6 +377,7 @@ Additional curl arguments and flags may be specified after a double dash (--).`,
 		Args: UpToDashDash(ArgValidationConjunction(
 			cobra.MinimumNArgs(serviceInvokeMaxNumberOfArgs-1),
 			cobra.MaximumNArgs(serviceInvokeMaxNumberOfArgs),
+			// TODO validate that at most one content-type flag is set
 			AtPosition(serviceInvokeServiceNameIndex, ValidName()),
 		)),
 		RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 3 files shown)


cmd/commands/service.go
```diff
@@ -377,6 +377,7 @@ Additional curl arguments and flags may be specified after a double dash (--).`,
 		Args: UpToDashDash(ArgValidationConjunction(
 			cobra.MinimumNArgs(serviceInvokeMaxNumberOfArgs-1),
 			cobra.MaximumNArgs(serviceInvokeMaxNumberOfArgs),
+			// TODO validate that at most one content-type flag is set
```
ericbottard: This is better done in `PreRunE` in our conventions, because `Args` is for ... args.
Will add

cmd/commands/service.go
```diff
@@ -418,6 +426,8 @@ Additional curl arguments and flags may be specified after a double dash (--).`,
 	LabelArgs(command, "SERVICE_NAME", "PATH")
 
 	command.Flags().StringVarP(&serviceInvokeOptions.Namespace, "namespace", "n", "", "the `namespace` of the service")
+	command.Flags().BoolVar(&serviceInvokeOptions.ContentTypeJson, "json", false, "set the request's content type to `application/json`")
```
ericbottard: Use of backticks in flags usage is for naming the type of the flag. See https://github.com/projectriff/riff/pull/787/files#diff-d631fc29cfe09b0098260aa1a1e0ede2R28. It's ok to use backticks in command usage though (with moderation and consistency as it has Markdown impacts. See cdf25c12a5b0945c00ef56302d2f52084feaab97)
Will change to regular ticks on merge

ericbottard:
Merged with polish as 8ae59099

PR #788, opened 2018-09-18
Selectively support configuring and creating a kubeclient.

### LINKED ISSUE #743: Enable some commands to opt out of supporting --kubeconfig and --master flags
For instance, `riff image relocate` does not need these flags and yet it can't avoid inheriting them.

ericbottard:
Fixes #743

### DIFF
--- a/docs/riff_system_install.md
+++ b/docs/riff_system_install.md
@@ -35,13 +35,6 @@ riff system install [flags]
       --node-port         whether to use NodePort instead of LoadBalancer for ingress gateways
 ```
 
-### Options inherited from parent commands
-
-```
-      --kubeconfig path   the path of a kubeconfig (default "~/.kube/config")
-      --master address    the address of the Kubernetes API server; overrides any value in kubeconfig
-```
-
 ### SEE ALSO
 
 * [riff system](riff_system.md)	 - Manage system related resources
--- a/docs/riff_system_uninstall.md
+++ b/docs/riff_system_uninstall.md
@@ -26,13 +26,6 @@ riff system uninstall [flags]
       --istio   include Istio and the istio-system namespace in the removal
 ```
 
-### Options inherited from parent commands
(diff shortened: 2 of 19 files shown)


docs/riff_system_install.md
```diff
@@ -35,13 +35,6 @@ riff system install [flags]
       --node-port         whether to use NodePort instead of LoadBalancer for ingress gateways
 ```
 
-### Options inherited from parent commands
```
scothis: should still support kubeconfig

docs/riff_system_uninstall.md
```diff
@@ -26,13 +26,6 @@ riff system uninstall [flags]
       --istio   include Istio and the istio-system namespace in the removal
 ```
 
-### Options inherited from parent commands
```
scothis: should still support kubeconfig

docs/riff_system_install.md
ericbottard: It actually doesn't (currently)

scothis: that's a bug, we should pass these flags to kubectl when we shell out to it.

ericbottard:
Let's address those when we do https://github.com/projectriff/riff/issues/803

glyn:
@ericbottard Then please could you change the FIXMEs that refer to bug 743 to refer to bug 803 *and* update bug 803 to mention this. All in all, wouldn't it be simpler to just delete the code tagged by the FIXMEs and leave bug 803 to add the new support for passing the relevant parms to kubectl?

ericbottard:
Did the latter and force pushed a rebased version

ericbottard:
This is ready for merge review, as far as I'm concerned. PTAL

glyn:
Merged as c995518217ff1e470d7e62b3d3c3f3542ca5f3a3

PR #789, opened 2018-09-18
Use path.Join instead of fmt.Sprintf

glyn:


pkg/core/image_mapper.go
```diff
@@ -48,17 +49,17 @@ func newImageMapper(mappedHost string, mappedUser string, images []imageName) (*
 			return nil, err
 		}
 
-		fullImg := fmt.Sprintf("%s/%s/%s", imgHost, imgUser, imgRepoPath)
+		fullImg := path.Join(imgHost, imgUser, imgRepoPath)
```
ericbottard: I'm not sure we want that here, because I guess on windows it does something different. This is not a file path, it's docker coordinates

glyn: Indeed, that's why I used the `path` rather than the `filepath` package.

PR #795, opened 2018-09-18
Add support for buildpacks via Build CRD and local source

### LINKED ISSUE #794: build java functions using java buildpack
This is the first step toward buildpack support across invokers.

scothis:
Adds a `riff-cnb` buildtemplate to the manifest namespace yamls.

Invokers are now distinct from runtimes. Each invoker build remains
available via `{name}-invoker` and the buildpack equivlent is available
as `{name}-buildpack`. Over time the `{name}` will switch from invokers
to buildpacks.

Fixes #794


scothis:
Also added local buildpack builds with `--local` as an alternative to `--git-repo`. Only buildpack builds can be run locally, invoker builds will error.

cmd/commands/function.go
```diff
@@ -121,9 +135,10 @@ func FunctionCreate(fcTool *core.Client) *cobra.Command {
 	command.Flags().StringVar(&createFunctionOptions.Image, "image", "", "the name of the image to build; must be a writable `repository/image[:tag]` with credentials configured")
 	command.MarkFlagRequired("image")
 	command.Flags().StringVar(&createFunctionOptions.GitRepo, "git-repo", "", "the `URL` for a git repository hosting the function code")
-	command.MarkFlagRequired("git-repo")
 	command.Flags().StringVar(&createFunctionOptions.GitRevision, "git-revision", "master", "the git `ref-spec` of the function code to use")
-	command.Flags().StringVar(&createFunctionOptions.Handler, "handler", "", "the name of the `method or class` to invoke, depending on the invoker used")
+	command.Flags().StringVarP(&createFunctionOptions.LocalPath, "local", "l", "", "path to local source to build the image from")
+	command.Flag("local").NoOptDefVal = "."
```
ericbottard: `--local` doesn't work with a space separated value then apparently:
`--local /tmp/foo` => wrong (thinks that's a 3rd arg)
`--local=/tmp/foo` => OK

cmd/commands/function.go
```diff
@@ -121,9 +135,10 @@ func FunctionCreate(fcTool *core.Client) *cobra.Command {
 	command.Flags().StringVar(&createFunctionOptions.Image, "image", "", "the name of the image to build; must be a writable `repository/image[:tag]` with credentials configured")
 	command.MarkFlagRequired("image")
 	command.Flags().StringVar(&createFunctionOptions.GitRepo, "git-repo", "", "the `URL` for a git repository hosting the function code")
-	command.MarkFlagRequired("git-repo")
 	command.Flags().StringVar(&createFunctionOptions.GitRevision, "git-revision", "master", "the git `ref-spec` of the function code to use")
-	command.Flags().StringVar(&createFunctionOptions.Handler, "handler", "", "the name of the `method or class` to invoke, depending on the invoker used")
+	command.Flags().StringVarP(&createFunctionOptions.LocalPath, "local-path", "l", "", "path to local source to build the image from")
+	command.Flag("local-path").NoOptDefVal = "."
```
ericbottard: Should handle default value in another way (if at all) as it introduces inconsistencies in how `--flags` can accept their value (see my previous comment)

scothis:
@ericbottard dropped the default value for `--local-path` it's too hard to strike the right balance of a required flag with a default value.

cmd/commands/function.go
```diff
@@ -121,9 +135,9 @@ func FunctionCreate(fcTool *core.Client) *cobra.Command {
 	command.Flags().StringVar(&createFunctionOptions.Image, "image", "", "the name of the image to build; must be a writable `repository/image[:tag]` with credentials configured")
 	command.MarkFlagRequired("image")
 	command.Flags().StringVar(&createFunctionOptions.GitRepo, "git-repo", "", "the `URL` for a git repository hosting the function code")
-	command.MarkFlagRequired("git-repo")
 	command.Flags().StringVar(&createFunctionOptions.GitRevision, "git-revision", "master", "the git `ref-spec` of the function code to use")
-	command.Flags().StringVar(&createFunctionOptions.Handler, "handler", "", "the name of the `method or class` to invoke, depending on the invoker used")
+	command.Flags().StringVarP(&createFunctionOptions.LocalPath, "local-path", "l", "", "path to local source to build the image from")
```
trisberg: --local-path /some/path throws error `Error: accepts 2 arg(s), received 1`

scothis:
@trisberg This is already in the command usage text. Do you have suggestions as to how can we make it more clear?

> Buildpack based runtimes support building from local source in addition to within the cluster. Locally built images prefixed with'dev.local/' are saved to the local Docker daemon while all other images are pushed to the registry specified in the image name.

...while we also technically support `ko.local/` local prefixes, it's not worth complicated the text.

scothis:
Travis passed, but the check here still shows as pending. Merging anyway.

Issue #796, opened 2018-09-19
riff image relocate should hard link images instead of copying the images directory


glyn:
Hard linking the `images` directory means that deleting one or more image files in the original directory will delete them from the linked directory and vice versa. So it is easy to corrupt the integrity of the original exploded archive.

A better alternative is to hard link each binary image.

glyn:
According to the man page of `ln`:
> Hard links may not span file systems

So if `riff image relocate` creates a hard link, then it will fail when the output directory is on a different file system to the input manifest and associated images.

glyn:
@scothis wrote in PR 798:
> What's the problem we're trying to solve? The referenced issue doesn't indicate why we would want to do this.

@jldec found that image relocate was very slow since it needed to copy about 3 GB of images for a typical riff install.

> Personally I would find a hard link to be surprising for a tool that is claiming to "relocate" a set of data. While the images should generally be immutable, it would be unfortunate to corrupt the original image when you thought you were modifying the relocated version.

Agreed. This is the main downside of a link.

@jldec Were you aware that modifying an image will modify the linked image, and vice versa?

> Modern filesystems will create a shallow copy by default and only allocate new disk space when the file handle is written to (copy-on-write). This has the speed of a link with the safety of a copy.

Well it appears that link is orders of magnitude faster than copy on macOS 10.13.6 which has a reasonably modern file system.

glyn:
@jldec After discussion with @scothis, I think hard links are too dangerous to employ in such a sensitive use case. The user needs to be **sure** of what bits they are deploying to their system and hard links make "surprises" possible with the wrong bits being uploaded. Closing.

PR #798, opened 2018-09-19
Hard link images in riff image relocate

glyn:
Fixes https://github.com/projectriff/riff/issues/796


scothis:
What's the problem we're trying to solve? The referenced issue doesn't indicate why we would want to do this.

Personally I would find a hard link to be surprising for a tool that is claiming to "relocate" a set of data. While the images should generally be immutable, it would be unfortunate to corrupt the original image when you thought you were modifying the relocated version.

Modern filesystems will create a shallow copy by default and only allocate new disk space when the file handle is written to (copy-on-write). This has the speed of a link with the safety of a copy.

glyn:
@scothis Good questions. Let's move this requirements discussion to the issue.

PR #799, opened 2018-09-19
Make the full Help message appear when there is an error, rather than Usage.

### LINKED ISSUE #797: Explore why the long help text is only shown when `-h` flag is used

ericbottard:
Fixes #797


fbiville:
LGTM

PR #810, opened 2018-09-20
Polish invoker definitions for riff function create

scothis:


cmd/commands/function.go
```diff
@@ -46,51 +46,44 @@ func FunctionCreate(fcTool *core.Client) *cobra.Command {
...
-			"\nBuildpack based runtimes support building from local source in addition to within the cluster. Locally built images prefixed with 'dev.local/' are saved to the local Docker daemon while all other images are pushed to the registry specified in the image name.\n" +
+			"\nThe INVOKER arg defines the language runtime that is added to the function code in the build step. The resulting image is then used to create a Knative Service (`service.serving.knative.dev`) instance of the name specified for the function. The following invokers are available:\n\n" +
+			"- 'java': uses the riff Buildpack to build Maven or Gradle projects from source\n" +
+			"- 'jar': uses riff's java-function-invoker build for a prebuilt JAR file\n" +
+			"- 'node': uses riff's node-function-invoker build\n" +
+			"- 'command': uses riff's command-function-invoker build\n" +
+			"\nBuildpack based builds support building from local source in addition to within the cluster. Locally built images prefixed with 'dev.local/' are saved to the local Docker daemon while all other images are pushed to the registry specified in the image name.\n" +
```
ericbottard: should this say "are *also* pushed to the registry? Usage of something other than `dev.local` still requires a local daemon

cmd/commands/function.go
```diff
@@ -46,51 +46,44 @@ func FunctionCreate(fcTool *core.Client) *cobra.Command {
...
+		"java": "projectriff/buildpack",
 	}
 	invokers := map[string]string{
-		"java":            "https://github.com/projectriff/java-function-invoker/raw/v0.1.1/java-invoker.yaml",
-		"java-invoker":    "https://github.com/projectriff/java-function-invoker/raw/v0.1.1/java-invoker.yaml",
-		"command":         "https://github.com/projectriff/command-function-invoker/raw/v0.0.7/command-invoker.yaml",
-		"command-invoker": "https://github.com/projectriff/command-function-invoker/raw/v0.0.7/command-invoker.yaml",
-		"node":            "https://github.com/projectriff/node-function-invoker/raw/v0.0.8/node-invoker.yaml",
-		"node-invoker":    "https://github.com/projectriff/node-function-invoker/raw/v0.0.8/node-invoker.yaml",
+		"jar":     "https://github.com/projectriff/java-function-invoker/raw/v0.1.1/java-invoker.yaml",
```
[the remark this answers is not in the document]
scothis: We need to update the projectriff/buildpack image to use the latest java invoker. @nebhale is going to show @trisberg and myself how to do that.

cmd/commands/function.go
scothis: builds are either published to a remote registry or the local docker daemon, not both

cmd/commands/function.go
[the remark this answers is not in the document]
trisberg: the buildpack adds the `invoker-exec.jar` to the start command, we need an updated buildpack to use the latest v0.1.1 of that invoker

cmd/commands/function.go
ericbottard: Builds always happen in the local daemon (and hence are available there) and *then* maybe are pushed to a (remote) registry. The code clearly uses a boolean flag named `publish`. My point is that when you use `dev.local`, there is no extra push

trisberg: I don't think non `dev.local` builds use or require the daemon, they interact directly with the registry

ericbottard: Try it :)

trisberg: I guess it does require the daemon ;)

docs/riff_function_create.md
```diff
@@ -6,17 +6,14 @@ Create a new function resource
...
-- 'detect': uses the riff Buildpack's detection (currently limited to Java functions) 
+- 'java': uses the riff Buildpack to build Maven or Gradle projects from source
+- 'jar': uses riff's java-function-invoker build for a prebuilt JAR file
+- 'node': uses riff's node-function-invoker build
+- 'command': uses riff's command-function-invoker build
 
-Classic riff Invoker runtimes are available in addition to experimental Buildpack runtimes.
-
-Buildpack based runtimes support building from local source in addition to within the cluster. Locally built images prefixed with 'dev.local/' are saved to the local Docker daemon while all other images are pushed to the registry specified in the image name.
+Buildpack based builds support building from local source in addition to within the cluster. Locally built images prefixed with 'dev.local/' are saved to the local Docker daemon while all other images are pushed to the registry specified in the image name.
```
markfisher: s/in addition to/or/

cmd/commands/function.go
markfisher: s/in addition to/or/

I would change the "Locally built..." sentence to:

> Images will be pushed to the registry specified in the image name, unless prefixed with 'dev.local/' in which case the image will only be available within the local Docker daemon."

trisberg: the non `dev.local` images are pushed to the registry and not to the local daemon, at least I can't find them

cmd/commands/function.go
```diff
@@ -46,51 +46,44 @@ func FunctionCreate(fcTool *core.Client) *cobra.Command {
...
-			// TODO make runtime help data-driven
-			"- 'java': uses riff's java-function-invoker (aliased as java-invoker)\n" +
-			"- 'node': uses riff's node-function-invoker (aliased as node-invoker)\n" +
-			"- 'command': uses riff's command-function-invoker (aliased as command-invoker)\n" +
-			"- 'java-buildpack': uses the riff Buildpack \n" +
-			"- 'detect': uses the riff Buildpack's detection (currently limited to Java functions) \n" +
-			"\nClassic riff Invoker runtimes are available in addition to experimental Buildpack runtimes.\n" +
-			"\nBuildpack based runtimes support building from local source in addition to within the cluster. Locally built images prefixed with 'dev.local/' are saved to the local Docker daemon while all other images are pushed to the registry specified in the image name.\n" +
+			"\nThe INVOKER arg defines the language runtime that is added to the function code in the build step. The resulting image is then used to create a Knative Service (`service.serving.knative.dev`) instance of the name specified for the function. The following invokers are available:\n\n" +
```
trisberg: Could we say `The INVOKER arg defines the language runtime and function invoker that is added to the function code in the build step.`

trisberg:
Can we change the local error from 

```
Error: invoker based builds are not available locally
```

to something like

```
Error: the selected invoker 'jar' does not support local builds
```

scothis:
I :+1:ed comments that have been addressed

cmd/commands/function.go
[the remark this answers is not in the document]
nebhale: [Already done](https://github.com/projectriff/riff-buildpack/commit/e5f959c9786c166ef14b5fd200e0ab345100e708), it's the building of new images that's still to be automated.  That work has taken a back seat as we're working on the caching issues but I'll make sure that it gets done eventually.

PR #813, opened 2018-09-20
Enable namespace cache for buildpack builds

trisberg:
- update stable and latest manifests to point to build-template with cache enabled

- a PersistentVolumeClaim named riff-build-cache is created in the namespace

- this requires that a default StorageClass is defined

- both GKE and Minikube has a `standard` class as the default

- the claim is for 8Gi


scothis:
This change looks ok. I provided feedback about the underlying yaml out of band

ericbottard:
Correctly seeing the JRE using the cache, as well as maven not re-downloading deps

PR #815, opened 2018-09-21
Stop trying to be smart about regenerating mocks.

### LINKED ISSUE #812: make gen-mock doesn't run when needed
I made a change to a struct that is causing the tests to panic without updated mocks. Running `make gen-mock` says there is nothing to do. It is wrong.

I shouldn't need to understand the inner working of Make in order to get it to work.

ericbottard:
Mocks are now have to be regerenated explicitly, if needed, via `make gen-mocks`.

Fixes #812

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 .PHONY: build clean test all release gen-mocks
 
 OUTPUT = ./riff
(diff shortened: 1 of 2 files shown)


glyn:
I'm pleased to see this.

Merged as 37511cbc8aa40286a8a392a3640e6ff9ce5fa4ba

PR #823, opened 2018-09-24
Bump buildpack/pack dependency

scothis:
Refs projectriff/riff-buildpack#3


scothis:
We should use a released tag of `buildpack/pack` before merging

glyn:
@scothis wrote:
> We should use a released tag of buildpack/pack before merging

Until then, I suggest we put "[DO NOT MERGE]" in the title of this PR.

Issue #806, opened 2018-09-20
`riff image load` for loading/tagging images in docker daemon


glyn:
This provides a subset of the function of riff image push, namely loading images and tagging them, but _not_ pushing them.

Syntax:
```
riff image push --images=<image-manifest>
```

PR #828, opened 2018-09-25
[DO NOT MERGE] riff image load

glyn:
Fixes https://github.com/projectriff/riff/issues/806


glyn:
Flagged as DO NOT MERGE as this is currently stacked on another PR.

Issue #811, opened 2018-09-20
riff image relocate should not shell out to `cp`


glyn:
It should use Go code instead to do the copy. Then it has a chance of working on Windows.

Note: if we used `cp -v` to provide feedback to the user, some equivalent feedback will need to be provided.

Issue #808, opened 2018-09-20
riff image relocate should provide more feedback to user


glyn:
When copying the images as the last step, there may be a pregnant pause depending on the quantity of image data to be copied and the type of file system. When there is a large amount of data and the file system is relatively slow (e.g. on macO with a fusion drive), image copying can take many seconds.

The command should print when it is about to start copying.

ericbottard:
Might be as simple as using `cp -v`

glyn:
Yeah, logically at least. Remember we plan to replace shelling out to `cp` with some Go code. ;-)

PR #830, opened 2018-09-25
[WIP, DO NOT MERGE] Replace shelling out to cp with go equivalent

glyn:
Fixes https://github.com/projectriff/riff/issues/811
Fixes https://github.com/projectriff/riff/issues/808


glyn:
Flagging as WIP since this PR is incomplete:
- [ ] inject `fileutils.Utils` into `imageClient`
- [ ] mock `fileutils.Utils` in `pkg/core/image_test.go`
- [ ] convert `fileutils_test.go` to ginkgo/gomega

Flagging as DO NOT MERGE since this PR is stacked on another.

ericbottard:
Overall, LGTM.
I'm not too keen on having something called `*Utils` that is then used in object oriented style (`*Helper` would be slightly better).
Also, I think it would be a bit better to pass a Writer to the constructor instead of a boolean flag for verbosity. One can then pass `/dev/null`, or the actual command `stdout`, or a writer for tests, etc.

So overall, I think this is definitely going into the right direction. The feedback to the user is really good I think (to the extent that we way want to add feeback about rewriting the yaml file, because now it seems only file copying is happening)

ericbottard:
Also, this PR is against a branch that is in the `projectriff` repo. I don't know if that is intended or not. I'm not sure I understand what you mean about all those PRs being stacked on top of one another (please let me know in which order you think it's best we review them)

ericbottard:
Merged as 903ad02c.

Will need to follow up with test conversion to Ginkgo, as well as some additional logging for the whole command (not just the copy part). cc @glyn

PR #835, opened 2018-09-25
Add knative-monitoring to the namespaces that system commands use

### LINKED ISSUE #834: Add knative-monitoring to namespaces used in system commands
The monitoring namespace is now renamed to knative-monitoring and we should include it in the system commands that delete or check the Knative namespaces

trisberg:
- don't show as error if namespace not found, just print as info

Fixes #834

### DIFF
--- a/pkg/core/system.go
+++ b/pkg/core/system.go
@@ -46,7 +46,7 @@ type SystemUninstallOptions struct {
 }
 
 var (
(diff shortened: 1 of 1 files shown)


glyn:
Merged as 4ed17398340f5aa06471045009778f26d21ff139.

PR #836, opened 2018-09-26
riff image load

glyn:
Fixes https://github.com/projectriff/riff/issues/806


cmd/commands/image.go
```diff
@@ -119,14 +119,55 @@ func ImageRelocate(c *core.ImageClient) *cobra.Command {
...
+
 func ImagePush(c *core.ImageClient) *cobra.Command {
 	options := core.PushImagesOptions{}
 
 	command := &cobra.Command{
 		Use:   "push",
 		Short: "Push (relocated) docker image names to an image registry",
 		Long: "Push the set of images identified by the provided image manifest into a remote registry, for later consumption by `riff system install`.\n\n" +
-			"NOTE: This command requires the `docker` command line tool, as well as a (local) docker daemon and will load and tag the images using that daemon.",
+			"NOTE: This command requires the `docker` command line tool, as well as a (local) docker daemon and will load and tag the images using that daemon.\n\n"+
```
ericbottard: technically, I don't think the daemon has to be local. Could be remote and `docker` will honor the ENV vars it uses to configure itself

ericbottard:
Merged as 0d7ad319

Issue #755, opened 2018-09-12
riff-distro system download


glyn:
This command takes a manifest as input and writes the manifest and the corresponding k8s configuration files to an output directory. The input manifest is typically one which was constructed manually and shipped by riff.

This command is a step in building an archive which a user can use to run riff with a private registry.

PR #837, opened 2018-09-26
[DO NOT MERGE] riff-distro system download

glyn:
Fixes https://github.com/projectriff/riff/issues/755


glyn:
Flagging as DO NOT MERGE since this PR is stacked on another.

distro/commands/system.go
```diff
@@ -0,0 +1,68 @@
...
+
+	command := &cobra.Command{
+		Use:   "download",
+		Short: "Download a riff system.",
+		Long: "Download the kubernetes configuration files for a given riff manifest.\n\n" +
+			"Use the `--output` flag to specify the path of a " +
+			"directory to contain the resultant kubernetes configuration files and rewritten riff manifest." +
+			"The riff manifest is rewritten to refer to the downloaded configuration files.\n",
+		Example: `  riff system download --manifest=path/to/manifest.yaml --output=path/to/output/dir`,
+		PreRunE: func(cmd *cobra.Command, args []string) error {
```
ericbottard: Should be changed to use the `Args` cobra field. Will do myself

distro/commands/wiring.go
```diff
@@ -55,8 +55,14 @@ See https://projectriff.io and https://github.com/knative/docs`,
 		ImagePull(&imageClient),
 	)
 
+	system := System()
```
ericbottard: Wondering if "system" is the best word to use. Although this is the `distro` tool, the word "system" here would have a different meaning than in `riff`.

Why not make `download` a level 1 command? This way, users run `riff-distro download`?

ericbottard:
Merged as f8b4bce6 with revised Args validation

PR #840, opened 2018-09-26
Add missing license headers and normalize all

### LINKED ISSUE #839: Some files lack proper license header

fbiville:
Fixes #839.


Makefile
```diff
@@ -40,6 +40,9 @@ release: $(GO_SOURCES) vendor VERSION
 docs: $(OUTPUT)
 	rm -fR docs && $(OUTPUT) docs
 
+license:
+	go get github.com/fbiville/header && header
```
glyn: Don't we generally try to avoid dependencies on personal git repositories?

license.json
```diff
@@ -0,0 +1,10 @@
+{
+  "headerFile": "./license-header.txt",
+  "style": "SlashStar",
+  "includes": ["**/*.go"],
+  "excludes": ["vendor/**/*"],
+  "data": {
+    "Year": "2018",
```
glyn: Applying a fixed year to the license header is unfortunately incorrect in principle and in practice.

In principle, only the files which are substantially modified should have their license headers updated. For instance, next year some would say "2018-2019", but unmodified files should have the date unchanged. See the first answer in [Do copyright dates need to be updated?](https://stackoverflow.com/questions/2390230/do-copyright-dates-need-to-be-updated) for a reasonable summary of the requirements.

Ignoring the above principle for 2018, even then it may be invalid to use "2018". For example, one of my PRs introduces files with license headers specifying "2014-2018" since I wrote the code four years ago and modified them substantially for riff.

fbiville: Interesting, I didn't know. The approach I followed is completely incorrect then.

fbiville:
Closing as the approach is incorrect.

Issue #756, opened 2018-09-12
riff-distro image list


glyn:
This command searches an input manifest and associated k8s files for image names and creates an image manifest listing the images.

It does _not_ guarantee to find all images referenced by the k8s files and so the resultant list of images needs to be validated by the user, e.g. by manual inspection or testing.

This command is a step in building an archive which a user can use to run riff with a private registry.

PR #843, opened 2018-09-27
riff-distro image list

glyn:
Take this for a spin:
```
$ riff-distro image list --images=stable-image-manifest.yaml --check=false
```
If you have a docker client/daemon, you can also run without `--check=false`, but it will take a while.

Fixes https://github.com/projectriff/riff/issues/756


glyn:
Flagged as DO NOT MERGE since this PR is stacked on another.

distro/commands/image.go
```diff
@@ -57,3 +57,41 @@ func ImagePull(c *core.ImageClient) *cobra.Command {
...
+			return nil
+		},
+	}
+	command.Flags().StringVarP(&options.Manifest, "manifest", "m", "stable", "manifest to be searched; can be a named manifest (stable or latest) or a path of a manifest file")
+	command.MarkFlagFilename("manifest", "yml", "yaml")
+
+	command.Flags().StringVarP(&options.Images, "images", "i", "image-manifest.yaml", "path of the image manifest to be created")
+	command.MarkFlagFilename("images", "yml", "yaml")
+
+	command.Flags().BoolVarP(&options.Check, "check", "", true, "check the images and omit any which are not known to docker")
```
ericbottard: Having a boolean flag whose default value is a pain to handle.
To force to false, one has to use the `--check=false` form (notice the use of the `=` sign), which we never use elsewhere.

Would be better to turn this around in some way

ericbottard: Also, what does "not known to docker" mean? Do you mean that can't be resolved/pulled?

glyn: @ericbottard 

> Would be better to turn this around in some way

@jldec had a similar  thought because of the poor usability of `--check=false`.

> Also, what does "not known to docker" mean? Do you mean that can't be resolved/pulled?

Yes.

glyn:
Removed DO NOT MERGE flagging as the PRs on which this PR was stacked have kindly been merged by @ericbottard.

PR #845, opened 2018-09-27
More idiomatic use of interfaces.

ericbottard:
This is what I originally intended with `Namespaced` and being able
to pass an instance of it to `explicitOrConfigNamespace()`


ericbottard:
Totally agree, sometimes it's best to let go :)

Merged as 9711ea39

PR #847, opened 2018-10-01
Add explanatory local path help text

matthewmcnew:
When diving into riff it was a bit unclear where local-source could be used. 

A bit of explanation would go a long way. 😄

### DIFF
--- a/cmd/commands/function.go
+++ b/cmd/commands/function.go
@@ -121,7 +121,7 @@ func FunctionCreate(fcTool *core.Client) *cobra.Command {
 	command.MarkFlagRequired("image")
 	command.Flags().StringVar(&createFunctionOptions.GitRepo, "git-repo", "", "the `URL` for a git repository hosting the function code")
 	command.Flags().StringVar(&createFunctionOptions.GitRevision, "git-revision", "master", "the git `ref-spec` of the function code to use")
-	command.Flags().StringVarP(&createFunctionOptions.LocalPath, "local-path", "l", "", "`path` to local source to build the image from")
+	command.Flags().StringVarP(&createFunctionOptions.LocalPath, "local-path", "l", "", "`path` to local source to build the image from; only the java invoker is supported at this time")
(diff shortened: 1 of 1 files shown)


cmd/commands/function.go
```diff
@@ -121,7 +121,7 @@ func FunctionCreate(fcTool *core.Client) *cobra.Command {
 	command.MarkFlagRequired("image")
 	command.Flags().StringVar(&createFunctionOptions.GitRepo, "git-repo", "", "the `URL` for a git repository hosting the function code")
 	command.Flags().StringVar(&createFunctionOptions.GitRevision, "git-revision", "master", "the git `ref-spec` of the function code to use")
-	command.Flags().StringVarP(&createFunctionOptions.LocalPath, "local-path", "l", "", "`path` to local source to build the image from")
+	command.Flags().StringVarP(&createFunctionOptions.LocalPath, "local-path", "l", "", "`path` to local source to build the image from; only the java invoker is supported at this time")
```
glyn: I'm not sure we need to talk about invokers here. Couldn't we just say "only java is supported at this time"?

scothis: Would it be clear enough if we instead say that only buildpack builds are supported?

We'll be adding more build packs over time and this message will grow stale unless we remember to keep it in sync.

glyn: I was also concerned about staleness, but I think it's also important to give users precise information. I like the idea of saying only buildpack builds are supported, especially if there is some way of finding out which those are, ideally as part of the CLI.

matthewmcnew: Avoiding staleness makes sense. 

How does this sound?
" `path` to local source to build the image from; only build-pack builds are supported at this time" 

I will update the PR.

scothis: Sounds good. We'll probably never support non-buildpack local builds again (at least in the foreseeable future), but the wiggle room is ok.

scothis:
@matthewmcnew thanks for the PR, sorry I let it slip for so long :-/

Issue #634, opened 2018-07-28
Bring Your Own Riff Invokers (?)


kingdonb:
So a couple of days ago I wandered in and found #226 which helped me run Ruby on Riff 0.0.7

Then Knative happened, and I remember hearing in some of the promotional materials that were released that all of the same invokers were supported...

https://projectriff.io/blog/riff-in-60ish-seconds/

"Riff now provides support for building function containers via a build template that uses the same invokers as Riff used before knative"

This sounds great, but the CLI help no longer provides any way to install an invoker, or really any mention of invoker at all.  This also sounds paradoxical, because Knative says Ruby is supported, while Riff has not had official Ruby support before... also, the `riff function create` help only mentions two invokers:

```
Usage:
  riff function create INVOKER FUNCTION_NAME  [flags]

Examples:
  riff function create node square --git-repo https://github.com/acme/square --image acme/square --namespace joseph-ns
  riff function create java tweets-logger
```

Can we get some comment regarding the new state of invokers in Riff and how you go about bringing your own?  I totally get that this is a new project, and I'm excited about it, but my whole team does Ruby primarily, ... the first utility of Riff for me would be to use during early development of APIs that haven't been through a design stage... so this is obviously a lot less compelling for me if I can't use it to run Ruby functions.

Hope you can understand my frustration as I only really started diving into the project about 5 days ago, and all my workflows are broken 🗡

trisberg:
I hear you, and great that you are working on a Ruby invoker!

We are still working on improving the new CLI after the move to base the runtime on Knative and the invokers are currently hard coded here: https://github.com/projectriff/riff/blob/master/cmd/commands/function.go#L47-L48

We need to make this configurable, maybe add an `--invoker-url` flag ...

trisberg:
You should also be able to test your functions by building the Docker image separately using a Dockerfile like:
```
FROM <your-invoker-image>
ARG FUNCTION_URI="<function-uri>"
ENV FUNCTION_URI ${FUNCTION_URI}
ADD <your artifact> /
```
and then pushing that image to DockerHub

You could then create a service for the function using something like:
```
riff service create my-fun --image acme/my-fun:1.0
```

kingdonb:
Alright, super! I can edit that file and give it a shot.  Also thanks for the tip about `riff service create`.

Glad to see this project has really got some serious legs again!

kingdonb:
A little rough around the edges, but your guidance was enough to make it work!

```
[...]

PR #848, opened 2018-10-01
Add custom '--invoker-url' to function create

matthewmcnew:
As referenced in https://github.com/projectriff/riff/issues/634, as a possible solution for greater invoker flexibility...

While working on/testing an update to the python invoker, adding a custom flag definitely made my life easier. 

This still maintains the 'riff function create INVOKER FUNCTION_NAME  [flags]' pattern but adds a 'custom' type that requires the '-invoker-url' flag.

### DIFF
--- a/cmd/commands/function.go
+++ b/cmd/commands/function.go
@@ -54,6 +54,8 @@ func FunctionCreate(fcTool *core.Client) *cobra.Command {
 		"node":    "https://github.com/projectriff/node-function-invoker/raw/v0.0.8/node-invoker.yaml",
 	}
 
+	flagsValidator := AtLeastOneOf("git-repo", "local-path")
+
 	command := &cobra.Command{
 		Use:   "create",
 		Short: "Create a new function resource",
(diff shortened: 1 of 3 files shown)


cmd/commands/function.go
```diff
@@ -83,6 +84,9 @@ func FunctionCreate(fcTool *core.Client) *cobra.Command {
 				createFunctionOptions.BuildpackImage = buildpack
 			} else if invokerURL, exists := invokers[invoker]; exists {
 				createFunctionOptions.InvokerURL = invokerURL
+			} else if invoker == "custom" && createFunctionOptions.InvokerURL != "" {
```
scothis: add a comment to indicate there is no work to do and not an oversight or duplication of the next line

matthewmcnew: Good point. It looks suspiciously similar to line below.

matthewmcnew:
@scothis that makes sense. 

Hopefully, this change and invokers can be be removed soon with full buildpack support.

matthewmcnew:
Updated with expiatory comment.

scothis:
@matthewmcnew thanks again

PR #854, opened 2018-10-02
riff-distro image list

glyn:
Fixes https://github.com/projectriff/riff/issues/756


glyn:
This supersedes https://github.com/projectriff/riff/pull/843 as the PRs on which 843 was stacked have now been merged.

distro/commands/image.go
```diff
@@ -57,3 +57,41 @@ func ImagePull(c *core.ImageClient) *cobra.Command {
...
+			}
+
+			commands.PrintSuccessfulCompletion(cmd)
+			return nil
+		},
+	}
+	command.Flags().StringVarP(&options.Manifest, "manifest", "m", "stable", "manifest to be searched; can be a named manifest (stable or latest) or a path of a manifest file")
+	command.MarkFlagFilename("manifest", "yml", "yaml")
+
+	command.Flags().StringVarP(&options.Images, "images", "i", "image-manifest.yaml", "path of the image manifest to be created")
```
ericbottard: The default value should be understood relative to the manifest file IMO

glyn: Agreed. Would you like to make that change as part of merging?

ericbottard: I can do that. And the inversion of `check` as well

ericbottard:
Merged as cadf6ecf with a change to make the default of image-manifest be relative to the manifest

PR #856, opened 2018-10-03
Preserve file extension during relocation

### LINKED ISSUE #852: collision avoidance should preserve `.yaml` file name suffix
When `riff image relocate` or `riff-distro system download` operate on a manifest with duplicate YAML files names, e.g. `serving/release.yaml` and `eventing/release.yaml`, the names are disambiguated in the output by adding a SHA to the end of the file name. E.g. `release.yaml-8d905d520d6f972833e50634ebe36682`. However, tooling such as editors won't recognise those files as YAML, so it would be better to preserve the file name suffix e.g. `release-8d905d520d6f972833e50634ebe36682.yaml`

fbiville:
Fixes #852.

### DIFF
--- a/pkg/core/image.go
+++ b/pkg/core/image.go
@@ -22,13 +22,12 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"github.com/ghodss/yaml"
 	"github.com/projectriff/riff/pkg/fileutils"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"reflect"
-
-	"github.com/ghodss/yaml"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/core/image.go
```diff
@@ -125,7 +127,18 @@ func sha256Flattener(input string) string {
 	}
 	hasher := sha256.New()
 	hasher.Write([]byte(input))
-	return base + "-" + hex.EncodeToString(hasher.Sum(nil))
+	base, extension := extractExtension(base)
+	return base + "-" + hex.EncodeToString(hasher.Sum(nil)) + extension
+}
+
+func extractExtension(path string) (string, string) {
```
ericbottard: I believe there could be a problem if `path` was something like `gonna.trick/you`.
Have a look at `filepath.Ext()`

fbiville: Thanks. I did not know about filepath.Ext and initially thought we were only dealing with filenames.

ericbottard: Maybe we are. But someone is bound to use that func with something else one day or another...

glyn:
Merged as a9f0e2ca71b5ac92a7488cdfe611f2a975232e71

Issue #851, opened 2018-10-02
Extend logging for riff image relocate


glyn:
Currently, copying files is logged in some detail so it looks like the command is doing only that. The command should log the other things it is doing.

PR #859, opened 2018-10-04
More logging for riff image relocate

glyn:
Fixes https://github.com/projectriff/riff/issues/851


cmd/commands/image.go
```diff
@@ -90,7 +90,7 @@ func ImageRelocate(c *core.ImageClient) *cobra.Command {
 			return nil
 		},
 		RunE: func(cmd *cobra.Command, args []string) error {
-			err := (*c).RelocateImages(options)
+			err := (*c).RelocateImages(options, cmd.OutOrStdout())
```
ericbottard: Would be better to inject the Writer into `imageClient` in `wiring.go` IMO, to be consistent with other signatures

glyn: Yes, of course, good idea. I'll look into that now.

glyn: Fixed.

ericbottard:
Merged as 6c5b7469

Issue #838, opened 2018-09-26
Make the handling of unexpected positional arguments consistent


glyn:
The `riff` and `riff-distro` commands should handle unexpected arguments in a consistent way: the commands should fail with a helpful message and the messages for the various commands should have a consistent form.

Currently some commands such as `riff channel list` fail with the terse message:
```
Error: accepts 0 arg(s), received 1
```
Others commands such as `riff image relocate` fail with a longer message such as:
```
Error: the `image relocate` command does not support positional arguments
```
Still others commands such as `riff system install` ignore the extra arguments and proceed.

ericbottard:
Cobra handles argument (not flags) validation via a "standard" mechanism: the `Args` [field](https://github.com/spf13/cobra/blob/8d114be902bc9f08717804830a55c48378108a28/command.go#L61). So argument validation should use that, always (instead of `PreRun()`, which we decided to use for *flags* validation, given that Cobra does not provide a standard hook for that).
Second, Cobra comes with a (small) set of validators [out-of-the-box](https://github.com/spf13/cobra/blob/8d114be902bc9f08717804830a55c48378108a28/args.go). So those should be used when they're applicable, even though arguably they're not "great" (the messages are a little sparse). But that's a first start towards consistency.

glyn:
Adopt the Cobra-standard code from `riff channel list` in:
- [ ] `riff image relocate`
- [ ] `riff image load`
- [ ] `riff image push`
- [ ] `riff system install`
- [ ] `riff system uninstall`
- [ ] `riff version`
- [ ] `riff-distro image pull`
- [ ] `riff-distro image list`

PR #862, opened 2018-10-04
Handle unexpected positional args consistently

glyn:
Fixes https://github.com/projectriff/riff/issues/838


ericbottard:
Merged as 38948925

PR #865, opened 2018-10-04
Replace check flag with no-check for riff-distro image list

### LINKED ISSUE #853: Improve usability of riff-distro image list --check=false
Currently it is necessary to add the value `false` as in `--check=false`. If we change the flag to `--no-check` and give it a default value (when the flag is not specified) of `false` and a `NoOptDefVal` (when the flag is specified without a value) of `true`, then it won't be necessary to use an equal sign.

fbiville:
Fixes #853.

### DIFF
--- a/pkg/core/image_client.go
+++ b/pkg/core/image_client.go
@@ -55,7 +55,7 @@ type ListImagesOptions struct {
 	Manifest string
 	Images   string
 	Force    bool
-	Check    bool
+	NoCheck  bool
 }
 
 type imageLister func(resource string, baseDir string) ([]string, error)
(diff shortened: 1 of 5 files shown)


pkg/core/image_client.go
```diff
@@ -55,7 +55,7 @@ type ListImagesOptions struct {
 	Manifest string
 	Images   string
 	Force    bool
-	Check    bool
+	NoCheck  bool
```
glyn: Not sure about changing this variable. It's slightly easier to reason about `Check` being true or false rather than `NoCheck`. In general, I think the command layer should deal with the UX and the core layer should be biassed towards the implementors.

fbiville: I fully agree with that and I'm not of double negation either...
However, I've had trouble to bind a cobra flag to a simple bool variable (or even better, have a custom binding that would negate the flag value directly into the option struct field).

glyn: Ah, I see your predicament. Let's stick with `NoCheck` then! :-)

Issue #864, opened 2018-10-04
Introduce a value object/struct for Docker image coordinates


ericbottard:
This may help to reason about the code, and maybe even https://github.com/projectriff/riff/issues/849

glyn:
The [docker `Reference` type](https://github.com/docker/distribution/blob/master/reference/reference.go) seems to be the best starting point.

/cc @ImJasonH @mattmoor

PR #869, opened 2018-10-05
[Early review only - DO NOT MERGE] docker image reference types - WIP

glyn:
Fixes https://github.com/projectriff/riff/issues/864

### DIFF
--- a/pkg/core/image.go
+++ b/pkg/core/image.go
@@ -22,12 +22,14 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"github.com/ghodss/yaml"
-	"github.com/projectriff/riff/pkg/fileutils"
 	"io/ioutil"
 	"os"
 	"path/filepath"
 	"reflect"
+
+	"github.com/docker/distribution/reference"
+	"github.com/ghodss/yaml"
+	"github.com/projectriff/riff/pkg/fileutils"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


glyn:
This PR is just to surface my current work until I return from vacation. The code is far from complete and many tests fail. The changes are limited in scope and need to be expanded out so that images are converted to `Reference` types as soon as they are encountered by the `core` package.

The big change is that normalisation ensures that some of the edge cases previously coded for will no longer occur. This is the reason that many of the tests fail.

Any comments gratefully received.

pkg/core/image.go
```diff
@@ -239,8 +241,12 @@ func relocateImageManifest(imageManifestPath string, mapper *imageMapper, output
...
 
 func applyMapper(name imageName, mapper *imageMapper) imageName {
-	quotedName := imageName(mapper.mapImages([]byte(fmt.Sprintf("%q", name))))
-	return quotedName[1 : len(quotedName)-1]
+	mapped := string(mapper.mapImages([]byte(fmt.Sprintf("%q", name.String()))))
+	ref, err := reference.Parse(mapped[1 : len(mapped)-1])
+	if err != nil {
+		panic(err) // TODO: handle this gracefully
+	}
+	return imageName(ref)
```
ericbottard: I had already forgotten that we already had an `imageName` (opaque) type...

PR #873, opened 2018-10-08
Bump stable manifest versions

### LINKED ISSUE #870: Use separate build and serving install artifacts
Build and Serving have been decoupled. We can now install serving without build or a different/newer version of build. (Once the release artifacts are built tonight, anyway)

https://github.com/knative/serving/pull/2162#issue-220768702

The stable and latest manifests should be updated to use the stand alone `serving/serving.yaml` and a `build/release.yaml`.

Note, the current `serving/release.yaml` will continue to contain both serving and build for the immediate future.

scothis:
Build can now be released independently of serving, we will no longer
need serving to update the build release for us.

Fixes #870

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -1,10 +1,11 @@
 manifestVersion: 0.1
 istio:
-- https://storage.googleapis.com/knative-releases/serving/previous/v20181002-536fa6b/istio.yaml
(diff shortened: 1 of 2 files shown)


scothis:
I'm not seeing the build-controller create a pod for a build resource with this change.

scothis:
The build/serving issue should be resolved by tonight's builds.

scothis:
Updated to last nights nightlies.

@trisberg ready for review

PR #885, opened 2018-10-18
Generate riff-distro documentation

### LINKED ISSUE #881: Check in riff-distro docs
This should be done analogously to the `riff` docs.

The docs could appear either in `docs/distro` or `distro/docs`. The former may make them easier to find; the latter could provide better separation of docs for different audiences.

fbiville:
Fixes #881

### DIFF
--- a/cmd/commands/docs.go
+++ b/cmd/commands/docs.go
@@ -17,10 +17,6 @@
 package commands
 
 import (
-	"os"
-
-	"fmt"
-
 	"github.com/spf13/cobra"
 	"github.com/spf13/cobra/doc"
 )
@@ -42,15 +38,8 @@ func Docs(rootCmd *cobra.Command, fs Filesystem) *cobra.Command {
 }
 
 func GenerateDocs(rootCommand *cobra.Command, directory string, fs Filesystem) error {
-	fileinfo, err := fs.Stat(directory)
-	if os.IsNotExist(err) {
-		if err = fs.Mkdir(directory, 0744); err != nil {
-			return err
-		}
-	} else if err != nil {
+	if err := fs.MkdirAll(directory, 0744); err != nil {
 		return err
-	} else if !fileinfo.Mode().IsDir() {
-		return fmt.Errorf("path %q already exists but is not a directory", directory)
 	}
 	return doc.GenMarkdownTree(rootCommand, directory)
 }
--- a/cmd/commands/docs.go
+++ b/cmd/commands/docs.go
@@ -25,7 +25,7 @@ import (
 	"github.com/spf13/cobra/doc"
 )
 
-func Docs(rootCmd *cobra.Command) *cobra.Command {
+func Docs(rootCmd *cobra.Command, fs Filesystem) *cobra.Command {
 
 	var directory string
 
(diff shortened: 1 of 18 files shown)


fbiville:
For early review, I still plan to reconcile the two slightly different implementations.

distro/commands/docs.go
```diff
@@ -0,0 +1,53 @@
...
+		Use:    "docs",
+		Short:  "generate riff-distro command documentation",
+		Hidden: true,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			fileinfo, err := fs.Stat(directory)
+			if os.IsNotExist(err) {
+				if err = fs.Mkdir(directory, 0744); err != nil {
+					return err
+				}
+				fileinfo, err = fs.Stat(directory)
```
fbiville: Without the line above, the doc generation will fail if the docs folder is not present, which makes sense.
I do not understand how the other impl does not need that.

distro/commands/docs.go
```diff
@@ -0,0 +1,53 @@
...
+			}
+
+			if err != nil {
```
fbiville: this bit is missing in the pre-existing implementation, thus discarding possible os.Stat errors

distro/commands/wiring_test.go
```diff
@@ -0,0 +1,34 @@
...
+		It("including `riff-distro` docs", func() {
+			errMsg := "`%s` should be wired to root command"
+			Expect(find(rootCommand, "docs")).NotTo(BeNil(), fmt.Sprintf(errMsg, "docs"))
+		})
+
+	})
+
+})
+
+func find(command *cobra.Command, names ...string) *cobra.Command {
```
fbiville: it's a copy of the other wiring test, I will define it in a single place instead

cmd/commands/docs.go
```diff
@@ -34,19 +34,23 @@ func Docs(rootCmd *cobra.Command) *cobra.Command {
...
 		},
 	}
 	docsCmd.Flags().StringVarP(&directory, "dir", "d", "docs", "the output directory for the docs.")
 	return docsCmd
 }
+
+func GenerateDocs(rootCommand *cobra.Command, directory string, fs Filesystem) error {
+	fileinfo, err := fs.Stat(directory)
+	if os.IsNotExist(err) {
+		if err = fs.Mkdir(directory, 0744); err != nil {
```
ericbottard: Might as well call this (and use in the underlying implementation) `MkDirAll`

cmd/commands/wiring.go
```diff
@@ -220,5 +220,12 @@ func installKubeConfigSupport(command *cobra.Command, client *core.Client, kc *c
 		}
 		return nil
 	}
+}
 
+func FindSubcommand(command *cobra.Command, names ...string) *cobra.Command {
```
ericbottard: Can this be made private ?

fbiville: It's used in `github.com/projectriff/riff/cmd/commands` and `github.com/projectriff/riff/distro/commands`

fbiville: It's used in `github.com/projectriff/riff/cmd/commands` and `github.com/projectriff/riff/distro/commands`

fbiville: It's used in `github.com/projectriff/riff/cmd/commands` and `github.com/projectriff/riff/distro/commands`

fbiville: It's used in `github.com/projectriff/riff/cmd/commands` and `github.com/projectriff/riff/distro/commands`

cmd/commands/docs.go
fbiville: The UI does not show anything but it's now done (cf. status.github.com, they currently have DB issues)

cmd/commands/wiring.go
ericbottard: Maybe move to `cobra.go` then

ericbottard: Maybe move to `cobra.go` then

fbiville: 👍

ericbottard:
Merged as 73bd71f4

PR #891, opened 2018-10-19
docker image reference types and associated refactoring and new tests

glyn:
The previous code’s assumption about the structure of an image name was sadly lacking. For instance:
```
k8s.gcr.io/fluentd-elasticsearch:v2.0.4
```

has a surprisingly short path component. The previous code assumes the path
(of a canonicalised reference) is always of the form:
```
<user>/<rest of path>
```

but this isn't always the case as the above example shows. The previous code
happened to "work", but more by luck than management. The new code applies
flattening in this case to avoid collisions. The above example then maps to:
```
reg.org/user/fluentd-elasticsearch-e71ae1e56d9a15bbcf26be734d13b608
```

Fixes https://github.com/projectriff/riff/issues/864


ericbottard:
Gave this a very quick glance. I think at first as was expecting to see the introduction of a value type which would be very similar (if not a type alias) of docker's `Reference`. Meaning with a name AND a digest (albeit optional). This PR seems to always split the two, and also maybe re-implement stuff that we could get for free from `Reference`?

glyn:
> Gave this a very quick glance.

Thanks!

> I think at first [I] was expecting to see the introduction of a value type which would be very similar (if not a type alias) of docker's Reference. Meaning with a name AND a digest (albeit optional).

I started off with a type alias of docker's `Reference`, but that didn't allow me to add my own convenience methods since `Reference` is an interface and interface types cannot be method receivers.

The next step was to use a `struct` embedding a docker `Reference`, but that had some unfortunate effects as it could be used in some contexts where a `Reference` was expected, but actually malfunctioned. I'm afraid I didn't explore the problem deeply because the failure was obscure and depended on how the type was _used_, so it seemed too dangerous for general use.

So I decided to use a `struct` with a private field of type `Reference` to provide more control. I moved this type to its own package so that its private members could not be (accidentally) abused by the `core` package.

> This PR seems to always split the two, and also maybe re-implement stuff that we could get for free from Reference?

It's true that the docker `Reference` type can include an optional digest, but I need to be able to associate a digest with an `image.Name` in the case where the input string form of the name _does not include a digest_. Because of the way `image.Name` is used in image relocation, the string form must match the string used to construct the value, so we can't artificially add a digest to an `image.Name` which didn't start out with one.

Another way of putting this is that a `Reference` (and therefore also an `image.Name`) _refers to_ an image and includes a name, an optional tag, and an optional digest. The referenced image always has a digest, but the digest isn't intrinsic to the `Reference`.

If the above points are unclear or unconvincing, let's chat next week before you continue your review.

ericbottard:
Merged as two different commits, db57f384 and parent.
I amended the second one to try to further clarify the difference between Digest and Id. See https://github.com/projectriff/riff/blob/db57f38406599cdb7cf8525cd0b8dc9365549132/pkg/image/digest.go#L22-L28

PR #906, opened 2018-10-29
Make the 'command' & node languages use buildpacks

### LINKED ISSUE #905: Support 'command' invoker via buildpacks

ericbottard:
Expose artifact and handler as BuildTemplate parameters

Fixes #905

### DIFF
--- a/pkg/core/function.go
+++ b/pkg/core/function.go
@@ -133,6 +133,8 @@ func (c *client) CreateFunction(options CreateFunctionOptions, log io.Writer) (*
 					Name: "riff-cnb",
 					Arguments: []build.ArgumentSpec{
 						{Name: "IMAGE", Value: options.Image},
+						{Name: "FUNCTION_ARTIFACT", Value: options.Artifact},
+						{Name: "FUNCTION_HANDLER", Value: options.Handler},
 						// TODO configure buildtemplate based on buildpack image
 						// {Name: "TBD", Value: options.BuildpackImage},
 					},
--- a/configuration.go
+++ b/configuration.go
@@ -34,12 +34,12 @@ This file is defined in the root of this repository and symbolically linked from
 
 var (
 	buildpacks = map[string]string{
-		"java": "projectriff/buildpack",
+		"java":    "projectriff/buildpack",
+		"command": "projectriff/buildpack",
+		"node":    "projectriff/buildpack",
 	}
 	invokers = map[string]string{
-		"jar":     "https://github.com/projectriff/java-function-invoker/raw/v0.1.1/java-invoker.yaml",
-		"command": "https://github.com/projectriff/command-function-invoker/raw/v0.0.7/command-invoker.yaml",
-		"node":    "https://github.com/projectriff/node-function-invoker/raw/v0.0.8/node-invoker.yaml",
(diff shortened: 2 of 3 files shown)


ericbottard:
Rebased on latest master

Blocked by https://github.com/projectriff/riff-buildpack-group/pull/3 (or we could push a manual version of the builder for now)

main.go
```diff
@@ -48,7 +48,7 @@ var (
 			},
 			Namespace: []string{
 				"https://storage.googleapis.com/riff-releases/latest/riff-build.yaml",
-				"https://storage.googleapis.com/riff-releases/riff-cnb-buildtemplate-0.1.0.pre.3.yaml",
+				"https://raw.githubusercontent.com/projectriff/riff-buildpack-group/cf202c43b5695a74185929e393e5c1005e6f1f91/riff-cnb-buildtemplate.yaml",
```
scothis: We should keep release files like this on GCS, I uploaded the file as

```suggestion
				"https://storage.googleapis.com/riff-releases/riff-cnb-buildtemplate-0.1.1.pre.1.yaml",
```

pkg/core/function.go
```diff
@@ -133,6 +133,8 @@ func (c *client) CreateFunction(options CreateFunctionOptions, log io.Writer) (*
 					Name: "riff-cnb",
 					Arguments: []build.ArgumentSpec{
 						{Name: "IMAGE", Value: options.Image},
+						{Name: "FUNCTION_ARTIFACT", Value: options.Artifact},
```
scothis: 👍

scothis:
We'll also need to add artifact and handler support for local builds, but we can add that in a separate PR

scothis:
can you run `make docs` and include the updated output?

main.go
ericbottard: Agreed. This is really to unstuck the whole thing for now.
As a matter of fact, the CI for the buildpack group could reference timestamped tagged versions of the docker image (and push that to GCS)

distro/main.go
```diff
@@ -40,7 +40,7 @@ var (
 			},
 			Namespace: []string{
 				"https://storage.googleapis.com/riff-releases/latest/riff-build.yaml",
-				"https://storage.googleapis.com/riff-releases/riff-cnb-buildtemplate-0.1.0.yaml",
+				"https://storage.googleapis.com/projectriff/riff-buildtemplate/riff-cnb-buildtemplate-0.0.1-snapshot-ci-64830c3bbc6503beafdae382ead115806fa100ca.yaml",
```
glyn: What will stop us shipping this snapshot as part of a release?

glyn: Including a snapshot in `stable` could cause problems if it's difficult to rebuild at the correct version (which would make it hard to apply a point fix for a bug without getting a load of other changes that are not necessarily desirable).

Including a snapshot in `latest` does seem counter-intuitive as you point out. I guess it depends on the definition of `latest` - does that just apply to knative components or is it meant to include riff components too?

ericbottard: It's a *pinned* snapshot (as seen in the commit sha that is part of the name).
I'm fine switching it to a "tracking" snapshot for now (and then we can pin / release it when we release `riff`)

scothis: I agree with Eric, it's no different than the pinned snapshot builds we use for knative.

PR #909, opened 2018-10-29
 Update to artifacts from v20181029

### LINKED ISSUE #907: Update "stable" to recent release artifacts

### LINKED ISSUE #908: Function build with "latest" manifest seems to fail when using "--verbose" flag
riff function build appears to fail when using `--verbose` flag - the message is:

```
Error: function creation failed: RevisionMissing: Configuration "upper" does not have any ready Revision.; Revision "upper-00001" failed with message: "The target is not receiving traffic.".
```

The function does start though, so it might be some change in the reporting of Knative build status.

trisberg:
Function create fails unless we ignore the "scale to zero" status message that we now see between the build and the revision being ready

Fixes #907 
Fixes #908

### DIFF
--- a/pkg/core/function.go
+++ b/pkg/core/function.go
@@ -327,6 +327,11 @@ func (c *client) waitForSuccessOrFailure(namespace string, name string, gen int6
 						someStateIsUnknown = true
 						break
 					}
+					// ignore any "scaled to zero" status messages - wait for the revision to start
+					if strings.Contains(c.Message, "The target is not receiving traffic") {
+						someStateIsUnknown = true
+						break
+					}
 				}
 			}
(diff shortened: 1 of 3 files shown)


pkg/core/function.go
```diff
@@ -327,6 +327,11 @@ func (c *client) waitForSuccessOrFailure(namespace string, name string, gen int6
 						someStateIsUnknown = true
 						break
 					}
+					// ignore any "scaled to zero" status messages - wait for the revision to start
```
scothis: I assume we can remove this after https://github.com/knative/serving/issues/2346 is resolved

trisberg: Added a TODO

PR #913, opened 2018-10-30
Inject CLI config for manifests, invokers and buildpacks

### LINKED ISSUE #912: Allow cli config to be injected at build time
The riff cli bakes a lot of configuration into the cli binary. These values are currently hard coded, but should be opened up to allow for custom riff builds with alternate configurations.

In particular the config currently at:
- https://github.com/projectriff/riff/blob/215060a2fcfe2b89b03f37b311bd3389e6c9021c/pkg/core/manifest.go#L32
- https://github.com/projectriff/riff/blob/215060a2fcfe2b89b03f37b311bd3389e6c9021c/cmd/commands/function.go#L48
- https://github.com/projectriff/riff/blob/215060a2fcfe2b89b03f37b311bd3389e6c9021c/cmd/commands/function.go#L51

Each of these should be injected into the commands from the `main` via the `commands.CreateAndWireRootCommand()` method.

scothis:
Manifests, invokers and buildpacks are configuration that is built into
the riff cli. It should be easy to create a custom build of riff with a
different configuration. By moving the config into a central location
and then passing that configuration into other component, it will be
easier to customize in the future.

Fixes #912


scothis:
rebased

scothis:
discussed with @glyn in #914

PR #914, opened 2018-10-30
Parameterize 'riff' name

scothis:
It should be eaiser to interact with the riff cli when it has a
different name. Most instances of "riff" that would appear on the cli
have been parameterized with a value that can be defined at build time.

riff-distro has also been updated. Where there was no difference between
riff and riff-distro commands the distro command has been replaced with
the riff command.

This change attempted to preserve the current docs directory content,
unless the content was incompatible with a parameterized world. For
example, the listing of invokers/buildpacks is too bespoke to properly
express.

Refs: #912


scothis:
> How would just having a different name make it easier to interact with?

It makes it easier to either create a white labeled "riff" cli, or to embed riff commands into another cli. There is a lot of help text that assumed the name of the cli command was `riff` which may not always be true. This PR makes it possible to override the cli name.

glyn:
Merged as 1a7fcf2f40b6103be918137e635e3a8e64278983

PR #921, opened 2018-10-31
Add delete of "horizontalpodautoscaler.autoscaling/istio-pilot" resource

### LINKED ISSUE #918: The "riff system uninstall" leaves "horizontalpodautoscaler.autoscaling/istio-pilot" in default namespace
We should clean up the `horizontalpodautoscaler.autoscaling/istio-pilot` that is created in the default namespace when we uninstall the Istio system. This is due to https://github.com/istio/istio/pull/7763 not being part of the Istio release used by Knative. Also, see https://github.com/knative/serving/issues/2018.

trisberg:
Fixes #918

### DIFF
--- a/pkg/core/system.go
+++ b/pkg/core/system.go
@@ -288,9 +288,7 @@ func deleteSingleResource(kc *kubectlClient, resourceType string, name string) e
 	fmt.Printf("Deleting %s/%s resource\n", resourceType, name)
 	deleteLog, err := kc.kubeCtl.Exec([]string{"delete", resourceType, name})
 	if err != nil {
-		if strings.Contains(deleteLog, "NotFound") {
-			fmt.Printf("Resource \"%s/%s\" was not found\n", resourceType, name)
-		} else {
+		if !strings.Contains(deleteLog, "NotFound") {
 			fmt.Printf("%s", deleteLog)
 		}
 	}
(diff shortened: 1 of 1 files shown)


pkg/core/system.go
```diff
@@ -282,6 +284,19 @@ func deleteNamespaces(kc *kubectlClient, namespaces []string) error {
 	return nil
 }
 
+func deleteSingleResource(kc *kubectlClient, resourceType string, name string) error {
```
scothis: should also accept a namespace rather than assuming the default

pkg/core/system.go
```diff
@@ -282,6 +284,19 @@ func deleteNamespaces(kc *kubectlClient, namespaces []string) error {
...
+	fmt.Printf("Deleting %s/%s resource\n", resourceType, name)
+	deleteLog, err := kc.kubeCtl.Exec([]string{"delete", resourceType, name})
+	if err != nil {
```
scothis: Since the fact this this resource exists is a bug that will hopefully be fixed, we should avoid logging an error in the case it doesn't exist. Instead we can return the err and push the logging responsibility to the caller.

pkg/core/system.go
trisberg: we don't know the namespace since we don't know what was the "default" namespace during install. Best we can do is hope that it is the same when run the uninstall.

pkg/core/system.go
trisberg: ok, agree on this

pkg/core/system.go
scothis: good point

PR #931, opened 2018-11-02
Inject manifests into image client

georgeharley:
* resolves error seen with riff-distro system download
* fixes #929

PR #934, opened 2018-11-02
Restore correct manifests versions

scothis:
The manifests were moved in #931 and at the same time regressed to older
buggy versions of Knative. At the same time, the moved manifests were
referenced directly in commands ignoring the values injected into the
commands. For the purposes of OSL tracking, this cannot happen. The
manifest defintitions need to remain inside the main package and not be
directly referenced by other portions of the codebase.

The riff-distro main now creates it's own default set of manifests and
injects those values into the riff-distro codebase. For simplicity, I
removed all other manifests except for 'latest', since keeping two sets
in sync will be difficult. In the long run, riff-distro users should
provide am explicit path to a manifest rather than use the compiled in
versions.

cc @glyn @georgeharley


glyn:
> The manifests were moved in #931 and at the same time regressed to older
buggy versions of Knative.

Not sure how that happened. Apologies.

> At the same time, the moved manifests were
referenced directly in commands ignoring the values injected into the
commands. For the purposes of OSL tracking, this cannot happen. The
manifest [definitions] need to remain inside the main package and not be
directly referenced by other portions of the codebase.

The moved manifests were actually only referred to in the wiring code, which is part of dependency injection. But point taken that it would be safer to restrict the declarations of manifests to main packages.

> The riff-distro main now creates it's own default set of manifests and
injects those values into the riff-distro codebase. For simplicity, I
removed all other manifests except for 'latest', since keeping two sets
in sync will be difficult.

The commands need to use consistent versions. In particular, `riff-distro create` needs to be able to refer to `latest` for regression testing (e.g. in nightly builds) and `stable` when building a distribution for release.

This is implemented by https://github.com/projectriff/riff/pull/936.

> In the long run, riff-distro users should
provide am explicit path to a manifest rather than use the compiled in
versions.

One of the aims of `riff-distro system download` was to download one of the compiled-in manifests to a file. If `riff-distro` needs to be passed an explicit path to a manifest, then some other utility would need to be provided to obtain that manifest file.

georgeharley:
@scothis @glyn The manifests regression during the move/cut-n-paste was squarely down to me but as well as being horrified I'm also baffled as to how it could have happened. Sincere apologies for letting it happen.

PR #936, opened 2018-11-05
Make `riff` and `riff-distro` use consistent manifest declarations

glyn:
It is important that the manifests are consistent so that distributions built with `riff-distro create` can refer to the same stable and latest values as `riff`.

This change ensures that manifests are only declared in main packages (to avoid the possibility of direct imports bypassing dependency injection), but shares the declarations using a symbolically linked file to avoid dual maintenance.

/cc @scothis @trisberg


Makefile
```diff
@@ -37,7 +37,7 @@ install: build
 	cp $(OUTPUT) $(GOBIN)
 
 $(OUTPUT): $(GO_SOURCES) vendor VERSION
-	go build -o $(OUTPUT) -ldflags "$(LDFLAGS_VERSION)" main.go
+	go build -o $(OUTPUT) -ldflags "$(LDFLAGS_VERSION)" *.go
```
ericbottard: Why is this necessary?

glyn: To ensure that `main.go` and `configuration.go` are compiled together. When it specifies `main.go`, the code does not compile.

scothis: we should explicitly specify both main.go and configuration.go in order to avoid future confusion

`make release` also needs to be updated.

ericbottard: I'm not sure I understand *why* this is necessary, I thought `go build` found all required files (just the same way it knows how to find referenced packages).

Anyway, we could also change from "build these `.go` files" to "build this package"

georgeharley: FWIW @ericbottard's suggestion of just building the package by not specifying any go source files at all in the `go build` invocations works fine for both the `make build` and `make release` targets.

glyn:
Merged as 8cc21f1d6a84d973cd62aa4ade84c0538f88eddb

Issue #937, opened 2018-11-05
Include packs/run in riff-distro image list


jldec:
The RUN_IMAGE which is used as the base image for all buildpack builds, is not currently being picked up by `riff-distro image list`,

#### unrelocated [buildtemplate](https://github.com/projectriff/riff/blob/6ac2a6f1cdda48f552aac7c11fa18ceafc5a8b06/third_party/cnb/riff-cnb-buildtemplate-0.1.0.yaml#L21-L23)
```yaml
  - name: RUN_IMAGE
    description: The run image buildpacks will use as the base for IMAGE.
    default: packs/run
```

#### relocated buildtemplate
```yaml
  - name: RUN_IMAGE
    description: The run image buildpacks will use as the base for IMAGE.
    default: packs/run
```

This image should be included in the distro and relocate process, just like the other buildpack-related images in the buildtemplate.

PR #938, opened 2018-11-05
Improve image scan

glyn:
Add a heuristic for image parameters with default values.

Fixes https://github.com/projectriff/riff/issues/937


ericbottard:
Merged as 025e95e6 with the use of `ResolveManifest()` in `ImageList()`

PR #942, opened 2018-11-06
Pin Knative eventing "latest"

scothis:
The legacy Knative eventing CRDs are being removed from the repo. We'll need to pin to this version until we rebase onto the new CRDs.

### DIFF
--- a/configuration.go
+++ b/configuration.go
@@ -50,8 +50,8 @@ var (
 			Knative: []string{
 				"https://storage.googleapis.com/knative-releases/build/latest/release.yaml",
 				"https://storage.googleapis.com/knative-releases/serving/latest/serving.yaml",
-				"https://storage.googleapis.com/knative-releases/eventing/latest/release.yaml",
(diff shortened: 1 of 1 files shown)


configuration.go
```diff
@@ -50,8 +50,8 @@ var (
 			Knative: []string{
 				"https://storage.googleapis.com/knative-releases/build/latest/release.yaml",
 				"https://storage.googleapis.com/knative-releases/serving/latest/serving.yaml",
-				"https://storage.googleapis.com/knative-releases/eventing/latest/release.yaml",
-				"https://storage.googleapis.com/knative-releases/eventing/latest/release-clusterbus-stub.yaml",
+				"https://storage.googleapis.com/knative-releases/eventing/previous/v20181106-a99376f/release.yaml",
+				"https://storage.googleapis.com/knative-releases/eventing/previous/v20181106-a99376f/release-clusterbus-stub.yaml",
```
trisberg: should we bump stable to the same version as well?

scothis: there's nothing in that build that we need.

markfisher:
LGTM

PR #948, opened 2018-11-08
Drop usage of the invoker *arg*

### LINKED ISSUE #945: Drop "invoker based" approach in `function create`
I'll make this a pre-requisite of #944

ericbottard:
The invoker *arg* is gone, and its previous meaning too.
It is now an optional `--invoker` flag, which is used to override the
detection done by the buildpack. It does not drive selection of an
invoker URL anymore.

The --invoker-url and <custom> support is also removed

Adds support for latest BP build for --local-path

Fixes #945


pkg/core/function.go
```diff
@@ -167,6 +150,25 @@ func (c *client) CreateFunction(options CreateFunctionOptions, log io.Writer) (*
 	return s, nil
 }
 
+func (c *client) writeRiffToml(options CreateFunctionOptions) error {
```
glyn: @ericbottard What's the point of writing a `riff.toml` file? I couldn't see that in the code that was deleted, so is this a new feature?

cmd/commands/function.go
```diff
@@ -44,66 +41,27 @@ func Function() *cobra.Command {
...
 			AtPosition(functionCreateFunctionNameIndex, ValidName()),
 		),
-		PreRunE: func(cmd *cobra.Command, args []string) error {
-			err := flagsValidator(cmd)
-			if err != nil {
-				return err
-			}
-
-			invoker := args[functionCreateInvokerIndex]
-			if invoker != "custom" && createFunctionOptions.InvokerURL != "" {
```
glyn: What should users who are using customer invokers do instead? Do we need some docs or at least a notice in the release notes saying the function has been dropped?

VERSION
```diff
@@ -1 +1 @@
-0.1.4-snapshot
\ No newline at end of file
+0.2.0-snapshot
```
glyn: This is a kindness. I think semver would allow 0.1.5 to be a breaking change since all bets are off for v0.*.

pkg/core/function.go
ericbottard: That's the means the riff buildpack [uses](https://github.com/projectriff/riff-buildpack/blob/ecff5d07810a7d6d15f10164c4e1ea3c5b73917e/metadata.go#L50) to get some of its info (propagated [here](https://github.com/projectriff/riff-buildpack-group/blob/9d744887a7a9175d478e6c309fbd8bacc1453101/riff-cnb-buildtemplate.yaml#L48) when going the BuildTemplate way)

cmd/commands/function.go
ericbottard: They'll eventually use custom *builders*. Supported for the local case only (via this PR) for now

VERSION
ericbottard: True. But we agreed to bump with @markfisher and @scothis

pkg/core/function.go
ericbottard: Or to phrase it otherwise, yes this is a "new feature":
> Adds support for latest BP build for --local-path

scothis: Once there is a standard way to pass env values to the build pack we should use that instead of riff.toml,  it for now we should consider it an implementation detail.

glyn: Sounds ok. Can this overwrite a riff.toml provided by the user?

cmd/commands/function.go
```diff
@@ -44,66 +41,27 @@ func Function() *cobra.Command {
 	}
 }
 
-func FunctionCreate(fcTool *core.Client, invokers, buildpacks map[string]string) *cobra.Command {
+func FunctionCreate(fcTool *core.Client, builder string) *cobra.Command {
```
scothis: Rename to `defaultBuilder` since the value can be overridden with a flag

pkg/core/function.go
ericbottard: Yes, but that should never be a problem (what they pass via the CLI should be the source of truth. As @scothis stated, this should really be seen as an implementation detail IMHO)

cmd/commands/function.go
```diff
@@ -122,21 +80,22 @@ func FunctionCreate(fcTool *core.Client, invokers, buildpacks map[string]string)
...
-	LabelArgs(command, "INVOKER", "FUNCTION_NAME")
+	LabelArgs(command, "FUNCTION_NAME")
 
 	command.Flags().StringVarP(&createFunctionOptions.Namespace, "namespace", "n", "", "the `namespace` of the service")
 	command.Flags().BoolVarP(&createFunctionOptions.DryRun, "dry-run", "", false, dryRunUsage)
 	command.Flags().StringVar(&createFunctionOptions.Image, "image", "", "the name of the image to build; must be a writable `repository/image[:tag]` with credentials configured")
-	command.Flags().StringVar(&createFunctionOptions.InvokerURL, "invoker-url", "", "the path to a custom invoker url. Required if invoker is custom.")
 	command.MarkFlagRequired("image")
+	command.Flags().StringVar(&createFunctionOptions.Invoker, "invoker", "", "invoker runtime to override `language` detected by buildpack")
+	command.Flags().StringVar(&createFunctionOptions.BuildpackImage, "builder", builder, "the `repository/image[:tag]` coordinates of a custom buildpack builder [local builds only]")
```
scothis: The builder is required either as a switch or as a default. In other words, if the default is blank, the switch is required

pkg/core/function.go
glyn: The only problem is overwriting the user's file. Wouldn't it be better to fail and require the user to delete it?

ericbottard: I really don't think so. I think that slows down the workflow. There should not be a "user's file" in the first place anyway

configuration.go
```diff
@@ -33,14 +33,8 @@ This file is defined in the root of this repository and symbolically linked from
...
 var (
-	buildpacks = map[string]string{
-		"java":    "projectriff/buildpack",
-		"command": "projectriff/buildpack",
-		"node":    "projectriff/buildpack",
-	}
-	invokers = map[string]string{
-		"jar": "https://github.com/projectriff/java-function-invoker/raw/v0.1.1/java-invoker.yaml",
-	}
+	builder = "projectriff/buildpack:0.0.1-snapshot-ci-64830c3bbc6503beafdae382ead115806fa100ca"
```
scothis: Can we either drop the tag or use a stable tag that is updated?

glyn:
Merged as c0a4f76ae2247e196925a06a943b9db12aac0a3c

PR #949, opened 2018-11-08
delete image commands and distro CLI

glyn:
* delete dead code
* delete configuration.go and move buildpacks, invokers, and manifests into main.go


ericbottard:
Merged as 92c4b24b

Issue #961, opened 2018-11-14
function update should use the same --no-pull strategy as function create


jldec:
As a developer i would like the same behavior from both `function update` and `function create`

E.g. when the image names for the builder and run-image use dev.local, this should trigger --no-pull for local-path updates.

jldec:
per #964 `function build` is now renamed to `function update`

PR #974, opened 2018-11-19
Add test for function create/update

glyn:
Also:
* Add check of buildpack and run images when building locally

Fixes https://github.com/projectriff/riff/issues/961


cmd/commands/wiring.go
```diff
@@ -201,3 +210,11 @@ func installKubeConfigSupport(command *cobra.Command, client *core.Client, kc *c
 		return nil
 	}
 }
+
+type buildFactory struct {
+	bf *pack.BuildFactory
+}
+
+func (bf *buildFactory) BuildConfigFromFlags(f *pack.BuildFlags) (core.Build, error) {
+	return bf.bf.BuildConfigFromFlags(f)
```
ericbottard: Why the double indirection here? Can't we use (and mock, I assume) `pack.BuildFactory` directly?

pkg/core/function.go
```diff
@@ -536,14 +538,11 @@ func (c *client) UpdateFunction(options UpdateFunctionOptions, log io.Writer) er
 		repoName := configuration.RevisionTemplate.Spec.Container.Image
 		publish := publishImage(repoName)
 
-		if buildImage == "" || runImage == "" {
```
ericbottard: Is this error message replaced by https://github.com/projectriff/riff/pull/974/files#diff-962221b752f1f6c34ce677a7163a2875R150 ? This does not convey the same information to the user

pkg/core/function_test.go
```diff
@@ -14,106 +14,185 @@
...
+			createFunctionOptions core.CreateFunctionOptions
+			createdService        *v1alpha1.Service
+		)
 
-	Context("when a hard error occurs", func() {
 		BeforeEach(func() {
-			hardErrors = 1
+			mockServiceInterface.On("Create", mock.Anything).Return(testService, nil).Run(func(args mock.Arguments) {
+				createdService = args.Get(0).(*v1alpha1.Service)
+			}) // input to Create() should be thing returned by function create
```
ericbottard: Not sure I understand this. Stores the arg in `createdService`, but returns `testService`
Also, may be more readable to swap the calls to Return and Run

pkg/core/function.go
glyn: Yes. The two new messages were intended to be more precise and more specific. Since this is common code, we have to cover both create and update. The old message didn't make a lot of sense in the case of create. The new messages make sense in both cases.

cmd/commands/wiring.go
glyn: I started with that approach but couldn't get it to work. The problem is the signature of the `BuildFactory` method `BuildConfigFromFlags(f *BuildFlags) (*BuildConfig, error)`. If we mock this directly, the mock has to return a `*BuildConfig`, but we need to mock the `Run` method of `*BuildConfig`, so we need to return an interface which can then be mocked. I'm open to other solutions...

pkg/core/function.go
ericbottard: ~~Well, my point is, I'm not sure it does (I'm talking about the update case). Previously, it hinted to the user that she was trying to rebuild locally ...~~

Hmm, scratch that. I now realize this was covering the case trying to build locally an "invoker" based function. That case is gone now anyways

pkg/core/function_test.go
glyn: > Not sure I understand this. Stores the arg in createdService, but returns testService
I'll improve the comment.

PR #978, opened 2018-11-20
Delete dev.local

glyn:
Fixes https://github.com/projectriff/riff/issues/977


glyn:
Note that I left the unit tests of function create/update in place even though these were introduced as part of the support for dev.local. Having partial unit tests is better than nothing.

glyn:
> I would prefer to deprecate the feature vs removing it. We could remove any CLI documentation for it but still not push dev.local or ko.local images to provide a smoother transition to using a registry

I'm uncomfortable with shipping an undocumented feature as someone is likely to end up depending on it. If this support is really necessary to provide a smooth transition, we should document it and state that it is deprecated in the documentation.

jldec:
I would agree with @trisberg unless we can provide a straighforward alternative for localhost builds without depending on a hosted registry. 

I'm less worried about breaking/changing workflows, but deprecating + documenting seems like a reasonable option.

see #980

pkg/core/function.go
```diff
@@ -147,23 +145,21 @@ func (c *client) CreateFunction(options CreateFunctionOptions, log io.Writer) (*
...
 	if buildImage == "" {
 		return fmt.Errorf("unable to build function locally: buildpack image not specified")
 	}
 	if runImage == "" {
 		return fmt.Errorf("unable to build function locally: run image not specified")
 	}
 
-	noPull := strings.HasPrefix(buildImage, "dev.local") && strings.HasPrefix(runImage, "dev.local")
-
 	b, err := c.buildFactory.BuildConfigFromFlags(&pack.BuildFlags{
```
ericbottard: If we don't require special values for those two flags, then we can fall back to the `Run()` method we used before and simplify the code (may mess the mocking contortions you went through recently @glyn , but should simplify it I think)

glyn: @ericbottard Yes, thanks. I pushed an extra commit with the simplifications.

ericbottard: Sorry, I meant `pack.Build()` not `Run()`, as in the reversal of what happened [here](https://github.com/projectriff/riff/commit/2d33e1d414ae9c8ca7490ccd997a0d8be7d5f77d#diff-962221b752f1f6c34ce677a7163a2875L109)  (not sure this can be directly mocked though)

glyn: Yes - please see the latest commit for the replacement interface.

glyn:
This PR seems to be in a bad state as Travis can’t access the relevant branch. This may be due to me accidentally pushing the branch to origin, rather than my own fork, earlier and then deleting the branch on origin. Maybe Travis or github or some combination uses a branch on origin but doesn’t re-establish it properly. I’m going to create a fresh PR with the same contents.

Issue #984, opened 2018-11-27
Add an explicit `--unauthenticated` third option to namespace init


ericbottard:
Currently, `riff namespace init` requires either `--dockerhub` or `--gcr` to be set, or that the user points to a secret she has already set up. There should be another option, albeit explicit, to work without authentication

jldec:
how about `--no-secret` for symmetry with `-s, --secret`

PR #985, opened 2018-11-27
--no-secret flag on namespace init

glyn:
Fixes https://github.com/projectriff/riff/issues/984


cmd/commands/namespace.go
```diff
@@ -70,6 +71,7 @@ func NamespaceInit(manifests map[string]*core.Manifest, kc *core.KubectlClient)
 
 	command.Flags().StringVarP(&options.Manifest, "manifest", "m", "stable", "manifest of YAML files to be applied; can be a named manifest (stable or latest) or a path or URL of a manifest file")
 
+	command.Flags().BoolVarP(&options.NoSecret, "no-secret", "n", false, "no secret required (unauthenticated)")
```
trisberg: could this read "no secret required for the image registry (unauthenticated)"

glyn: Yes. How about dropping "(unauthenticated)" and saying simply "no secret required for the registry".

cmd/commands/namespace.go
```diff
@@ -49,7 +49,8 @@ func NamespaceInit(manifests map[string]*core.Manifest, kc *core.KubectlClient)
 		),
 		PreRunE: FlagsValidatorAsCobraRunE(
 			FlagsValidationConjunction(
-				AtMostOneOf("gcr", "dockerhub"),
+				AtMostOneOf("gcr", "dockerhub", "no-secret"),
+				AtMostOneOf("secret", "no-secret"),
 				NotBlank("secret"),
```
markfisher: not related to this PR necessarily, but does this `NotBlank` constraint mean the default specified in the help for "secret" is incorrect?

ericbottard: the default is "push-credentials" so it is NotBlank

markfisher: good point :) so how would it ever be blank, if a literal `""` were provided?

cmd/commands/namespace.go
trisberg: perfect

ericbottard: I wouldn't use `n` as a shorthand, as that value is used elsewhere for `--namespace`. I think all shorthands should be used consistently accross commands if possible. Obviously namespace as a flag doesn't make sense here, but still

markfisher: probably doesn't need to have a shorthand

glyn: I removed the shorthand.

glyn:
> Looks like --no-secret doesn't create a riff-build service account. That is needed for the builds even without a secret.

Excellent catch @trisberg. Fixed.

cmd/commands/namespace.go
```diff
@@ -49,7 +49,8 @@ func NamespaceInit(manifests map[string]*core.Manifest, kc *core.KubectlClient)
 		),
 		PreRunE: FlagsValidatorAsCobraRunE(
 			FlagsValidationConjunction(
-				AtMostOneOf("gcr", "dockerhub"),
+				AtMostOneOf("gcr", "dockerhub", "no-secret"),
```
ericbottard: I don't think this is correct.
We *do* want to support `--secret my-secret` and say `--gcr` (like before). What this means is that *we* create the secret contents (because of `--gcr` but we store it in a secret name chosen by the user)

Issue #987, opened 2018-11-27
Add additional retries to system install


trisberg:
We should add additional retries to avoid failing install when CRD creation is slow

jldec:
How about also making the delay-between-retries grow each time

glyn:
We could use exponential backoff with a backoff factor less than 1, but I wouldn't recommend it. ;-)

PR #989, opened 2018-11-28
[DNM] Retry system install steps with exponential backoff

glyn:
In case 5 minutes turns out to be too short in the future, the maximum wait time is:

backoffInitialDuration * (1-backoffFactor**(backoffSteps-1))/(1-backoffFactor)

Use this to work out the number of steps: https://play.golang.org/p/N_WG-DloGv0

Fixes: https://github.com/projectriff/riff/issues/987

### DIFF
--- a/pkg/core/system.go
+++ b/pkg/core/system.go
@@ -26,13 +26,22 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


glyn:
Exponential backoff seems unsuitable as retries are increasingly likely to succeed.

PR #994, opened 2018-12-05
Add the ability to output subscriptions as a graphviz dot file.

ericbottard:
This is a small thing I hacked for demos and talks and could turn out quite useful.

Example with the canonical "hello 49" example:

```
riff subscription list -odot | dot -T png | open -f -a /Applications/Preview.app
```

which gives

chain

### DIFF
--- a/cmd/commands/subscription.go
+++ b/cmd/commands/subscription.go
@@ -1,6 +1,11 @@
 package commands
 
 import (
+	"fmt"
+	"io"
+	"strings"
+	"text/template"
+
 	"github.com/knative/eventing/pkg/apis/channels/v1alpha1"
 	"github.com/projectriff/riff/pkg/core"
 	"github.com/projectriff/riff/pkg/env"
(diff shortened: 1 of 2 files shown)


cmd/commands/subscription.go
```diff
@@ -102,18 +109,53 @@ func SubscriptionList(client *core.Client) *Command {
...
 	}
 
 	flags := command.Flags()
 	flags.StringVarP(&listOptions.Namespace, "namespace", "n", "", "the namespace of the subscriptions")
+	flags.StringVarP(&displayFormat, "output", "o", "", "the custom output format to use. Use 'dot' to output graphviz representation")
 
 	return command
 }
 
+func displayAsDot(out io.Writer, subscriptions []v1alpha1.Subscription) error {
```
fbiville: it should not output any file if the subscription list is empty, WDYT?

fbiville: I'll do the change this afternoon

ericbottard:
No, It should Always output à valid dot file IMO

Le lun. 10 déc. 2018 à 12:54, Florent Biville <notifications@github.com> a
écrit :

fbiville:
Fair enough. At least, the command highlighted above is guaranteed to work no matter what.

fbiville:
Merged in 1df43181a2e24e01ff1444f87510cdc436146a1a

PR #1008, opened 2018-12-11
Add removal of "knative-ingressgateway" to "system uninstall"

### LINKED ISSUE #1002: Add removal of "knative-ingressgateway" to "system uninstall"
"riff system uninstall" does not remove "knative-ingressgateway" unless the `--istio` flag is specified. Since "knative-ingressgateway" belongs to the serving install we should remove it even if Istio is not removed.

Add corresponding commands:
$ kubectl delete -n istio-system service knative-ingressgateway
$ kubectl delete -n istio-system horizontalpodautoscaler knative-ingressgateway
$ kubectl delete -n istio-system deployment knative-ingressgateway

trisberg:
Fixes #1002

### DIFF
--- a/pkg/core/system.go
+++ b/pkg/core/system.go
@@ -191,9 +191,9 @@ func (kc *kubectlClient) SystemUninstall(options SystemUninstallOptions) (bool, 
 		if err != nil {
 			return false, err
 		}
-		deleteNamespaceResource(kc, "istio-system", "service", "knative-ingressgateway")
-		deleteNamespaceResource(kc, "istio-system", "horizontalpodautoscaler", "knative-ingressgateway")
-		deleteNamespaceResource(kc, "istio-system", "deployment", "knative-ingressgateway")
+		deleteSingleResource(kc, "service", "knative-ingressgateway", "istio-system")
+		deleteSingleResource(kc, "horizontalpodautoscaler", "knative-ingressgateway", "istio-system")
+		deleteSingleResource(kc, "deployment", "knative-ingressgateway", "istio-system")
 		err = deleteNamespaces(kc, knativeNamespaces)
 		if err != nil {
 			return false, err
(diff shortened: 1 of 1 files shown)


pkg/core/system.go
```diff
@@ -191,6 +191,9 @@ func (kc *kubectlClient) SystemUninstall(options SystemUninstallOptions) (bool,
 		if err != nil {
 			return false, err
 		}
+		deleteNamespaceResource(kc, "istio-system", "service", "knative-ingressgateway")
```
glyn: Why are these resources not deleted when the `istio-system` namespace is deleted? Or is there an ordering requirement?

I'd recommend logging an issue to record the requirements for this PR.

pkg/core/system.go
```diff
@@ -291,6 +294,17 @@ func deleteNamespaces(kc *kubectlClient, namespaces []string) error {
 	return nil
 }
 
+func deleteNamespaceResource(kc *kubectlClient, namespace string, resourceType string, name string) error {
+	fmt.Printf("Deleting %s/%s resource in %s\n", resourceType, name, namespace)
+	deleteLog, err := kc.kubeCtl.Exec([]string{"delete", "-n", namespace, resourceType, name})
+	if err != nil {
+		if !strings.Contains(deleteLog, "NotFound") {
```
glyn: Wouldn't it be more robust to use `errors.IsNotFound(err)` (see [this example](https://github.com/projectriff/riff/blob/10227fb7a68677476415ee290710c9cea00a0b34/pkg/core/namespace.go#L89)) instead of a string search?

glyn: Now that conditional logic is increasing, is it time to write a unit test?

pkg/core/system.go
trisberg: forgot to add "Fixes", fixed now, the issue is #1002

glyn: Ah, lovely. Thanks. Makes sense now.

pkg/core/system.go
glyn: Probably worth making that explicit by factoring out a common function (e.g. `isKubectlNotFound(error) bool`) and calling it from both places.

trisberg: We don't have any tests for `core.system`, I created #1010

glyn: Fair enough. It's probably non-trivial.

glyn:
Merged in spite of fats-lite having timed out on Travis.

PR #1012, opened 2018-12-12
Update FATS ready checks for Knative eventing 0.2

scothis:
The stub bus no longer exists, there is now an in-memory-channel cluster channel provisioner


scothis:
Will also need to update FATS eventing ready checks before this will pass

scothis:
@trisberg feel free to merge after CI goes green

PR #1015, opened 2018-12-13
Change the way builds are annotated

### LINKED ISSUE #939: Adapt to changes in providing build spec for a kservice/configuration
This becomes important since our current strategy of updating an annotation on the revisionTemplate to trigger a build won't continue to work.

See https://github.com/knative/serving/issues/439
and https://github.com/knative/serving/pull/2402

trisberg:
- due to changes in Knative Serving we need to add a full Build as an Unstructured object

- we can then annotated the Build instead of the Revision to trigger a build

- this is necessary after the changes in https://github.com/knative/serving/pull/2402

Fixes #939

### DIFF
--- a/pkg/core/annotations.go
+++ b/pkg/core/annotations.go
@@ -20,9 +20,11 @@ import (
 	"strconv"
 
 	"github.com/knative/serving/pkg/apis/serving/v1alpha1"
+	"github.com/knative/serving/pkg/reconciler/v1alpha1/configuration/resources"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-func (c *client) bumpNonceAnnotation(s *v1alpha1.Service) {
+func (c *client) bumpNonceAnnotationForRevision(s *v1alpha1.Service) {
 	annotations := s.Spec.RunLatest.Configuration.RevisionTemplate.Annotations
 	if annotations == nil {
 		annotations = map[string]string{}
@@ -34,4 +36,33 @@ func (c *client) bumpNonceAnnotation(s *v1alpha1.Service) {
 	}
 	annotations[buildAnnotation] = strconv.Itoa(i + 1)
 	s.Spec.RunLatest.Configuration.RevisionTemplate.SetAnnotations(annotations)
+}
+
+func (c *client) bumpNonceAnnotationForBuild(s *v1alpha1.Service) {
+	configurationSpec := s.Spec.RunLatest.Configuration
+	build := getBuild(configurationSpec)
+	if build != nil {
+		annotations := build.GetAnnotations()
+		if annotations == nil {
+			annotations = map[string]string{}
+		}
+		nonce := annotations[buildAnnotation]
+		i, err := strconv.Atoi(nonce)
+		if err != nil {
+			i = 0
+		}
+		annotations[buildAnnotation] = strconv.Itoa(i + 1)
+		build.SetAnnotations(annotations)
+		s.Spec.RunLatest.Configuration.Build = &v1alpha1.RawExtension{
+			Object: build,
+		}
+	}
+}
+
+func getBuild(configSpec v1alpha1.ConfigurationSpec) *unstructured.Unstructured {
+	if configSpec.Build == nil {
+		return nil
+	}
+	u := resources.GetBuild(&configSpec)
+	return u
 }
--- a/pkg/core/function.go
+++ b/pkg/core/function.go
@@ -38,7 +38,7 @@ import (
 	"github.com/projectriff/riff/pkg/env"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 )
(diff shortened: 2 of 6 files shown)


ericbottard:
I can confirm this allows the triggering of a new build with "latest" manifest (while master doesn't)

trisberg:
The `getBuild` function should really be part of knative. I created a PR here: https://github.com/knative/serving/pull/2737

trisberg:
We should use the `GetBuild` method from knative/serving#2737 once we move to 0.3 for serving

ericbottard:
Should we move forward and merge (an updated version of) this now?

pkg/core/function.go
```diff
@@ -120,18 +115,27 @@ func (c *client) CreateFunction(buildpackBuilder Builder, options CreateFunction
...
+			},
+		}
 		s.Spec.RunLatest.Configuration.Build = &v1alpha1.RawExtension{
-			BuildSpec: &build.BuildSpec{
-				ServiceAccountName: "riff-build",
-				Source:             c.makeBuildSourceSpec(options),
-				Template: &build.TemplateInstantiationSpec{
-					Name:      "riff-cnb",
-					Arguments: c.makeBuildArguments(options),
+			Object: &unstructured.Unstructured{
```
scothis: it shouldn't be necessary to fallback to the unstructured option since we are using a Build resource.

pkg/core/annotations.go
```diff
@@ -35,3 +37,39 @@ func (c *client) bumpNonceAnnotation(s *v1alpha1.Service) {
 	annotations[buildAnnotation] = strconv.Itoa(i + 1)
 	s.Spec.RunLatest.Configuration.RevisionTemplate.SetAnnotations(annotations)
 }
+
+func (c *client) bumpNonceAnnotationForBuild(s *v1alpha1.Service) {
+	configurationSpec := s.Spec.RunLatest.Configuration
```
scothis: should we also handle Pinned and Release config modes?

pkg/core/annotations.go
```diff
@@ -35,3 +37,39 @@ func (c *client) bumpNonceAnnotation(s *v1alpha1.Service) {
...
+	build := getBuild(configurationSpec)
+	if build != nil {
+		annotations := build.GetAnnotations()
```
scothis: There's no need to use the get/set method, you can interact with the Annotations field directly.

pkg/core/annotations.go
```diff
@@ -35,3 +37,39 @@ func (c *client) bumpNonceAnnotation(s *v1alpha1.Service) {
...
+		if annotations == nil {
+			annotations = map[string]string{}
+		}
+		nonce := annotations[buildAnnotation]
+		i, err := strconv.Atoi(nonce)
```
scothis: rather than futz with conversation and errors, can we simply assign a unique value like `build.Generation` or `build.ResourceVersion` if you'd rather a string. Both values are read-only (set by the api server) and guaranteed unique between mutations.

pkg/core/annotations.go
```diff
@@ -35,3 +37,39 @@ func (c *client) bumpNonceAnnotation(s *v1alpha1.Service) {
...
+		annotations[buildAnnotation] = strconv.Itoa(i + 1)
+		build.SetAnnotations(annotations)
+		s.Spec.RunLatest.Configuration.Build = &v1alpha1.RawExtension{
+			Object: build,
+		}
+	}
+}
+
+func getBuild(configSpec v1alpha1.ConfigurationSpec) *unstructured.Unstructured {
+	// Trap panic from the Knative resources.GetBuild function call
```
scothis: why does it panic? That seems like something that should be fixed upstream

pkg/core/annotations.go
trisberg: Hmmm, just keeping the same code we already had, also not seeing how I can manipulate the annotations directly on an Unstructured object

pkg/core/annotations.go
trisberg: again, just keeping the same logic we already had, maybe we can create a separate PR if we want to change this

pkg/core/annotations.go
trisberg: there are a couple of panics in the upstream code, seems to be when the data doesn't match what is expected

trisberg: I reworked our tests to create a more complete Build object to avoid triggering a panic

trisberg:
rebased

pkg/core/annotations.go
trisberg: not in this PR, maybe separate issue

pkg/core/annotations.go
```diff
@@ -20,9 +20,11 @@ import (
 	"strconv"
 
 	"github.com/knative/serving/pkg/apis/serving/v1alpha1"
+	"github.com/knative/serving/pkg/reconciler/v1alpha1/configuration/resources"
```
scothis: We shouldn't use this package, it's an implementation detail for serving

trisberg: hmm, I agree that it is not ideal, but do you have a better idea how to access the annotations of the build definition embedded in the kservice?

scothis: Do the equivalent of https://github.com/knative/serving/blob/99ceee84e0c158ea1388af3420feff63b37f60ef/pkg/reconciler/v1alpha1/configuration/resources/build.go#L62-L96 yourself.

If this is something we want Knative to maintain, we should add it as a method in https://github.com/knative/serving/tree/master/pkg/apis/serving/v1alpha1

trisberg: That seems worse, at least the implementation in knative is more likely to be maintained even if it is not exposed

trisberg:
Rebased and squashed

trisberg:
Rebased and squashed, again

PR #1018, opened 2018-12-14
Stop update if local path is provided for cluster-built functions

### LINKED ISSUE #1017: riff service update --local-path with an on-cluster build is misleading
When a function has been created with a a cluster Build and is later rebuilt with `riff function update`, the `--local-path` flag is accepted and silently ignored, which is misleading

fbiville:
Fixes #1017

### DIFF
--- a/pkg/core/function_test.go
+++ b/pkg/core/function_test.go
@@ -141,7 +141,7 @@ var _ = Describe("Function", func() {
 				testService.Spec = v1alpha1.ServiceSpec{
 					RunLatest: &v1alpha1.RunLatestType{
 						Configuration: v1alpha1.ConfigurationSpec{
-							Build: nil, // null build === local build
+							Build: nil, // nil build === local build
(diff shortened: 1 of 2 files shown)


pkg/core/function_test.go
```diff
@@ -150,6 +137,19 @@ var _ = Describe("Function", func() {
 
 		Context("when building locally", func() {
 			BeforeEach(func() {
+				mockServiceInterface.On("Get", mock.Anything, mock.Anything).Return(testService, nil)
+				testService.Spec = v1alpha1.ServiceSpec{
+					RunLatest: &v1alpha1.RunLatestType{
+						Configuration: v1alpha1.ConfigurationSpec{
+							Build: nil, // null build === local build
```
ericbottard: `null`? What's `null`? ;)

fbiville: fixed :)

ericbottard:
Merged as 92fb5fad

PR #1021, opened 2018-12-17
Restore eventing tests

scothis:


scothis:
When cherry-picking to v0.2.x the fats commits will need to come from the fats v0.2.x branch and not master

PR #1024, opened 2018-12-18
Upgrade to Knative Serving 0.2.3

scothis:
See https://github.com/knative/serving/releases/tag/v0.2.3


scothis:
There are also a bunch of vendor changes for pack@v0.0.7, which was previously lost in a bad merge.

cc @trisberg

scothis:
Serving 0.2.3 is broken. We should not upgrade

scothis:
Reopening v0.2.3 was fixed and republished

PR #1026, opened 2018-12-18
Allow the windows test to fail without failing the build

scothis:


.travis.yml
```diff
@@ -10,14 +10,12 @@ install: true
...
   include:
   - stage: test
-    name: "Test on Linux"
-    os: linux
     install:
     - "go get -u github.com/golang/dep/cmd/dep"
     - "go get -u github.com/vektra/mockery/.../"
     script: make build test verify-docs
-  - stage: test
-    name: "Test on Windows (Git Bash)"
```
georgeharley: Sure, having a name is not necessary but it was an opportunity to remind maintainers that (currently) Windows builds are being carried out in a Windows GitBash shell.

scothis: I removed the name because the stage and os are already represented in the UI and redundant. When a name is specified the UI hides the language version and env vars which are generally more useful.

Is there something we're using to set Git Bash or is that the travis default way to run a script?

scothis: used a traditional comment to indicate Git Bash

georgeharley: Git Bash is currently the Travis default for Windows builds. Easy to miss it in [the current documentation](https://docs.travis-ci.com/user/reference/windows/#git-bash) (I initially did).

PR #1028, opened 2018-12-19
DNM: Allow configuration of image pull secret during namespace init

fbiville:
This will allow users to run functions from private registries.
See https://github.com/pivotal-cf/pfs/issues/159.

### DIFF
--- a/pkg/core/namespace_test.go
+++ b/pkg/core/namespace_test.go
@@ -17,6 +17,7 @@
 package core
 
 import (
+	"fmt"
 	"os"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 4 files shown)


fbiville:
The various secret options start to become confusing in my opinion.
Shall we rename some of them and maybe alias them to their former name for backwards compatibility?

pkg/core/namespace_test.go
```diff
@@ -69,6 +70,55 @@ var _ = Describe("The NamespaceInit function", func() {
 		Expect(err).To(MatchError(ContainSubstring("wrong: "))) // error message is quite different on Windows and macOS
 	})
 
+	It("should fail to add an image pull secret if the default service account is not found", func() {
+		namespaceName := "foo"
+		pullSecretName := "pullitzer-prize"
```
glyn: Ha ha. Very good!

ericbottard:
Some of my €.02:
1. Push to registry is handled by knative (according to [this](https://github.com/knative/docs/blob/master/build/auth.md)). Pull from registry is handled by k8s PullSecrets, using the special docker-registry [type](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#create-a-secret-in-the-cluster-that-holds-your-authorization-token). So it is technically possible for an SA to be configured to do one, but not both
2. per the above, somewhat unrelated
3. secret -> user -> role
4. Depending on the registry used, there are many out-of-the-box magical tricks it seems: https://kubernetes.io/docs/concepts/containers/images/#using-a-private-registry

scothis:
Using an imagePullSecret feels like the wrong approach since we'd need to add the secret to every service account, in every namespace, past, present and future. It would be better to configure this within the cluster itself to all pods can pull from the registry.

fbiville:
This needs to be done for private registries as many times as `namespace init` is needed and affects only the default namespace's service account as @trisberg says.

jldec:
@scothis you said:
> It would be better to configure this within the cluster itself to all pods can pull from the registry.

Could you elaborate what you meant by that.

scothis:
@jldec the problem with using a pull secret is that it needs to be attached to each service account. In the best case, this secret will need to be attached to the default service account in each namespace that will host functions. It gets worse if the Knative images also need to be pulled from this private repo. For a default Knative install, I count 20 service accounts. That's 20 service accounts that will need this secret attached (and the secret will need to be recreated in each of 4 namespaces). That's a lot of secrets to manage.

Instead, I'd like to see [each node authenticate](https://kubernetes.io/docs/concepts/containers/images/#configuring-nodes-to-authenticate-to-a-private-registry) to pull images from the private repo. It's more complex for us to implement, but will provide a better user experience. A cross cutting solution to a cross cutting problem.

It might also be fair for us to push this onto the PKS team to support, as every other major k8s distribution has a private container registry built-in with managed authentication.

jldec:
per sync discussion, we will hold off on offering a generic solution a-la projectriff/riff#1028 that covers all registries / environments for now.

PR #1029, opened 2018-12-19
Switch from dep to go modules

### LINKED ISSUE #752: Investigate switching to go modules

ericbottard:
Fixes #752


.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: go
 go:
-- '1.11'
```
scothis: It's also worthwhile adding a cache to avoid redownloading modules all the time

```yaml
cache:
  directories:
  - $GOPATH/pkg/mod
```

.travis.yml
```diff
@@ -1,6 +1,6 @@
...
+- '1.11.4'
```
scothis: Not sure if multiline suggestions work...

```suggestion
- '1.11.4'
cache:
  directories:
  - $GOPATH/pkg/mod
```

PR #1034, opened 2018-12-20
Remove vendored dependency

fbiville:


fbiville:
I can confirm that PFS, installed against a PKS install with internal UAA enabled, still works with this change.

Issue #1036, opened 2018-12-21
riff system install fails to install using manifest from local directory on Windows


glyn:
To reproduce this, download a riff manifest and the corresponding kubernetes configuration files so that `manifest.yaml` looks something like this:
```
manifestVersion: "0.1"
istio:
- istio-riff-knative-serving-v0-2-3.yaml
knative:
- release.yaml
- serving.yaml
- eventing.yaml
- in-memory-channel.yaml
namespace:
- riff-cnb-buildtemplate-0.1.0.yaml
```
Notice that the kubernetes configuration files are referenced using relative file paths rather than URLs. These files are in the same directory as the manifest.

With this in place, issue the following from a Windows command prompt:
```
riff system install -m file:///C:\path\to\manifest\manifest.yaml
```

The failure is as follows:
```
Istio is already installed, do you want to install the Knative components for riff? [y/N]: y
Waiting for the Istio components to start . all components are 'Running'

Installing Knative components
Error applying resources, trying again

Error: open file:\C:\path\to\manifest\release.yaml: The filename, directory name, or volume label syntax is incorrect.
```

glyn:
Re-opening until @trisberg has had a chance to break the code again. :-)

trisberg:
Tried with `riff system install -m file:///C:\path\to\manifest\manifest.yaml` and that still doesn't work which could be expected.

This works fine: `riff system install -m file:///C:/path/to/manifest/manifest.yaml`, so I think we can close this issue.

trisberg:
This seems to be failing again, can't find a file URL combination that works with the latest riff CLI

glyn:
I've unassigned this so others can take a crack at it. My Windows VM has probably rotted by now.

Also, I'm not comfortable with the absence of automatically-triggered and mandatory Windows CI as we can easily break Windows support without knowing (e.g. if those running Windows tests go on vacation).

trisberg:
I agree that we do need to set up automated CI build somewhere. Question is where should we set this up? We could try using build.spring.io which had Windows support.

PR #1037, opened 2018-12-21
Allow file URL to be specified as manifest for riff system install

glyn:
Fixes https://github.com/projectriff/riff/issues/1036


glyn:
This should also allow a file URL to be specified as manifest for `riff namespace init`.

trisberg:
Sorry, I still can't get namespace init to work with a file URL.

It works without `file:///` prefix both as `C:\Users\trisberg\Downloads\pfs-files\manifest.yaml` and as `C:/Users/trisberg/Downloads/pfs-files/manifest.yaml` 

```
PS C:\Users\trisberg> riff namespace init test --no-secret --manifest file:///C:\Users\trisberg\Downloads\pfs-files\manifest.yaml
Initializing namespace "test"

Applying riff-cnb-buildtemplate-0.1.0-66c3cc7a6054c438d8505b30cf4772a9.yaml in namespace "test"
error: the path "file:\\C:\\Users\\trisberg\\Downloads\\pfs-files\\riff-cnb-buildtemplate-0.1.0-66c3cc7a6054c438d8505b30cf4772a9.yaml" cannot be accessed: CreateFile file:\C:\Users\trisberg\Downloads\pfs-files\riff-cnb-buildtemplate-0.1.0-66c3cc7a6054c438d8505b30cf4772a9.yaml: The filename, directory name, or volume label syntax is incorrect.


Error: exit status 1
```

pkg/core/manifest.go
```diff
@@ -84,6 +91,29 @@ func (m *Manifest) VisitResources(f func(resource string) error) error {
 	return nil
 }
 
+// ResourceAbsolutePath takes a path to a resource and returns an equivalent absolute path.
+// If the input path is a http(s) URL or is an absolute file path, it is returned without modification.
+// If the input path is a file URL, the corresponding absolute file path is returned.
```
ericbottard: This comment could be a bit misleading: a `file:` URL is absolute by definition, there is no way to express relative paths with a `file:` protocol (because what comes after is the host, even if implicit and after that an *absolute* path, by definition)

pkg/fileutils/read.go
```diff
@@ -95,6 +55,17 @@ func readAbsFile(file string) ([]byte, error) {
 	return ioutil.ReadFile(file)
 }
 
+func fileURLPath(u *url.URL) string {
+	file := u.Path
+
+	// Sanitise path on Windows. See https://github.com/golang/go/issues/6027
+	if runtime.GOOS == "windows" && file[0] == '/' {
```
ericbottard: Can't `file` be the empty string?

pkg/fileutils/abs.go
```diff
@@ -0,0 +1,81 @@
...
+	if filepath.IsAbs(file) {
+		return true, file, nil
+	}
+
+	u, err := url.Parse(file)
+	if err != nil {
+		return false, "", err
+	}
+
+	if u.IsAbs() && (u.Scheme == "http" || u.Scheme == "https") {
```
ericbottard: While not wrong, I think the `IsAbs()` check is redundant because a non-empty scheme implies an absolute URL

pkg/fileutils/abs.go
```diff
@@ -0,0 +1,81 @@
...
+
+	b, err := url.Parse(base)
+	if err != nil {
+		return "", err
+	}
+	if b.IsAbs() && (b.Scheme == "http" || b.Scheme == "https" || b.Scheme == "file") {
+		b.Path = path.Join(b.Path, filepath.ToSlash(file))
+		return b.String(), nil
+	}
+
```
ericbottard: Should check from here that there is no scheme / `!IsAbs()`  (ie could be say, a `ftp://` url)

pkg/fileutils/read.go
glyn: I don't _think_ file can be empty because, currently, the callers create the input URL by parsing a string and I don't think that can result in an empty path. However, I agree that the test isn't robust since a URL path could in general be empty, so I'll replace the condition with:
```
runtime.GOOS == "windows" && len(file) > 0 && file[0] == '/'
```

pkg/fileutils/abs.go
glyn: True. I'll remove the redundant check.

pkg/fileutils/abs.go
glyn: Checked, here and elsewhere.

pkg/core/manifest.go
glyn: I'm not sure what is misleading. Does the comment somehow imply that the URL path could be relative?

ericbottard:
Merged as squashed 41409f51

PR #1041, opened 2018-12-31
Let the CLI be fully responsible of riff.toml file

### LINKED ISSUE #1040: riff should create and delete riff.toml

ericbottard:
The CLI now deletes the file when it is done building.
Fails with an explanation if the file is already there.

Fixes #1040

### DIFF
--- a/pkg/core/function.go
+++ b/pkg/core/function.go
@@ -29,7 +29,7 @@ import (
 
 	"github.com/BurntSushi/toml"
 
-	"github.com/boz/go-logutil"
+	logutil "github.com/boz/go-logutil"
 	"github.com/boz/kail"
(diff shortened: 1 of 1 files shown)


pkg/core/function.go
```diff
@@ -29,7 +29,7 @@ import (
 
 	"github.com/BurntSushi/toml"
 
-	"github.com/boz/go-logutil"
+	logutil "github.com/boz/go-logutil"
```
trisberg: why do we need this alias?

ericbottard: Without it Goland automatically removes the import under me...

fbiville:
Merged in 70553fcd72bf62a61fdd588cacd811bb4ef76b8d, with extra unit tests.

PR #1046, opened 2019-01-03
1036 install relative win

glyn:
This is a new PR for https://github.com/projectriff/riff/pull/1037 after Travis builds were moved from travis.org to travis.com.


glyn:
Still referring to travis.org failed build. Closing.

scothis:
@glyn The build status is saved on the commit, so the same commits in a new PR will have the same status. If you need to retrigger a build that was done on .org, the best bet is to push a new commit.

glyn:
@scothis Yeah, I spotted that and push a couple of extra doc comments. Thanks!

PR #1047, opened 2019-01-03
[DNM] Make Windows test failures fail PRs

glyn:


glyn:
Flagged as "do not merge" since the branch Windows build failure (!) caused the branch build check to fail. Maybe setting an env var after a job has started doesn't count?

PR #1057, opened 2019-01-10
Delete several resources (in same namespace)

### LINKED ISSUE #1052: riff {channel,service,subscription} delete should accept several names

fbiville:
Fixes #1052

### DIFF
--- a/cmd/commands/errors.go
+++ b/cmd/commands/errors.go
@@ -1,53 +1,53 @@
 package commands
 
 import (
-	"fmt"
 	"github.com/pkg/errors"
 	"strings"
 )
 
-// Joins and formats results.
+// Joins and formats results, with the provided formatter.
+// The result is skipped if the formatter returns an empty string.
 // Returns nil if all errors are nil.
-// If there is only one non-nil error, the  error is returned as-is.
-func MergeResults(input []error) error {
-	stats := computeStats(input)
+func MergeResults(results []CorrelatedResult, formatter func(CorrelatedResult) string) error {
+	stats := computeStats(results)
 	length := stats.nilFreeSliceLength
 	if length == 0 {
 		return nil
 	}
 	if length == 1 {
-		return input[stats.firstNonNilIndex]
+		return errors.New(format(formatter, results[stats.firstNonNilIndex]))
 	}
-	return errors.New(formatErrorMessages(input))
+	return errors.New(formatErrorMessages(results, formatter))
 }
 
-func formatErrorMessages(input []error) string {
+func formatErrorMessages(input []CorrelatedResult, formatter func(CorrelatedResult) string) string {
 	mergedMessage := strings.Builder{}
-	for i, err := range input {
-		if err == nil {
-			mergedMessage.WriteString(fmt.Sprintf("# Call %d - no error\n", i+1))
-			continue
+	for _, err := range input {
+		formattedMessage := format(formatter, err)
+		if formattedMessage != "" {
+			mergedMessage.WriteString(formattedMessage)
+			mergedMessage.WriteString("\n")
 		}
-		mergedMessage.WriteString(fmt.Sprintf("# Call %d - error\n", i+1))
-		leftPaddedMessage := strings.Replace(err.Error(), "\n", "\n\t", -1)
-		mergedMessage.WriteString("\t" + leftPaddedMessage + "\n")
 	}
-	s := mergedMessage.String()
-	return s
+	return strings.TrimSuffix(mergedMessage.String(), "\n")
 }
 
-func computeStats(errors []error) stats {
+func format(formatter func(CorrelatedResult) string, err CorrelatedResult) string {
+	return formatter(err)
+}
+
+func computeStats(results []CorrelatedResult) *stats {
 	firstNonNilIndex := -1
 	length := 0
-	for i, err := range errors {
-		if err != nil {
+	for i, result := range results {
+		if result.Error != nil {
 			length++
 			if firstNonNilIndex == -1 {
 				firstNonNilIndex = i
 			}
 		}
 	}
-	return stats{
+	return &stats{
 		nilFreeSliceLength: length,
 		firstNonNilIndex:   firstNonNilIndex,
 	}
--- a/cmd/commands/errors.go
+++ b/cmd/commands/errors.go
@@ -1,0 +1,59 @@
+package commands
+
+import (
+	"fmt"
+	"github.com/pkg/errors"
+	"strings"
+)
+
+// Joins and formats results.
+// Returns nil if all errors are nil.
+// If there is only one non-nil error, the  error is returned as-is.
+func MergeResults(input []error) error {
+	stats := computeStats(input)
+	length := stats.nilFreeSliceLength
+	if length == 0 {
+		return nil
+	}
+	if length == 1 {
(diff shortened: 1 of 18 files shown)


pkg/core/client.go
```diff
@@ -38,7 +38,7 @@ type Client interface {
 
 	ListChannels(options ListChannelOptions) (*eventing.ChannelList, error)
 	CreateChannel(options CreateChannelOptions) (*eventing.Channel, error)
-	DeleteChannel(options DeleteChannelOptions) error
+	DeleteChannel(options DeleteChannelOptions) []error
```
scothis: consider:

- pluralize channel
- merge errors before returning

```
DeleteChannels(options DeleteChannelOptions) error
```


We should also seriously consider leaving this interface unchanged and instead invoke it multiple times for a batch delete. That would give the caller more control over concurrency and retry.

cmd/commands/errors.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	"errors"
+	"strings"
+)
+
+func MergeErrors(input []error) error {
+	mergedMessage := strings.Builder{}
+	for _, err := range input {
+		if err != nil {
+			mergedMessage.WriteString(err.Error() + "\n")
```
scothis: not sure how I feel about new lines in the error message text when it's not formatted in a meaningful way.

pkg/core/client.go
fbiville: 👍

cmd/commands/errors.go
fbiville: Shall I introduce a clearer demarcation or do you have something else in mind?

pkg/core/client.go
fbiville: Thinking about it, I guess the last option you mention is probably the best.

cmd/commands/errors.go
fbiville: Or shall the client be responsible for providing a delimiter? WDYT?

cmd/commands/service.go
```diff
@@ -304,22 +304,25 @@ Additional curl arguments and flags may be specified after a double dash (--).`,
 	return command
 }
 
-func ServiceDelete(fcClient *core.Client) *cobra.Command {
+func ServiceDelete(riffClient *core.Client) *cobra.Command {
 
-	deleteServiceOptions := core.DeleteServiceOptions{}
+	options := core.DeleteServiceOptions{}
```
fbiville: feedback from @ericbottard: introduce a CLI option struct that accept a slice of names

cmd/commands/service.go
```diff
@@ -304,22 +304,25 @@ Additional curl arguments and flags may be specified after a double dash (--).`,
...
+			cobra.MinimumNArgs(serviceDeleteMinNumberOfArgs),
+			StartingAtPosition(serviceDeleteNameStartIndex, ValidName()),
 		),
 		RunE: func(cmd *cobra.Command, args []string) error {
-			fnName := args[serviceDeleteServiceNameIndex]
-			deleteServiceOptions.Name = fnName
-			err := (*fcClient).DeleteService(deleteServiceOptions)
+			names := args[serviceDeleteNameStartIndex:]
+			err := MergeErrors(DeleteResourcesInParallel(names, func(name string) error {
+				options.Name = name
```
fbiville: feedback from @ericbottard: unsafe, instantiate instead a service options struct from the CLI struct there

cmd/commands/resources.go
```diff
@@ -0,0 +1,19 @@
+package commands
+
+type done struct{}
+
+func DeleteResourcesInParallel(names []string, deleteResource func(string) error) []error {
```
fbiville: feedback from @ericbottard: rename

cmd/commands/resources.go
```diff
@@ -0,0 +1,19 @@
...
+	count := len(names)
+	deleteErrors := make([]error, count)
+	semaphore := make(chan done, count)
```
fbiville: feedback fron @ericbottard: use https://golang.org/pkg/sync/#WaitGroup instead

cmd/commands/parallel.go
```diff
@@ -0,0 +1,18 @@
+package commands
+
+import "sync"
+
+func ApplyInParallel(names []string, deleteResource func(string) error) []error {
+	count := len(names)
+	deleteErrors := make([]error, count)
+	waitGroup := sync.WaitGroup{}
```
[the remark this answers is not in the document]
fbiville: I'd rather stick with waitgroup, as errgroup stops at the first error.

cmd/commands/parallel.go
```diff
@@ -0,0 +1,18 @@
+package commands
+
+import "sync"
+
+func ApplyInParallel(names []string, deleteResource func(string) error) []error {
```
scothis: Old param names are leaking into a generic method.

cmd/commands/errors.go
```diff
@@ -0,0 +1,59 @@
...
+	if length == 0 {
+		return nil
+	}
+	if length == 1 {
+		return input[stats.firstNonNilIndex]
+	}
+	return errors.New(formatErrorMessages(input))
+}
+
+func formatErrorMessages(input []error) string {
```
scothis: I'm not crazy about the generic error format, we may want to allow a caller to provide a template for the error message.

For example, when deleting a channel, I'd like to see an error like:

```
fmt.Sprintf("Unable to delete channel %s: %v", context, err)
```

fbiville: what would `context` be in this example?

fbiville: context == channel name, alright 😅

fbiville: Done

cmd/commands/parallel.go
```diff
@@ -2,17 +2,25 @@ package commands
 
 import "sync"
 
-func ApplyInParallel(names []string, action func(string) error) []error {
+type CorrelatedResult struct {
+	Input string
```
scothis: In the future we may want this be type `interface{}` to allow for more than just strings. It's fine for today though.

cmd/commands/errors.go
```diff
@@ -1,53 +1,53 @@
...
 )
 
-// Joins and formats results.
+// Joins and formats results, with the provided formatter.
+// The result is skipped if the formatter returns an empty string.
 // Returns nil if all errors are nil.
-// If there is only one non-nil error, the  error is returned as-is.
-func MergeResults(input []error) error {
-	stats := computeStats(input)
+func MergeResults(results []CorrelatedResult, formatter func(CorrelatedResult) string) error {
```
scothis: It might be nice to pass the input and error directly to the formatter function rather than forcing users to understand what the CorrelatedResult type is.

cmd/commands/parallel.go
fbiville: we may even have generics at that point, who knows? :D

cmd/commands/errors.go
fbiville: Not sure about that change, though. Would `ApplyInParallel` still return `CorrelatedResult`s in your opinion?

fbiville: Having it exposed on one side and not the other would feel a bit strange to me.

fbiville:
Now merged.

PR #1059, opened 2019-01-11
Create a build cache per Function

### LINKED ISSUE #993: create a PersistentVolumeClaim for a Function build
This will replace the PVC that has been thus far defined in `riff-cnb-buildtemplate.yaml` which will be parameterized to support this configuration once https://github.com/projectriff/riff-buildpack-group/issues/12 is resolved.

We also should add a controlling Owner Reference on the volume claim for the Knative Service that's created. That way the volume claim will be deleted when the Knative Service is deleted.

In the future, we may expose this as a configurable property for the user, including the ability to suppress it (use "emptyDir"), change its size/storage-class, etc.

scothis:
A PersistentVolumeClaim is restricted from being bound to multiple pods
with read/write access concurrently. This effectively limits all builds
to be sequential.

Instead we can create a volume claim per service to enable concurrent
builds between services. The volume is marked as being controlled by the
service so that it is automatically deleted when the service is deleted.

Fixes #993
Requires projectriff/riff-buildpack-group#35


scothis:
Depends on:
- https://github.com/knative/build/pull/528
- updating the build template to parameterize the volume claim

pkg/core/function.go
```diff
@@ -119,25 +122,67 @@ func (c *client) CreateFunction(buildpackBuilder Builder, options CreateFunction
 			}
 		}
 	} else {
+		// create build cache for service
+		buildCache = &corev1.PersistentVolumeClaim{
```
ericbottard: The buildCache should now be part of the result in case DryRun == true IMO

trisberg:
testing this change  together with #1015 I don't see the maven repo being cached and reused for a function rebuild

trisberg:
never mind, forgot to apply the latest riff-cnb clusterbuildtemplate, with that it works as expected

PR #1064, opened 2019-01-23
Make --cluster-provisioner optional for create channel

scothis:
The cluster will automatically assign a new channel to the default
cluster channel provisioner. This means that we don't need to force the
user to specify the provisioner when creating a channel.


fbiville:
Merged in https://github.com/projectriff/riff/commit/15225f1a185ee5bb83b0de0c1a16f7d84b78b2b2.

PR #1067, opened 2019-01-23
riff function update works (again) with local-path

### LINKED ISSUE #1055: function update with --local-path fails detection
```
$ riff function update pipeline --local-path .
2019/01/09 16:41:08 Pulling builder image 'projectriff/builder' (use --no-pull flag to skip this step)
2019/01/09 16:41:08 Using user provided run image 'packs/run'
*** DETECTING:
2019/01/09 21:41:13 Group: Cloud Foundry OpenJDK Buildpack: pass | Node.js Buildpack: pass | Cloud Foundry Build System Buildpack: pass | NPM Buildpack: fail | riff Buildpack: fail
2019/01/09 21:41:13 Error: failed to detect

Error: run detect container: failed with status code: 6
```

ericbottard:
The CLI now saves info required for riff.toml in annotations in
the ksvc, to be later used by `riff function update`.

Some refactoring of the local build code to make it more atomic.

Fixes #1055

### DIFF
--- a/pkg/core/function.go
+++ b/pkg/core/function.go
@@ -43,28 +43,31 @@ import (
 )
 
 const (
-	functionLabel   = "riff.projectriff.io/function"
-	buildAnnotation = "riff.projectriff.io/nonce"
-	// annotation names with slashes are rejected :-/
+	functionLabel                 = "riff.projectriff.io/function"
+	buildAnnotation               = "riff.projectriff.io/nonce"
 	buildpackBuildImageAnnotation = "riff.projectriff.io-buildpack-buildImage"
 	buildpackRunImageAnnotation   = "riff.projectriff.io-buildpack-runImage"
+	functionArtifactAnnotation    = "riff.projectriff.io/artifact"
+	functionOverrideAnnotation    = "riff.projectriff.io/override"
+	functionHandlerAnnotation     = "riff.projectriff.io/handler"
 	pollServiceTimeout            = 10 * time.Minute
 	pollServicePollingInterval    = time.Second
 )
 
-type CreateFunctionOptions struct {
-	CreateOrUpdateServiceOptions
-
-	LocalPath   string
-	GitRepo     string
-	GitRevision string
-
+type BuildOptions struct {
 	Invoker        string
+	Handler        string
+	Artifact       string
+	LocalPath      string
 	BuildpackImage string
 	RunImage       string
+}
+type CreateFunctionOptions struct {
+	CreateOrUpdateServiceOptions
+	BuildOptions
 
-	Handler  string
-	Artifact string
+	GitRepo     string
+	GitRevision string
 }
 
 func (c *client) CreateFunction(buildpackBuilder Builder, options CreateFunctionOptions, log io.Writer) (*v1alpha1.Service, error) {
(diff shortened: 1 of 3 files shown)


pkg/core/function.go
```diff
@@ -43,28 +43,31 @@ import (
...
 
 const (
-	functionLabel   = "riff.projectriff.io/function"
-	buildAnnotation = "riff.projectriff.io/nonce"
-	// annotation names with slashes are rejected :-/
+	functionLabel                 = "riff.projectriff.io/function"
+	buildAnnotation               = "riff.projectriff.io/nonce"
 	buildpackBuildImageAnnotation = "riff.projectriff.io-buildpack-buildImage"
 	buildpackRunImageAnnotation   = "riff.projectriff.io-buildpack-runImage"
+	functionArtifactAnnotation    = "riff.projectriff.io/artifact"
```
glyn: The format of this annotation is inconsistent with the format of `buildpackRunImageAnnotation`. The rationale for the latter format is probably due to the belief in the deleted comment that "annotation names with slashes are rejected", but then how can `buildAnnotation` contain a slash? Since we are still in Alpha, I think it would make sense to switch the other annotation names to use slashes so the whole lot are consistent. Happy to defer that to a separate issue if you want to avoid scope creep in this PR.

pkg/core/function.go
```diff
@@ -43,28 +43,31 @@ import (
...
+	functionOverrideAnnotation    = "riff.projectriff.io/override"
```
glyn: Should this be modified for consistency with the other build option annotations? E.g.
```
functionInvokerAnnotation = "riff.projectriff.io/invoker"
```

pkg/core/function.go
```diff
@@ -43,28 +43,31 @@ import (
 )
 
 const (
-	functionLabel   = "riff.projectriff.io/function"
-	buildAnnotation = "riff.projectriff.io/nonce"
-	// annotation names with slashes are rejected :-/
+	functionLabel                 = "riff.projectriff.io/function"
+	buildAnnotation               = "riff.projectriff.io/nonce"
```
glyn: I'm not suggesting we change this here, but I'd like to point out that this "revision count" isn't a (cryptographic) nonce. Also, "nonce" has an [unfortunate connotation](https://www.urbandictionary.com/define.php?term=nonce) in the UK.

pkg/core/function.go
```diff
@@ -581,12 +553,46 @@ func (c *client) UpdateFunction(buildpackBuilder Builder, options UpdateFunction
...
+	return builder.Build(options.LocalPath, options.BuildpackImage, options.RunImage, image)
+}
+
+func writeRiffToml(options BuildOptions) error {
+	t := struct {
+		Override string `toml:"override"`
+		Handler  string `toml:"handler"`
+		Artifact string `toml:"artifact"`
+	}{
+		Override: options.Invoker,
```
glyn: Again, it would make sense to rename `Override` to `Invoker`.

pkg/core/function.go
ericbottard: Yes, came to the same conclusion. Indeed slashes are accepted for me.
Also think it's fine to change the old ones, in another PR

pkg/core/function.go
ericbottard: There is some discrepancy between riff.toml and the CLI terms. The name "invoker" became wrong imo, override is better. So we need to decide at which point we switch from one to the other.

pkg/core/function.go
ericbottard: ouch

pkg/core/function.go
glyn: I've raised issue https://github.com/projectriff/riff/issues/1076 for now as any PR would doubtless clash with this one and there's no point _making_ work.

pkg/core/function.go
glyn: Logged https://github.com/projectriff/riff/issues/1077.

pkg/core/function.go
ericbottard: This is part of the riff.toml (hidden) contract. Could tackle in a different PR, but would require some coordination

scothis:
Looks good to me. I was about to factor out BuildOptions independently.

We can follow up with the naming questions Glyn raised.

PR #1068, opened 2019-01-23
Increase retries of system install steps

glyn:
Simply repeats every 5 seconds and gives up after 5 minutes. Exponential
backoff, for example, is not suitable as the chance of success does not reduce
over time.

Fixes: https://github.com/projectriff/riff/issues/987

### DIFF
--- a/pkg/core/system.go
+++ b/pkg/core/system.go
@@ -32,7 +32,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const istioNamespace = "istio-system"
+const (
+	istioNamespace = "istio-system"
+
+	retryDuration = 5 * time.Second
+	retrySteps    = 5
+)
 
 type SystemInstallOptions struct {
 	Manifest string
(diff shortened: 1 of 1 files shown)


pkg/core/system.go
```diff
@@ -21,18 +21,22 @@ import (
...
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const istioNamespace = "istio-system"
+const (
+	istioNamespace = "istio-system"
+
+	retryDuration = 5 * time.Second
+	retrySteps    = 60
```
trisberg: isn't this a bit excessive, I've never had a case where I had to try more than 3-4 times. So if there is a real issue other than the time it takes to create the cluster resources we would just be spinning our wheels for 5 minutes.

scothis:
Is this actually an issue that needs to be solved? I haven't seen a case where our current retry logic fails, but would eventually succeed simply by waiting longer. Usually, that indicates the cluster is sick.

pkg/core/system.go
glyn: I'm happy to reduce `retrySteps` to 5 on that basis.

glyn:
> Is this actually an issue that needs to be solved? I haven't seen a case where our current retry logic fails, but would eventually succeed simply by waiting longer. Usually, that indicates the cluster is sick.

I think @trisberg was hitting this problem repeatedly and that's why he raised https://github.com/projectriff/riff/issues/987. So I think it's worth retrying more than once.

trisberg:
LGTM, merging ...

PR #1071, opened 2019-01-24
Ignore credentials file

glyn:
This file is created when following the getting started docs and should not be
checked in.


fbiville:
Will merge once CI is ✅

Issue #972, opened 2018-11-16
Allow function builds without creating a Knative service


jldec:
As a developer i would like to do a function build without immediately creating a Knative service. This would allow me to take smaller steps when getting started E.g. to validate that the build environment is working and the necessary build parameters (like --artifact or --invoker) are correct.

PR #1070, opened 2019-01-24
DNM: Bump pack to 0.0.9

scothis:
Consume pack 0.0.9 in riff. We don't want to merge this as is because of buildpack/roadmap#42

PR #1080, opened 2019-01-25
DNM: `riff function local` commands

scothis:
Introduces two new commands:
- `riff function local build` which builds a local directory and publishes an image
- `riff function local run` which builds a local directory and runs the function

Both commands have no dependency on a Kubernetes cluster or an installed riff system. They do require a local docker daemon.

This PR is a superset of #1070, it must merge first.

So far this PR only contains superficial docs and tests. It is demo quality.


scothis:
Fixes #972

trisberg:
I think we should add flag for env vars to the `local run` command

```
  --env stringArray                environment variable expressed in a 'key=value' format
```

trisberg:
If I specify a port we should set environment variable for the port as PORT and SERVER_PORT so the app can pick these up

scothis:
> I think we should add flag for env vars to the local run command

I assume you're proposing these as runtime env vars? Would we ever need to also set build env vars, if so how would a user know the difference?

> If I specify a port using `function local run --port` we should set environment variable for the port as PORT and SERVER_PORT so the app can pick these up

The running container shouldn't care about the value of `--port`. The flag exposes a port that is running inside the container to the host. The build pack already exposes 8080, which is what provides the default `--port` value. We could hard code `PORT=8080` which is normally set as part of the runtime contract.

fbiville:
Nit: so far, the riff command structure has always been `riff <resource_type> <action>`, it would be nice if these new commands could stick to this scheme.

trisberg:
> I assume you're proposing these as runtime env vars

Yes, that was my intention, although looking at `pack run` I don't see a way to pass those in, so ignore my env var and port suggestions

scothis:
This PR is now focused on only the run command, the description has been updated.

Issue #1081, opened 2019-01-28
For in-cluster builds, use RUN_IMAGE from build template rather than hard-coded default


glyn:
When `riff` does an in-cluster build, it overrides the value of `RUN_IMAGE` in the build template (typically `packs/run`) with a hard-coded value (`packs/run:v3alpha2` at the time of writing).

This prevents a user from configuring a different value in the build template.

It should use the value in the build template instead.

PR #1082, opened 2019-01-28
For in-cluster builds, use RUN_IMAGE from build template

glyn:
... rather than hard-coded default

Fixes https://github.com/projectriff/riff/issues/1081

### DIFF
--- a/pkg/core/function.go
+++ b/pkg/core/function.go
@@ -168,10 +168,6 @@ func (c *client) makeBuildArguments(options CreateFunctionOptions) []build.Argum
 		// TODO configure buildtemplate based on buildpack image
 		// {Name: "TBD", Value: options.BuildpackImage},
 	}
-	if options.RunImage != "" {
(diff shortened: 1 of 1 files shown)


ericbottard:
Isn't that going to break PFS?

glyn:
> Isn't that going to break PFS?

Not really. It would only break PFS if in-cluster builds didn't work with the build template, but then the user could edit the build template `RUN_IMAGE` value rather than attempting to pass a value using an environment variable (which I think should be reserved for local builds).

glyn:
For in-cluster builds, the relocated image value should come from the build template.

ericbottard:
The build template uses a BuildTemplate variable. How does that variable gets its value is my point.
You're saying it will use the default value (after it has been rewritten by relocation)? If so, then it shouldn't be a BuildTemplate parameter at all, don't you agree?

glyn:
Yes. Thanks for filling in a gap in my knowledge.

PR #1083, opened 2019-01-28
Use a cluster BuildTemplate

### LINKED ISSUE #968: Centralize the riff-cnb buildtemplate (cluster scope)
The riff-cnb build template is currently a namespaced resource. At the time, this was the only option short of baking the build steps into each build definition. Now there is a ClusterBuildTemplate in addition to BuildTemplate that is cluster scoped. We could move the build template install as part of the system install and leave the namespace init to setup credentials.

scothis:
A ClusterBuildTemplate has the advantage of being defined once in the
cluster and can be updated and maintained with the rest of riff.

The cache PVC has moved into a separate resource as it and a service
account are still required in each namespace in order to run builds.

Fixes #968

### DIFF
--- a/main.go
+++ b/main.go
@@ -60,7 +60,7 @@ var (
 				"https://storage.googleapis.com/knative-releases/serving/latest/serving.yaml",
 				"https://storage.googleapis.com/knative-releases/eventing/latest/eventing.yaml",
 				"https://storage.googleapis.com/knative-releases/eventing/latest/in-memory-channel.yaml",
-				"https://storage.googleapis.com/projectriff/riff-buildtemplate/riff-cnb-clusterbuildtemplate.yaml"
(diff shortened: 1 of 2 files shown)


scothis:
Depends on https://github.com/projectriff/riff-buildpack-group/pull/21

scothis:
@glyn you're looking for https://github.com/projectriff/riff/pull/1083/files#diff-7ddfb3e035b42cd70649cc33393fe32cR46, although this file doesn't exist yet since we're waiting for 
https://github.com/projectriff/riff-buildpack-group/pull/21 to merge which will then publish that file.

glyn:
Ok, I missed that. Thanks.

scothis:
Updated to use `0.2.0-snapshot-ci-8ee79f9144fc` which is published as a ClusterBuildTemplate

Issue #817, opened 2018-09-21
Allow for multiple push credentials


scothis:
It should be possible to have active credentials for multiple image registries. So that Build CRD can push images to difference locations depending on the function.

Currently `riff namespace init` requires one of a manual secret, Docker Hub or GCR credentials. If a preexisting secret exists, it is replaced with the new value. New credentials should augment the service account to allow for migrating to a new image registry.

fbiville:
Apologies in advance if I misunderstand the current state of riff and Knative:
> If a preexisting secret exists, it is replaced with the new value
It seems to me that the current impl preserves the pre-existing secret, if they share the same name (https://github.com/projectriff/riff/blob/master/pkg/core/namespace.go#L142).
TheIsn't it sufficient to bind a new secret with a different name to the existing service account to have subsequent builds

scothis:
> It seems to me that the current impl preserves the pre-existing secret, if they share the same name 

Good question, registering the secret is a two step process:
1. create the Secret resource
2. reference the Secret in the ServiceAccount

The line you linked to is when the secret is registered with the ServiceAccount. We [delete a preexisting secret](https://github.com/projectriff/riff/blob/master/pkg/core/namespace.go#L109) as a part of step 1.

> is it sufficient to bind a new secret with a different name to the existing service account

I think so, but we'll need to try it to verify.

fbiville:
Thanks for clearing that up.

Without changing code, my assumption is that running `namespace init` against the same namespace with different secret names each time would do the job (although I'm not saying this is a good solution). 

Back to the initial suggestion, do we want to get rid of the secret overwrite in 100% of cases? Shouldn't we remove the preexisting secrets if the metadata annotations of the existing and new secrets are equivalent?

trisberg:
I think we should add a command to add credentials to an already existing `riff-build` service account.

We would require the secret name to be something other than `push-credentials` and the service account `riff-build` to already exist, the rest of the flags would match `namespace init`
```
riff namespace add-credentials --secret extra-creds [--gcr <path>] [--docker-hub <docker-id>] [--registry <registry-url> --registry-user <user>]
```

fbiville:
Note: these added credentials should be removed when running `riff namespace cleanup`.

fbiville:
Here is the updated specification for this issue:

[...]

PR #1088, opened 2019-01-31
Namespace cleanup

fbiville:
I added an initial unrelated commit to fix the Goland warnings in code areas near the ones that needed to change (it's just too annoying to the eye).


cmd/commands/cobra.go
```diff
@@ -333,11 +333,11 @@ func Visit(cmd *cobra.Command, f func(c *cobra.Command) error) error {
 }
 
 func PrintSuccessfulCompletion(cmd *cobra.Command) {
-	fmt.Fprintf(cmd.OutOrStdout(), "\n%s completed successfully\n", cmd.CommandPath())
+	_, _ = fmt.Fprintf(cmd.OutOrStdout(), "\n%s completed successfully\n", cmd.CommandPath())
```
scothis: What's producing the warning? Neither `go vet`, nor `golint` show anything on this line for me.

I'm not sure the visual noise of the blank identifier helps.

cmd/commands/namespace.go
```diff
@@ -92,3 +97,40 @@ func NamespaceInit(manifests map[string]*core.Manifest, kc *core.KubectlClient)
...
+			}
+
+			PrintSuccessfulCompletion(cmd)
+			return nil
+		},
+	}
+
+	LabelArgs(command, "NAMESPACE")
+
+	command.Flags().BoolVarP(&options.RemoveNamespace, "remove-ns", "", false, "removes the namespace as well")
```
scothis: What's the point? If you're going to delete the namespace just run `kubectl delete ns <name>` and call it a day.

`riff namespace init` doesn't create the namespace, it feels asymmetrical for riff to delete the namespace.

cmd/commands/namespace.go
```diff
@@ -92,3 +97,40 @@ func NamespaceInit(manifests map[string]*core.Manifest, kc *core.KubectlClient)
...
+	command.Flags().StringVarP(&options.SecretName, "secret", "s", "", "the name of the `secret` to delete")
```
scothis: I'd rather we attach an annotation/label to secrets that `riff namespace init` creates and use that as a key to clean up the secret. In many cases, users will not know the name of the secret that was created for them. It's also possible there will be more than one secret if we implement #817 

Alternately, we could look at the `riff-build` service account and uncover the secrets attached to it.

pkg/core/namespace.go
```diff
@@ -32,6 +32,7 @@ import (
 )
 
 const serviceAccountName = "riff-build"
+const persistentVolumeClaimName = "riff-cnb-cache"
```
scothis: This value is entirely determined by riff-buildpack-group and not the riff cli.

fbiville: It's indirectly defined by the riff cli, in the [manifest#Namespace](https://github.com/projectriff/riff/blob/10dee37c4e5769b069cc1edeb861b8dcb24000a0/main.go#L50) section. I agree this is definitely a shortcut though but I'm not sure about the alternatives.

Shall we ask again for the manifest? I think it's not a viable option, because:

 - users could forget what manifest they used (you said so for secrets ;-))
 - users may have upgraded riff and the manifest may have changed completely

Shall we attach an annotation somewhere?

cmd/commands/cobra.go
fbiville: Goland does. Here is an example:

screenshot 2019-02-01 at 09 34 44

The blank identifier may be a visual noise but clearly states that the error is ignored, while one cannot make the difference between a statement that does not yield anything versus something that we ignore.

fbiville:
Flagging as DNM, as it depends now on https://github.com/projectriff/riff/pull/1090.

cmd/commands/namespace.go
scothis: Ah, I missed that we do that.

pkg/core/namespace.go
scothis: If we slapped a label on all resource created by `namespace init` then we can find which resources should be deleted.

I agree that we don't want to require a manifest here. Ideally, the manifest should only be used for `system install` and never asked for again.

fbiville:
PR rebased now that master includes https://github.com/projectriff/riff/commit/e7d65db24ce39375ca3f3eb535ffb1ad4177c58d.

Demo:

## Cleanup without namespace deletion
```shell
 $ kubectl get pvc,secret,serviceaccount --namespace foobar
No resources found.
 $ riff namespace init foobar --gcr /path/to/some.json
 $ kubectl get pvc,secret,serviceaccount --namespace foobar
NAME                                   STATUS    VOLUME                                     CAPACITY
ACCESS MODES   STORAGECLASS   AGE
persistentvolumeclaim/riff-cnb-cache   Bound     pvc-8f0ae081-2bc6-11e9-b6a6-42010aa40030   8Gi
RWO            standard       29s

NAME                            TYPE                                  DATA      AGE
secret/default-token-hvkgl      kubernetes.io/service-account-token   3         30s
secret/istio.default            istio.io/key-and-cert                 3         30s
secret/istio.riff-build         istio.io/key-and-cert                 3         29s
secret/push-credentials         kubernetes.io/basic-auth              2         30s
secret/riff-build-token-87jk2   kubernetes.io/service-account-token   3         30s

NAME                        SECRETS   AGE
serviceaccount/default      1         30s
serviceaccount/riff-build   2         30s
 $ riff namespace cleanup foobar
Deleting serviceaccounts matching labels map[created-by:riff-unknown] in namespace "foobar"
Deleting persistentvolumeclaims matching labels map[created-by:riff-unknown] in namespace "foobar"
Deleting secrets matching labels map[created-by:riff-unknown] in namespace "foobar"

riff namespace cleanup completed successfully
 $ kubectl get pvc,secret,serviceaccount --namespace foobar
NAME                         TYPE                                  DATA      AGE
secret/default-token-hvkgl   kubernetes.io/service-account-token   3         1m
secret/istio.default         istio.io/key-and-cert                 3         1m
secret/istio.riff-build      istio.io/key-and-cert                 3         36s

NAME                     SECRETS   AGE
serviceaccount/default   1         1m
```

## Cleanup with namespace deletion
```shell
 $ kubectl get pvc,secret,serviceaccount --namespace foobar # same as before
 $ riff namespace init foobar --gcr /path/to/some.json
 $ kubectl get pvc,secret,serviceaccount --namespace foobar  # same as before
 $ riff namespace cleanup foobar --remove-ns
Deleting serviceaccounts matching labels map[created-by:riff-unknown] in namespace "foobar"
Deleting persistentvolumeclaims matching labels map[created-by:riff-unknown] in namespace "foobar"
Deleting secrets matching labels map[created-by:riff-unknown] in namespace "foobar"
Deleting namespace "foobar"

riff namespace cleanup completed successfully
 $ kubectl get pvc,secret,serviceaccount --namespace foobar
No resources found.
```

## Cleanup with default namespace deletion 🔥 

```shell
 $ riff namespace cleanup default --remove-ns

Error: the default namespace cannot be removed
```

fbiville:
@scothis @glyn let me know if the cleanup is sufficient in the case we don't delete the namespace (as you can see, some secrets and service account still float around afterwards).

pkg/core/namespace.go
fbiville: Done in https://github.com/projectriff/riff/commit/e7d65db24ce39375ca3f3eb535ffb1ad4177c58d

cmd/commands/namespace.go
fbiville: This option is now removed as riff has labels it can rely upon.

pkg/core/namespace.go
```diff
@@ -248,3 +285,96 @@ func readPassword(s string) (string, error) {
 		return string(res), err
 	}
 }
+
+func getInitLabels() map[string]string {
+	return map[string]string{"created-by": fmt.Sprintf("%s-%s", env.Cli.Name, env.Cli.Version)}
```
glyn: I'd like to understand the intention here. Which versions of riff should be able to cleanup namespaces created by which other versions of riff?

glyn:
> @scothis @glyn let me know if the cleanup is sufficient in the case we don't delete the namespace (as you can see, some secrets and service account still float around afterwards).

Presumably the resources which are not cleaned up in the case where we don't delete the namespace, are precisely those which don't have the correct riff label. If so, that seems reasonable, but then the big question is: which resources should have the riff label? For instance, if I issue `riff system install`, will all the relevant, namespace-related resources created be labelled with the riff label? If not, why not.

pkg/core/namespace.go
fbiville: That's a good question and I'm not sure what we want to do here.

glyn: Unless `created-by` is a conventional label, which doesn't appear to be the case, we could change this label to, for example, `"projectriff.io/riff-version":"0.3.0-snapshot"`. I'm suggesting using a prefix in the key to avoid clashing with user-defined labels.

I would suggest that we keep the semantics simple at first and ignore the value of the version in the label, so any version of riff could cleanup the resources of any other version. But having the version in the label is informative for users and would allow us to take into account the version if that becomes necessary.

scothis: Shouldn't these be two separate labels? Who installed the resource is relevant during cleanup. The version of the resource is relevant during upgrade.

pkg/core/namespace.go
```diff
@@ -287,13 +288,17 @@ func readPassword(s string) (string, error) {
...
+		"projectriff.io/version":   env.Cli.Version,
+	}
 }
 
 func transformToSelector(labels map[string]string) string {
 	builder := strings.Builder{}
-	for key, value := range labels {
-		builder.WriteString(fmt.Sprintf("%s=%s,", key, value))
+
+	for _, key := range sortedKeysOf(labels) {
```
glyn: I thought we agreed not to match on version when cleaning up?

glyn:
Just a thought: should we allow pfs to clean up resources from riff and possibly _vice versa_? Think about the upgrade/downgrade paths between using riff and pfs.

fbiville:
@glyn now both installer name and version are ignored, which means PFS can clean up riff resources and vice versa

cmd/commands/namespace.go
```diff
@@ -92,3 +97,37 @@ func NamespaceInit(manifests map[string]*core.Manifest, kc *core.KubectlClient)
...
+		Example: fmt.Sprintf(`  %[1]s namespace cleanup my-ns
+  %[1]s namespace cleanup my-ns --remove-ns`, env.Cli.Name),
+		Args: ArgValidationConjunction(
+			cobra.ExactArgs(namespaceCleanupNumberOfArgs),
+			AtPosition(namespaceCleanupNameIndex, ValidName()),
+		),
+		RunE: func(cmd *cobra.Command, args []string) error {
+			options.NamespaceName = args[namespaceCleanupNameIndex]
+			if options.NamespaceName == "default" && options.RemoveNamespace {
+				return fmt.Errorf("the default namespace cannot be removed")
```
glyn: It might be worth adding some words to this error message to make it clear that the resources have not being cleaned up.

fbiville: changed to `"cleanup canceled: the default namespace cannot be removed"`

glyn: Thanks. That's better.

fbiville:
Merged in a5454c3f and dc53fdd5

Issue #1074, opened 2019-01-24
Provide inverse operation to `riff namespace init`


glyn:
This is desirable from a UX perspective. See, for example, rule 6 [here](https://faculty.washington.edu/jtenenbg/courses/360/f04/sessions/schneidermanGoldenRules.html).

glyn:
See https://github.com/projectriff/riff/issues/1025 for a description of the resources that need cleaning up.

fbiville:
After chatting with @glyn, proposal is to have:
```
riff namespace cleanup [--remove-ns]
```
The associated service account, its secret the related PVC will be removed.
If `--remove-ns` is provided, the namespace itself will be deleted.

For now, the implementation would rely on the hardcoded names of service account/secret/PVC to remove them.

ericbottard:
Cool.
Note that `--remove-ns` doesn't bring any value over `kubectl delete ns`, has deleting it will destroy the riff things

ericbottard:
Or [maybe not](https://github.com/projectriff/riff/issues/1025)

fbiville:
That's the whole point of this command: cleaning up the remaining service account, secret and PVC. The logic could be reused for #1025.

The `--remove-ns` option is just to indicate whether the namespace itself should be deleted as well.
`riff namespace init` creates the namespaces only it it wasn't there in the first place. 
We cannot assume that riff created the namespace hence we should not delete it unless prompted to.

ericbottard:
Indeed, my bad. My point was more because I thought the namespace "owned" the account, secret, etc and hence it would have teared them down with itself.

Issue #1025, opened 2018-12-18
Uninstall orphans namespace init resources


scothis:
`riff system uninstall` cleans up the installed system, however, it does not attempt to cleanup resources created by `riff namespace init`. The init command will install a BuildTemplate (deleted when the CRD is removed) a Secret and a PersistentVolumeClaim. The secret and pvc are left behind. They should be cleaned up with the rest of riff.

If we implement #993 the pvc will be cleaned up leaving only the secret.

scothis:
Fixed by #1088

PR #1090, opened 2019-02-01
Label namespace resources at init time

fbiville:
The label will be leveraged by #1074 and possibly #1025 to
remove the resources (secrets, PVC) when their namespace is to be cleaned up or when riff/PFS
is uninstalled.


fbiville:
PR updated with integration of [`kustomize`](https://github.com/kubernetes-sigs/kustomize) as a library to apply the same label on namespace resources as done on secrets in the previous commit.

fbiville:
PR rebased and ready for (re-)review.

pkg/core/kustomize/kustomize.go
```diff
@@ -0,0 +1,141 @@
...
+	return kust.runBuild()
+}
+
+func (kust *kustomizer) writeResourceFile(resourceUri *url.URL) (string, error) {
+	resourceContents, err := kust.fetch(resourceUri)
+	if err != nil {
+		return "", err
+	}
+	resourcePath := "resource.yaml"
+	err = kust.fs.WriteFile(kust.fakeDir+resourcePath, []byte(resourceContents))
```
glyn: What stops resource files for distinct resources colliding?

pkg/core/kustomize/kustomize.go
```diff
@@ -0,0 +1,141 @@
...
+	kustomizeBuildCommand.SetArgs([]string{kust.fakeDir})
+	kustomizeBuildCommand.SetOutput(ioutil.Discard)
+	_, err := kustomizeBuildCommand.ExecuteC()
+	if err != nil {
+		return nil, err
+	}
+	return out.Bytes(), nil
+}
+
+func (kust *kustomizer) fetch(url *url.URL) (string, error) {
```
glyn: Why not use `fileutils.Read` since that can read from local files as well as URLs?

pkg/core/kustomize/kustomize.go
```diff
@@ -0,0 +1,141 @@
...
+
+func (kust *kustomizer) fetchUri(url *url.URL) (string, error) {
+	response, err := kust.httpClient.Get(url.String())
+	if err != nil {
+		return "", err
+	}
+	defer func() {
+		err := response.Body.Close()
+		if err != nil {
+			panic(err)
```
glyn: Panicking seems a bit brutal.

pkg/core/kustomize/kustomize.go
fbiville: there is only one resource file per `ApplyLabel` call, I don't think there can be be collisions in that case.

pkg/core/kustomize/kustomize_test.go
```diff
@@ -0,0 +1,156 @@
...
+
+	It("customizes local resources with provided label", func() {
+		file, fileUri := localFile("pvc.yaml", initialResourceContent)
+		defer unsafeClose(file)
+
+		resourceUrl := unsafeParseUrl(fileUri)
+
+		result, err := kustomizer.ApplyLabel(resourceUrl, ownerLabel)
+
+		Expect(err).To(Not(HaveOccurred()))
```
glyn: Isn't `Expect(err).NotTo(HaveOccurred())` is more popular in our other tests? It reads more like English.

pkg/core/kustomize/kustomize_test.go
```diff
@@ -0,0 +1,156 @@
...
+func serveSlow(listener net.Listener, response string, sleepDuration time.Duration) {
+	err := http.Serve(listener, http.HandlerFunc(func(responseWriter http.ResponseWriter, req *http.Request) {
+		time.Sleep(sleepDuration)
+		responseWriter.WriteHeader(200)
+		responseWriter.Header().Add("Content-Type", "application/octet-stream")
+		_, _ = io.WriteString(responseWriter, response)
+	}))
+
+	if err != nil {
+		panic(err)
```
glyn: Usually in tests, we use `Expect(err).NotTo(HaveOccurred())` rather than panicking.

pkg/core/kustomize/kustomize_test.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	return
+}
+
+func localFile(name string, content string) (*os.File, string) {
```
glyn: Take a look at the `test_support` package and how that is used elsewhere. This seems to duplicate some of the code.

pkg/core/kustomize/kustomize.go
fbiville: one thing I could do at the end of `ApplyLabel` is have a deferred function call that resets the `map[string]*FakeFile`, which underlies the in-memory filesystem implementation that is used here.

pkg/core/kustomize/kustomize.go
fbiville: shall I just log a warning?

pkg/core/kustomize/kustomize.go
```diff
@@ -0,0 +1,141 @@
...
+
+type kustomizer struct {
+	fakeDir    string
+	fs         fs.FileSystem
+	httpClient *http.Client
+}
+
+func MakeKustomizer(timeout time.Duration) Kustomizer {
+	return &kustomizer{
+		fs:      fs.MakeFakeFS(),
```
glyn: Probably worth a comment about why a _fake_ FS is being used. Is it really just an in-memory FS?

pkg/core/kustomize/kustomize.go
glyn: > there is only one resource file per `ApplyLabel` call, I don't think there can be be collisions in that case.

Fair enough, thanks.

Strictly speaking (and I know the current caller doesn't do this, but) if multiple calls to `ApplyLabel` were made concurrently, then collisions would occur. If this bothers you, you could add a godoc comment to `ApplyLabel` saying it is not suitable for concurrent invocation.

fbiville:
Regarding the kustomize approach, I agree that the integration is not straightforward.
What kustomize buys us is the out-of-the-box ability to decorate Kubernetes resources with any metadata we want.

https://github.com/kubernetes-sigs/kind has followed this path (and the current implementation is heavily inspired from it): https://github.com/kubernetes-sigs/kind/pull/77 (context: https://github.com/kubernetes-sigs/kustomize/issues/142).

pkg/core/kustomize/kustomize.go
fbiville: No strong reasons for it, I'm open to revisit that assumption.
I started with that as the [other integration](https://github.com/kubernetes-sigs/kind/pull/77) did the same.
I also find it easier to reason about, as we do not need to deal with permissions and other OS peculiarities. Moreover, it seems we do not process huge documents at the moment.

pkg/core/kustomize/kustomize.go
fbiville: done and it's now responsibility of pkg/fileutils/read.go#downloadFile

pkg/core/kustomize/kustomize_test.go
fbiville: fixed, thanks for the correction, I never know which one between "to not do" and "not to do" is correct

pkg/core/kustomize/kustomize_test.go
fbiville: done

pkg/core/kustomize/kustomize_test.go
fbiville: done

pkg/core/kustomize/kustomize.go
fbiville: done with a twist, extracted the common logic in a new function `pkg/fileutils/read.go#ReadUrl`, with a timeout. Existing code will work as before (a timeout of 0 is passed).

pkg/core/namespace.go
```diff
@@ -97,13 +99,17 @@ func (c *kubectlClient) NamespaceInit(manifests map[string]*Manifest, options Na
 		return err
 	}
 
+	ownerLabel := &kustomize.Label{
```
scothis: might also be nice to include the version of riff as well

pkg/core/kustomize/kustomize.go
```diff
@@ -0,0 +1,100 @@
...
+	// Returns the customized resource contents
+	// Returns an error if
+	// - the URL scheme is not supported (only file, http and https are)
+	// - retrieving the content fails
+	// - applying the customization fails
+	// As of the current implementation, it is not safe to call this function concurrently
+	ApplyLabel(resourceUri *url.URL, label *Label) ([]byte, error)
+}
+
+type Label struct {
```
scothis: k8s just uses a `map[string]string`. This also gives us the ability to apply multiple labels.

pkg/core/namespace.go
```diff
@@ -165,7 +171,12 @@ func (c *kubectlClient) NamespaceInit(manifests map[string]*Manifest, options Na
 		}
 
 		fmt.Printf("Applying %s in namespace %q\n", release, ns)
-		log, err := c.kubeCtl.Exec([]string{"apply", "-n", ns, "-f", resource})
+		resourceUrl, _ := url.Parse(resource)
+		labeledContent, err := c.kustomizer.ApplyLabel(resourceUrl, ownerLabel)
```
scothis: Should we also apply this to every other resource that is applied from the manifest?

fbiville: I was focusing on namespace-related resources. Would you mind if I address this in another issue/PR?

pkg/core/namespace.go
fbiville: done

pkg/core/kustomize/kustomize.go
fbiville: done

fbiville:
Merged in e7d65db24ce39375ca3f3eb535ffb1ad4177c58d

Issue #1027, opened 2018-12-19
define a CRD for riff system


sbawaska:
Since riff system is made up of a number of components like istio, knative, buildpacks etc and components around eventing planned, we should define a CRD which will help tying these components together. This CRD will be leveraged to install/uninstall/upgrade the riff system.

```
apiGroup: "projectriff.io/v1alpha1"
kind: Manifest
metadata:
  name: riff, riff-namespace, in-memory-channel, kafka-channel, container-source
spec:
    dependencies:
    - istio
    images:
    - projectriff/correlator:s1p2018
    - ....
    resources:
    - path: https://example.com/foo.yaml
      content: | # inlined path
      hash: deadbeef1234
      checks:
      - apiGroup: v1
        kind: Pod
        namespace: knative-serving
        name:# using a label selector instead
        labelSelector: app=controller
        jsonPath: '...'
        pattern: ;Ready=True;
    - # second resource set 
```

Once we have an object like this, we could guide the install by iterating over all resources and checking  the status of the resource before deploying the next resource. In the case of a repeat install due to a previous install failure, we can quickly do statusCheck of the resource and continue from the failure point.

Thanks @scothis for your valuable inputs in defining the structure of the CRD/object!

glyn:
What are the semantics, if any, of the metdata names?

scothis:
done in https://github.com/projectriff/cnab-k8s-installer-base

PR #1091, opened 2019-02-02
#1027 define crd

sbawaska:
As described in #1027 introducing a crd to help keep track of components installed by riff.


pkg/core/system.go
```diff
@@ -164,24 +204,79 @@ To fix this you need to:
...
+	for k,v := range m {
+		s += k + "=" + v + ","
+	}
+	if last := len(s) - 1; last >= 0 && s[last] == ',' {
+		s = s[:last]
+	}
+	return s
+}
+
+func (c *client) SystemUninstall(options SystemUninstallOptions) (bool, error) {
```
scothis: these changes should be pulled into a separate PR. The Manifest CRD has no direct impact on uninstall. At least not in its current incarnation.

pkg/constants/constants.go
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package constants
+
+import (
+	"fmt"
+)
+
+var (
```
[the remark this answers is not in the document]
sbawaska: I will move this out of `pkg` folder. Had to move out of main due to cyclic package imports.

pkg/core/system.go
```diff
@@ -55,100 +63,132 @@ var (
...
+	})
+	if err == wait.ErrWaitTimeout {
+		return nil, errors.New(fmt.Sprintf("timed out creating %s custom resource defiition", env.Cli.Name))
 	}
+	return crdManifest, err
+}
 
-	return nil
+//TODO this is a stop-gap, remove core.Manifest in favor of crd.Manifest
+func buildCrdManifest(manifest *Manifest) (*crd.Manifest, error) {
```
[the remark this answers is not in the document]
sbawaska: same question as above.

pkg/core/system.go
```diff
@@ -164,24 +204,79 @@ To fix this you need to:
...
 		}
-		if willRetry {
-			return false, nil // retriable error
+		return err
+	}
+	return nil
+}
+
+// TODO this only supports checking Pods for phases
+func (c *client) checkResource(resource crd.RiffResources) error {
```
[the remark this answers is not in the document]
sbawaska: cool! I will throw an error if Resource kind is not Pod.

pkg/core/system.go
sbawaska: cool!

pkg/crd/client.go
```diff
@@ -0,0 +1,96 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
+)
+
+type riffClient struct {
+	restClient rest.Interface
+}
+type Client interface {
```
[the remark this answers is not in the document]
sbawaska: mainly because I did not know it existed :)

pkg/crd/crd.go
```diff
@@ -0,0 +1,65 @@
...
+	Group = "projectriff.io"
+	Version = "v1alpha1"
+	Kind = "RiffSystem"
+)
+
+func CreateCRD(clientset extClientset.Interface) error {
+	_, err := clientset.ApiextensionsV1beta1().CustomResourceDefinitions().Create(
+		&extApi.CustomResourceDefinition{
+			ObjectMeta: meta_v1.ObjectMeta{
+				Name: "riff-system.projectriff.io",
```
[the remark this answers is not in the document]
sbawaska: will address.

pkg/crd/crd.go
```diff
@@ -0,0 +1,65 @@
...
+				Versions: []extApi.CustomResourceDefinitionVersion {
+					{
+						Name:    Version,
+						Served:  true,
+						Storage: true,
+					},
+				},
+				Scope: extApi.ClusterScoped,
+				Names: extApi.CustomResourceDefinitionNames{
+					Singular: "riff-system",
```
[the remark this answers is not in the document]
sbawaska: will address.

pkg/crd/crd.go
```diff
@@ -0,0 +1,65 @@
...
+					Plural: "riff-system",
+					Kind: reflect.TypeOf(Manifest{}).Name(),
```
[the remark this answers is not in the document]
sbawaska: will address.

pkg/crd/manifest.go
```diff
@@ -0,0 +1,280 @@
...
+							Pattern:  "Running",
+						},
+					},
+				},
+				{
+					Path: fmt.Sprintf("https://storage.googleapis.com/projectriff/riff-buildtemplate/riff-cnb-clusterbuildtemplate-%s.yaml", constants.BuilderVersion),
+					Name: "riff-build-template",
+				},
+				{
+					Path: fmt.Sprintf("https://storage.googleapis.com/projectriff/riff-buildtemplate/riff-cnb-cache-%s.yaml", constants.BuilderVersion),
```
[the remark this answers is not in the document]
sbawaska: will address.

pkg/crd/crd.go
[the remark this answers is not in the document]
scothis: That error is a bit strange give the current code, but I'd expect:

```
metadata.name = "manifests.system.projectriff.io"
spec.group = "system.projectriff.io"
spec.names.plural = "manifests"
spec.names.singular = "manifest"
spec.names.kind = "Manifest"
```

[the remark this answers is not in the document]
sbawaska: since there is only ever going to be one `riff system install` in a kubernetes cluster, I kept singular and plural the same.

sbawaska:
@scothis I have incorporated your feedback. Can you please review?

main.go
```diff
@@ -18,79 +18,235 @@ package main
...
+								Pattern:  "Running",
+							},
+						},
+					},
+					{
+						Path: fmt.Sprintf("https://storage.googleapis.com/projectriff/riff-buildtemplate/riff-cnb-clusterbuildtemplate-%s.yaml", builderVersion),
+						Name: "riff-build-template",
+					},
+				},
+				Init: []crd.RiffResource {
```
sbawaska: The current structure of the yaml is:
```
spec:
  resources: []
  init: []
```
I would like to change it to 
```
spec:
  resources:
    system: []
    initialization: []
```
I will create a new PR once this is merged.

pkg/crd/manifest.go
```diff
@@ -14,28 +14,63 @@
...
+	Pattern  string               `json:"pattern,omitempty"`
+}
 
-	"github.com/ghodss/yaml"
-)
+type Resource struct {
+	Path      string           `json:"path,omitempty"`
+	Content   string           `json:"content,omitempty"`
+	Name      string           `json:"name,omitempty"`
+	Namespace string           `json:"namespace,omitempty"`
```
scothis: How are Name and Namespace used? `Resource` here represents a config file that may contain multiple k8s resources. Each of those resources will have a name and namespace (if applicable)

pkg/crd/manifest.go
```diff
@@ -14,28 +14,63 @@
...
+	Checks    []ResourceChecks `json:"checks,omitempty"`
+}
 
-const manifestVersion_0_1 = "0.1"
+type RiffResource struct {
+	System         []Resource `json:"system,omitempty"`
```
scothis: What is the a distinction between system and initialization?

I'd expect the difference between `riff system install` and `riff namespace init` to be which manifest is applied, not a split within a single manifest.

pkg/crd/manifest.go
```diff
@@ -14,28 +14,63 @@
...
+	Initialization []Resource `json:"initialization,omitempty"`
+}
+
+type RiffSpec struct {
+	Images    []string     `json:"images,omitempty"`
```
scothis: Is this sufficient? The image manifest structure was previously `map[string]string`

pkg/crd/manifest.go
```diff
@@ -14,28 +14,63 @@
...
+	Resources RiffResource `json:"resources,omitempty"`
+}
+
+type RiffStatus struct {
+	Status string `json:"status,omitempty"`
```
scothis: Is this used?

pkg/crd/manifest.go
```diff
@@ -14,28 +14,63 @@
...
-	Istio           []string `json:"istio"`
-	Knative         []string `json:"knative"`
-	Namespace       []string `json:"namespace"`
-	manifestDir     string
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec        RiffSpec   `json:"spec,omitempty"`
+	Status      RiffStatus `json:"status,omitempty"`
+	manifestDir string
```
scothis: CRDs are public interfaces, they should not have unexported fields.

pkg/crd/manifest.go
```diff
@@ -116,24 +156,22 @@ func (m *Manifest) ResourceAbsolutePath(path string) (string, error) {
 
 func checkCompleteness(m Manifest) error {
 	var omission string
-	if m.Istio == nil {
-		omission = "istio"
-	} else if m.Knative == nil {
-		omission = "knative"
-	} else if m.Namespace == nil {
-		omission = "namespace"
+	if m.Spec.Resources.System == nil {
```
scothis: Validating the CR should be part of the CRD, instead of the runtime usage. For example https://github.com/knative/serving/blob/master/pkg/apis/serving/v1alpha1/service_validation.go

The amount of validation also needs to increase corresponding with the new capabilities in the CRD.

pkg/crd/manifest.go
```diff
@@ -14,28 +14,63 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
 	"net/url"
 	"path/filepath"
+	"strings"
+)
 
-	"github.com/projectriff/riff/pkg/fileutils"
+type ResourceChecks struct {
+	Kind     string               `json:"kind,omitempty"`
+	Selector metav1.LabelSelector `json:"selector,omitempty"`
```
scothis: How to I specify the namespace? Is this what the namespace on the Resource is for? What if the Resource contains resources in different namespaces?

Can I lookup a resource by name?

pkg/crd/manifest.go
```diff
@@ -14,28 +14,63 @@
...
+	JsonPath string               `json:"jsonpath,omitempty"`
+	Pattern  string               `json:"pattern,omitempty"`
```
scothis: Is Pattern a literal string match, partial string match, glob, regex...?

pkg/crd/manifest.go
```diff
@@ -14,28 +14,63 @@
...
+}
 
-	"github.com/ghodss/yaml"
-)
+type Resource struct {
+	Path      string           `json:"path,omitempty"`
+	Content   string           `json:"content,omitempty"`
```
scothis: Are Path and Context mutually exclusive? What happens if I have both?

pkg/crd/manifest.go
```diff
@@ -14,28 +14,63 @@
...
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec        RiffSpec   `json:"spec,omitempty"`
+	Status      RiffStatus `json:"status,omitempty"`
+	manifestDir string
+}
+
+func (orig *Manifest) DeepCopyObject() runtime.Object {
+	result := &Manifest{}
+	copier.Copy(result, orig)
```
scothis: Is it worth introducing a new dependency for a deep copy?

k8s.io/code-generator generated clients will also generate the DeepCopy method.

pkg/crd/client.go
```diff
@@ -0,0 +1,96 @@
...
+	Create(obj *Manifest) (*Manifest, error)
+	Update(obj *Manifest) (*Manifest, error)
+	Delete(obj *Manifest) (*Manifest, error)
+	Get() (*Manifest, error)
+}
```
scothis: missing `List()`

If we generate the client, we don't have to implement list by hand.

pkg/core/system.go
```diff
@@ -55,58 +63,184 @@ var (
...
+			if strings.EqualFold(check.Kind, "Pod") {
+				ready, err = c.isPodReady(check, resource.Namespace)
+				if err != nil {
+					return err
+				}
+				if ready {
+					break
+				}
+			} else {
+				return errors.New("only Kind:Pod supported for resource checks")
```
scothis: How do we check that a LoadBalancer Service is ready?

The world is more than Pods.

pkg/core/system.go
```diff
@@ -55,58 +63,184 @@ var (
...
+	yaml, err := fileutils.Read(res.Path, filepath.Dir(options.Manifest))
+	if err != nil {
+		return err
+	}
+	if options.NodePort {
+		yaml = bytes.Replace(yaml, []byte("type: LoadBalancer"), []byte("type: NodePort"), -1)
+	}
+	// TODO HACK: use the RESTClient to do this
+	kubectl := kubectl.RealKubeCtl()
+	istioLog, err := kubectl.ExecStdin([]string{"apply", "-f", "-"}, &yaml)
```
scothis: istio?

pkg/core/system.go
```diff
@@ -55,58 +63,184 @@ var (
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (c *client) installResource(res crd.Resource, options SystemInstallOptions) error {
+	if res.Path == "" {
+		return errors.New("cannot install anything other than a url yet")
```
scothis: It will need to

pkg/core/system.go
```diff
@@ -55,58 +63,184 @@ var (
...
+	}
+	fmt.Printf("installing %s from %s...", res.Name, res.Path)
+	yaml, err := fileutils.Read(res.Path, filepath.Dir(options.Manifest))
+	if err != nil {
+		return err
+	}
+	if options.NodePort {
+		yaml = bytes.Replace(yaml, []byte("type: LoadBalancer"), []byte("type: NodePort"), -1)
```
scothis: Can we do this in kustomize, or some other tool with more semantic awareness?

pkg/core/kubectl_client.go
```diff
@@ -17,14 +17,14 @@
...
+	"github.com/projectriff/riff/pkg/crd"
 	"github.com/projectriff/riff/pkg/kubectl"
 	"k8s.io/client-go/kubernetes"
 )
 
 type KubectlClient interface {
-	SystemInstall(manifests map[string]*Manifest, options SystemInstallOptions) (bool, error)
 	SystemUninstall(options SystemUninstallOptions) (bool, error)
-	NamespaceInit(manifests map[string]*Manifest, options NamespaceInitOptions) error
+	NamespaceInit(manifests map[string]*crd.Manifest, options NamespaceInitOptions) error
```
scothis: Why does NamespaceInit accept manifests? I would expect it to read the manifest from the cluster and install that into a namespace.

ericbottard:
The riff repo only contains CLI code at the moment, and maybe it is best this way. If this CRD actually becomes real, shouldn't it live in a server-side dedicated repo anyway?

sbawaska:
Closing in favor of CNAB spike.

Issue #982, opened 2018-11-22
riff service invoke hides curl http error responses like 404


jldec:
It would be helpful if `riff service invoke` relayed http error responses like 404 to stderr.
e.g. currently you see this with `-v` but nothing if you don't use `-v`.

```
$ pfs service invoke square --text -- -w '\n' -d 7 -v
curl http://192.168.99.100:32380/ -H 'Host: square.default.example.com' -H 'Content-Type: text/plain' -w '\n' -d 7 -v
*   Trying 192.168.99.100...
* TCP_NODELAY set
* Connected to 192.168.99.100 (192.168.99.100) port 32380 (#0)
> POST / HTTP/1.1
> Host: square.default.example.com
> User-Agent: curl/7.60.0
> Accept: */*
> Content-Type: text/plain
> Content-Length: 1
> 
* upload completely sent off: 1 out of 1 bytes
< HTTP/1.1 404 Not Found
< location: http://square.default.example.com/
< date: Thu, 22 Nov 2018 13:31:05 GMT
< server: envoy
< connection: close
< content-length: 0
< 
* Closing connection 0

~/riff/node-fun-square (master)$

fbiville:
Reopening as the current implementation prints every status other than 200.

PR #1097, opened 2019-02-07
[DNM?] Ensure riff service invoke always prints any HTTP error

glyn:
This has only been tested on macOS, but it should be tested on Windows too, so please do not merge 
_unless_ you have tested on Windows. To provoke a 404, hack `service.go` thus:
```
hostHeader := fmt.Sprintf("Host: %s", hostName+"x")
```
(I can do this next week - I left my Windows VM at home!)

The curl flags -sS are supposed to suppress output, but print any errors. This
would be the ideal solution here. Unfortunately, this doesn't appear to work
(at least with curl 7.54.0 on macOS).

Fixes https://github.com/projectriff/riff/issues/982

### DIFF
--- a/cmd/commands/service.go
+++ b/cmd/commands/service.go
@@ -17,9 +17,13 @@
 package commands
 
 import (
+	"bytes"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
+	"regexp"
+	"strings"
 	"time"
 
 	"k8s.io/api/core/v1"
@@ -281,8 +285,16 @@ Additional curl arguments and flags may be specified after a double dash (--).`,
 				curlCmd.Args = append(curlCmd.Args, "-H", "Content-Type: text/plain")
 			}
 
+			verbose := false
+
 			if argsLengthAtDash > 0 {
-				curlCmd.Args = append(curlCmd.Args, args[argsLengthAtDash:]...)
+				curlArgs := args[argsLengthAtDash:]
+				for _, a := range curlArgs {
+					if verboseCurl(a) {
+						verbose = true
+					}
+				}
+				curlCmd.Args = append(curlCmd.Args, curlArgs...)
 			}
 
 			quoted, err := shellquote.Quote(curlCmd.Args)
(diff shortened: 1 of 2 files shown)


cmd/commands/service.go
```diff
@@ -304,6 +331,33 @@ Additional curl arguments and flags may be specified after a double dash (--).`,
...
+	matched, err := regexp.MatchString("< HTTP/[\\d\\.]* 200", curlErrOutput)
+	if err != nil {
+		panic(err) // Should never happen unless the regex is invalid
+	}
+
+	if !matched {
+		for _, line := range strings.Split(curlErrOutput, "\n") {
+			matched, err := regexp.MatchString("< HTTP/[\\d\\.]* ", line)
+			if err != nil {
+				panic(err) // Should never happen unless the regex is invalid
```
fbiville: nit: what about using https://golang.org/pkg/log/#Panicf instead?

cmd/commands/service_test.go
```diff
@@ -663,3 +664,73 @@ func pathAwareHttpServer(path string, pathMatchedChannel chan<- bool) net.Listen
...
 	return listener
 }
+
+var _ = Describe("PrintCurlHttpErrors", func() {
+	var (
+		input  string
+		output string
+	)
+
+	JustBeforeEach(func() {
```
fbiville: purely personal opinion: I tend to avoid `JustBeforeEach` as it makes the reading less linear but the setup is simple enough at this point.

cmd/commands/service.go
fbiville: nit: what about using https://golang.org/pkg/log/#Panicf?

fbiville:
Works both on Windows and Mac 🎉

cmd/commands/service.go
```diff
@@ -304,6 +331,33 @@ Additional curl arguments and flags may be specified after a double dash (--).`,
 	return command
 }
 
+// Print any HTTP errors in the given string to the given writer
+func PrintCurlHttpErrors(curlErrOutput string, w io.Writer) {
+	matched, err := regexp.MatchString("< HTTP/[\\d\\.]* 200", curlErrOutput)
```
fbiville: you can instead use `regexp.MustCompile` that gives you a Regexp struct or panics, and then call `MatchString` without the extraneous `err` (same applies below)

cmd/commands/service.go
```diff
@@ -304,6 +331,33 @@ Additional curl arguments and flags may be specified after a double dash (--).`,
...
+	if err != nil {
+		panic(err) // Should never happen unless the regex is invalid
+	}
+
+	if !matched {
+		for _, line := range strings.Split(curlErrOutput, "\n") {
+			matched, err := regexp.MatchString("< HTTP/[\\d\\.]* ", line)
```
fbiville: as pointed out by Goland, you don't need to escape `.` in a character class

fbiville:
I took the liberty of carrying out the changes suggested for `service.go`. Merged as c5e123d5144c474f9d30624d5e6c3f339594add3.

cmd/commands/service.go
glyn: No objections. Like I said, should never happen...

PR #1104, opened 2019-02-15
 Support basic auth secret 

### LINKED ISSUE #1103: Add option for creating secret in "namespace init" supporting basic auth of any registry
We currently only support basic auth credentials for dockerhub as part of `namespace init`

We should add an option to do this for any docker registry that uses basic auth.

I suggest adding:
   `riff namespace init default --registry-user someone --registry-server registry.example.com`

fbiville:
Fixes #1103.


pkg/core/namespace.go
```diff
@@ -210,6 +217,15 @@ func (kc *kubectlClient) createGcrSecret(options NamespaceInitOptions, labels ma
 	return kc.createBasicAuthSecret(options.NamespaceName, options.SecretName, "_json_key", string(token), "https://gcr.io", labels)
 }
 
+func (kc *kubectlClient) createRegistrySecret(options NamespaceInitOptions, labels map[string]string) error {
+	username := options.RegistryUser
+	password, err := readPassword(fmt.Sprintf("Enter password for user %q", username))
+	if err != nil {
+		return err
+	}
+	return kc.createBasicAuthSecret(options.NamespaceName, options.SecretName, username, password, options.RegistryServer, labels)
```
trisberg: should we add th eprotocol `https://` if it is not provided? Seems all example here use a protocol - https://github.com/knative/docs/blob/master/build/auth.md#guiding-credential-selection

fbiville: or should we reject it maybe? We could parse the URL and reject anything that does not have a http or https protocol?

trisberg: should we change the flag to be `--repository-url` then?

fbiville: `repository-url` or `registry-url`?

ericbottard:
I dont think this is a good Idea as typically an image coordinate does not
look like a URL

Le lun. 18 févr. 2019 à 18:25, Florent Biville <notifications@github.com> a
écrit :

pkg/core/namespace.go
trisberg: sorry, meant registry-url, although I would prefer to keep registry-host and just add the protocol

trisberg:
should we also require `--registry-default-path` in order not to conflict with #1105 

Then we could treat 

  `--dockerhub trisberg`

to be equivalent to

  `--registry-host index.docker.io`+`--registry-username trisberg`+`--registry-default-path trisberg`

trisberg:
and we could also add `--registry-protocol` that defaults to `https` but can be overridden for `http`  registries

fbiville:
Added the protocol option.
I just noticed that if I initialize a namespace with the registry host `index.docker.io/v1`, I get 401s when trying to create a function. It works fine with `index.docker.io/v1/`. That's quite strange.

fbiville:
I'd rather wait for #1105 to be merged before addressing the addition of `--registry-default-path`.

ericbottard:
LGTM, successfully tested against docker.io.
The flag validation logic seems legit too

ericbottard:
Merged as 7984f036

PR #1105, opened 2019-02-18
Default image for function create based on namespace init credentials

### LINKED ISSUE #1101: Provide default `--image` for `riff function create`
When creating a function, the location of the source code is often top of mind, but the image is secondary. It adds friction to manually specify an image repository. Particularly when much of the information is redundant with the credentials used to initialize the namespace. I constantly struggle to remember simple things like which gcloud project my cluster is running in.

I'd like to see the `--image` flag be optional with riff providing a reasonable default based on a base value saved in the namespace and the name of the function. The image repository base can be saved when running `riff namespace init`. Since there is strong correlation between the credentials being set and the image repository base, in many cases we can automatically set the default value, in other cases, we can require the user specify the base value.

- `--gcr` => `gcr.io/{<creds> | jq -r .project_id}`
- `--dockerhub` => `{username}`
- others => require `--image-prefix` to be set

In all cases, the user must be able to provide an explicit image to `riff function create`. It may be desirable to offer a shortcut to use the image repository prefix with a different repository, ie `--image _base/some-name-that-is-not-the-same-as-my-function`.

The exact mechanism to persist the prefix in the namespace is TBD, however, a ConfigMap named `riff-build` is a logical choice.

Current command:

```
riff function create square \
  --git-repo https://github.com/projectriff-samples/node-square \
  --image gcr.io/project-id/square
```

Becomes:

```
riff function create square \
  --git-repo https://github.com/projectriff-samples/node-square
```

The change is event more dramatic for local builds. Before:

```
riff function create square --local-path . --image gcr.io/project-id/square
```

After:

```
riff function create square --local-path .
```

scothis:
During`riff function create` if `--image` is not defined, a default value is provided based on the name of the function and a value (default-image-prefix) looked up from a ConfigMap (riff-build) in the target namespace. If the ConfigMap doesn't exist or the prefix is not present in the map, then --image will be required.

The ConfigMap is created during namespace init if it doesn't exist, otherwise it is updated. The value is automatically derived from the credentials being set for Docker Hub and GCR. Defining the secret out of band or using generic basic auth defined in #1103 provide insufficient metadata to set a prefix.

Docker Hub: `docker.io/{username}`
GCR: `gcr.io/{project-id}`

This has minor conflicts with #1104.

Possible future direction:
- direct get/set of the prefix value via the cli
- allow `function create` to customize the value after the prefix without specifying the whole image repository

Fixes #1101


pkg/core/namespace.go
```diff
@@ -208,7 +210,11 @@ func (c *kubectlClient) createDockerHubSecret(options NamespaceInitOptions, labe
 	}
 	fmt.Printf("Creating secret %q with DockerHub authentication for user %q\n", options.SecretName, options.DockerHubUsername)
 	_, err = c.kubeClient.CoreV1().Secrets(options.NamespaceName).Create(secret)
-	return err
+	if err != nil {
+		return err
+	}
+	prefix := fmt.Sprintf("docker.io/%s", options.DockerHubUsername)
+	return c.client.SetDefaultBuildImagePrefix(options.NamespaceName, prefix)
```
fbiville: should we log this operation, for consistency?

pkg/core/namespace.go
```diff
@@ -232,8 +238,17 @@ func (c *kubectlClient) createGcrSecret(options NamespaceInitOptions, labels map
...
+	if err != nil {
+		return err
+	}
+	tokenMap := map[string]string{}
+	err = json.Unmarshal(token, &tokenMap)
+	if err != nil {
+		return err
+	}
+	prefix := fmt.Sprintf("gcr.io/%s", tokenMap["project_id"])
+	return c.client.SetDefaultBuildImagePrefix(options.NamespaceName, prefix)
```
fbiville: should we log this operation, for consistency?

cmd/commands/function.go
```diff
@@ -66,7 +66,29 @@ func FunctionCreate(buildpackBuilder core.Builder, fcTool *core.Client, defaults
...
+					return fmt.Errorf("unable to default image: %s", err)
+				}
+				if prefix != "" {
+					// combine prefix and function name to provide default image
+					createFunctionOptions.Image = fmt.Sprintf("%s/%s", prefix, args[functionCreateFunctionNameIndex])
+				}
+			}
+
+			if createFunctionOptions.Image == "" {
+				return fmt.Errorf("required flag(s) \"image\" not set")
```
fbiville: should we log something about the missing prefix as well?

cmd/commands/function.go
```diff
@@ -66,7 +66,29 @@ func FunctionCreate(buildpackBuilder core.Builder, fcTool *core.Client, defaults
...
+				if err != nil {
+					return fmt.Errorf("unable to default image: %s", err)
+				}
+				if prefix != "" {
+					// combine prefix and function name to provide default image
+					createFunctionOptions.Image = fmt.Sprintf("%s/%s", prefix, args[functionCreateFunctionNameIndex])
+				}
+			}
+
+			if createFunctionOptions.Image == "" {
```
ericbottard: Could this move to the `else` of the `if` at line 81?

fbiville:
Note to self: one more thing to clean up in https://github.com/projectriff/riff/pull/1088

fbiville:
Now that https://github.com/projectriff/riff/pull/1088 is merged, we also need to remove the config map during `namespace cleanup`.

fbiville:
This PR now also needs to figure out how to generate an image prefix for any basic auth registries.

scothis:
> This PR now also needs to figure out how to generate an image prefix for any basic auth registries.

I don't think that's possible. If you have ideas, please share.

The best option I can come up with is to add specific commands to get/set the prefix value explicitly. I'd like to treat that as a separate PR since everything will still work if the default value is not set, the only difference is that the `--image` flag becomes required.

fbiville:
I believe @trisberg suggested to add an explicit option in that case.

trisberg:
I think we should add a `--registry-default-path` flag that defaults to the username for the registry but can be overridden

scothis:
> I think we should add a `--registry-default-path` flag that defaults to the username for the registry but can be overridden

Interesting, but I have a couple concerns.

1. We're introducing a lot of conditional optionality into the flags, that's a hint to me that maybe separate commands are better.
2. If we consider ACR and ECR as examples where we don't provide dedicated support, but do want a default, the value for this flag would need to be blank. Overriding a default value with an empty value is often a poor UX for CLIs. Assuming we didn't have richer support for GCR, the value would need to be overridden with the project-id.

scothis:
> This PR now also needs to figure out how to generate an image prefix for any basic auth registries.

I added an `--image-prefix` flag for namespace init. If not specified explicitly, a default value will be derived for docker hub and GCR registries. Basic auth and predefined secretes will need to specify the value. Otherwise the value will be set in the config map. If no value is defined, the config map will not be set and `--image` will be required for `riff function create` (with a better error message).

To change the value one could run a command like:

```
riff namespace init <ns> --no-secret --image-prefix registry.example.com
```

scothis:
All of the comments have been addressed. Assuming CI passes, this is good to go.

Issue #1107, opened 2019-02-19
namespace init fails when run with a manifest with local yaml files


glyn:
With a manifest referring to local YAML files, `riff namespace init` fails with:
```
riff namespace init default -m distro-download/manifest.yaml --gcr gcr-storage-admin.json
Initializing namespace "default"

Creating secret "push-credentials" with GCR authentication key from file gcr-storage-admin.json
Serviceaccount "riff-build" already exists in namespace "default" with secret "push-credentials". Skipping.
Applying riff-cnb-cache-0.2.0-snapshot-ci-63cd05079e1f.yaml in namespace "default"

Error: unsupported scheme in /Users/glyn/workspace/riff/distro-download/riff-cnb-cache-0.2.0-snapshot-ci-63cd05079e1f.yaml:
```

PR #1108, opened 2019-02-20
namespace init support for manifest with local yamls

glyn:
Fixes https://github.com/projectriff/riff/issues/1107

### DIFF
--- a/pkg/core/namespace.go
+++ b/pkg/core/namespace.go
@@ -178,6 +178,9 @@ func (c *kubectlClient) NamespaceInit(manifests map[string]*Manifest, options Na
 
 		fmt.Printf("Applying %s in namespace %q\n", release, ns)
 		resourceUrl, _ := url.Parse(resource)
+		if resourceUrl.Scheme == "" {
+			resourceUrl.Scheme = "file"
+		}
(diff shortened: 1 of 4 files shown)


pkg/core/namespace.go
```diff
@@ -178,6 +178,9 @@ func (c *kubectlClient) NamespaceInit(manifests map[string]*Manifest, options Na
 
 		fmt.Printf("Applying %s in namespace %q\n", release, ns)
 		resourceUrl, _ := url.Parse(resource)
+		if resourceUrl.Scheme == "" {
+			resourceUrl.Scheme = "file"
```
ericbottard: If `path` does not begin with a "slash" (TBD windows), the url needs to be made absolute. All URLs with scheme == file must use absolute paths

glyn: Understood. `resource` was returned from `fileutils.IsAbsFile` and so will be absolute - see the godoc of that function for details.

ericbottard: Ah, sorry, my bad. Did a quick test with just `url.Parse()` and saw that it could accept relative paths.
Please disregard then

PR #1112, opened 2019-02-21
Improve system uninstall

### LINKED ISSUE #1110: Unable to uninstall system
There seems to be something blocking the uninstall process

```
$ riff system uninstall --istio --force
Removing Knative for riff components
Deleting CRDs for knative.dev

Error: context deadline exceeded
```

trisberg:
- Delete any Knative services, configurations and routes to avoid blocking issues when CRDs are deleted with existing Knative resources

- no longer necessary to delete "horizontalpodautoscaler.autoscaling" resource for istio-pilot

Fixes #1110

### DIFF
--- a/pkg/core/system.go
+++ b/pkg/core/system.go
@@ -374,15 +374,15 @@ func deleteClusterResources(kc *kubectlClient, resourceType string, prefix strin
 
 func deleteKnativeServices(kc *kubectlClient) error {
 	fmt.Printf("Deleting Knative Services\n")
-	err := deleteAllKnative(kc, "service.serving")
+	err := deleteAllKnative(kc, "service.serving.knative.dev")
 	if err != nil {
 		return err
 	}
-	err = deleteAllKnative(kc, "configuration.serving")
+	err = deleteAllKnative(kc, "configuration.serving.knative.dev")
 	if err != nil {
 		return err
 	}
-	err = deleteAllKnative(kc, "route.serving")
+	err = deleteAllKnative(kc, "route.serving.knative.dev")
 	if err != nil {
 		return err
 	}
--- a/pkg/core/system.go
+++ b/pkg/core/system.go
@@ -197,6 +197,10 @@ func (kc *kubectlClient) SystemUninstall(options SystemUninstallOptions) (bool, 
 				return false, nil
 			}
 		}
+		err = deleteKnativeServices(kc)
+		if err != nil {
+			return false, err
+		}
 		fmt.Print("Removing Knative for " + env.Cli.Name + " components\n")
 		err = deleteCrds(kc, "knative.dev")
 		if err != nil {
@@ -266,8 +270,6 @@ func (kc *kubectlClient) SystemUninstall(options SystemUninstallOptions) (bool, 
 		if err != nil {
 			return false, err
 		}
-		// TODO: remove this once https://github.com/knative/serving/issues/2018 is resolved
-		deleteSingleResource(kc, "horizontalpodautoscaler.autoscaling", "istio-pilot", "")
 	}
 	return true, nil
 }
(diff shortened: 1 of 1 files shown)


ericbottard:
When trying to unsinstall a half-dead system, I get a failure but `kubectl`'s output is not displayed.
I think `deleteAllKnative()` should return an `error` made of the output + `err`

trisberg:
Yeah, the half-dead situation is difficult, I had to re-install riff but at that point the delete of the route CRD took effect and I had to reinstall the route CRD to be able to attempt an uninstall successfully. I'll add the error output to the sysout log.

trisberg:
All I get in that situation is `exit status 1` which isn't that helpful

ericbottard:
This is what I got when I ran the kubectl command manually:

```
riff system uninstall --istio --force
Deleting Knative Services

Error: exit status 1
 ✘  ~  kubectl get service.serving.knative.dev --all-namespaces -ocustom-columns=ns:metadata.namespace,name:metadata.name --no-headers=true
error: the server doesn't have a resource type "service"
```

ericbottard:
Also, even after having re-installed riff properly, I'm still getting the following error:
```
riff system uninstall --istio --force
Deleting Knative Services
In namespace "default"
Error: context deadline exceeded
```

trisberg:
I changed the error logic a bit. If you get the system unstable message try running 
```
riff system install --force
```
twice

pkg/core/system.go
```diff
@@ -411,7 +413,7 @@ func deleteAllKnative(kc *kubectlClient, resourceType string) error {
 func deleteCrds(kc *kubectlClient, suffix string) error {
 	fmt.Printf("Deleting CRDs for %s\n", suffix)
 	crdList, err := kc.kubeCtl.Exec([]string{"get", "customresourcedefinitions","-ocustom-columns=name:metadata.name", "--no-headers=true"})
-	if err != nil {
+	if err == nil {
```
ericbottard: Are you sure about this!?

trisberg: yes, if we deleted one of the CRDs we can list the resources for it, so we get an error, should be able to just move on to the next resource type

trisberg:
I still can't uninstall a zombie system

trisberg:
You might have to force delete any remaining knative resource instances

pkg/core/system.go
ericbottard: I'm not sure I follow. I assume if `err != nil` (the now only case where this code path will stay in this function), then crdList is undefined

trisberg:
This fix doesn't help when uninstalling a zombied system but it should prevent you from getting to that state in the first place.

ericbottard:
Somehow, I'm still getting the timeout error even after re-installing riff:

```
riff system install
Istio is already installed, do you want to install the Knative components for riff? [y/N]: y
Waiting for the Istio components to start . all components are 'Running'

Installing Knative components
Applying resources defined in: https://storage.googleapis.com/knative-nightly/build/previous/v0.3.0/release.yaml
Applying resources defined in: https://storage.googleapis.com/knative-releases/serving/previous/v0.3.0/serving.yaml
Applying resources defined in: https://storage.googleapis.com/knative-releases/eventing/previous/v0.3.0/eventing.yaml
Applying resources defined in: https://storage.googleapis.com/knative-releases/eventing/previous/v0.3.0/in-memory-channel.yaml
Applying resources defined in: https://storage.googleapis.com/projectriff/riff-buildtemplate/riff-cnb-clusterbuildtemplate-0.2.0-snapshot-ci-63cd05079e1f.yaml
Knative components installed


riff system install completed successfully
 ~/Documents/projects/go/src/github.com/projectriff/riff   master  riff system uninstall --force --istio
Deleting Knative Services
In namespace "default"
Error: context deadline exceeded
```

trisberg:
You need to fix the zombie state manually

ericbottard:
I'm not sure how. What could I do manually that repetitive invocations of `riff system uninstall` couldn't do for me? (apart from destroying the cluster). If you think about it, this should eventually work

trisberg:
with a zombie system, try this:

1) list all kservices, configurations, routes
```
kubectl get routes --all-namespaces
kubectl get configurations --all-namespaces
kubectl get kservices --all-namespaces
```
2) reinstall serving twice
```
kubectl apply -f https://storage.googleapis.com/knative-releases/serving/previous/v0.4.0/serving.yaml
kubectl apply -f https://storage.googleapis.com/knative-releases/serving/previous/v0.4.0/serving.yaml
```
3) now check any of the resources in step 1
```
kubectl get routes --all-namespaces
kubectl get configurations --all-namespaces
kubectl get kservices --all-namespaces
```
if you find any, delete them with kubectl

4) uninstall riff
```
riff system uninstall --istio --force
```
(that should hopefully complete)

ericbottard:
Wow, not sure what amazes me the most:
a) the fact that *installing* serving actually *deletes* resources of the CRDs it defines, 
or
b) that you need to install it twice, which with `kubectl apply` should be an idempotent operation

Anyway, this PR makes riff better than it was before in any case. Will merge

ericbottard:
Merged as 8c28e974

trisberg:
I agree, it's strange that this can even happen, would be interesting to track down what changed in serving to make this blocking situation possible. It's a fairly recent change that causes this. Maybe we need to remove the CRDs in a different order? DO they depend on each other somehow?

PR #1117, opened 2019-02-25
Pass configuration overrides to kubectl shellout

### LINKED ISSUE #803: All shell-outs to kubectl need to pass `kubeconfig` and `server` switches if set on riff
The riff cli accepts `--kubeconfig` and `--master` to configure the kubernetes server the command should connect to. However, when we shell out to kubectl, we don't set the values and thus can only communicate with the default cluster, even if the user explicitly asked to connect to another cluster. This is potentially dangerous for an operator depending on the default cluster they have selected.

Note: the `--master` switch in riff is named `--server` in kubectl

fbiville:
Fixes #803

### DIFF
--- a/cmd/commands/wiring.go
+++ b/cmd/commands/wiring.go
@@ -19,6 +19,8 @@ package commands
 import (
 	"context"
 	"fmt"
+	"github.com/projectriff/riff/pkg/fileutils"
+	"github.com/projectriff/riff/pkg/kubectl"
 	"os"
 	"os/user"
 	"strings"
@@ -30,7 +32,6 @@ import (
 	"github.com/projectriff/riff/pkg/core"
 	"github.com/projectriff/riff/pkg/core/kustomize"
 	"github.com/projectriff/riff/pkg/env"
-	"github.com/projectriff/riff/pkg/kubectl"
 	"github.com/spf13/cobra"
 	"k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
@@ -187,16 +188,16 @@ See https://projectriff.io and https://github.com/knative/docs`,
 // reference here and to the command creation helpers) are correctly initialized.
 func installKubeConfigSupport(command *cobra.Command, client *core.Client, kc *core.KubectlClient) {
 
-	kubeconfig := ""
+	kubeConfigPath := ""
 	masterURL := ""
 
-	command.PersistentFlags().StringVar(&kubeconfig, "kubeconfig", "~/.kube/config", "the `path` of a kubeconfig")
+	command.PersistentFlags().StringVar(&kubeConfigPath, "kubeconfig", "~/.kube/config", "the `path` of a kubeconfig")
 	command.PersistentFlags().StringVar(&masterURL, "master", "", "the `address` of the Kubernetes API server; overrides any value in kubeconfig")
 
 	oldPersistentPreRunE := command.PersistentPreRunE
 	command.PersistentPreRunE = func(cmd *cobra.Command, args []string) error {
 		var err error
-		clientConfig, kubeClientSet, eventingClientSet, servingClientSet, err := realClientSetFactory(kubeconfig, masterURL)
+		clientConfig, kubeClientSet, eventingClientSet, servingClientSet, err := realClientSetFactory(kubeConfigPath, masterURL)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


pkg/fileutils/paths.go
```diff
@@ -0,0 +1,18 @@
...
+
+func ResolveTilde(path string) (string, error) {
+	if !strings.HasPrefix(path, "~/") {
+		return path, nil
+	}
+	usr, err := user.Current()
+	if err != nil {
+		return "", err
+	}
+	return filepath.Join(usr.HomeDir, path[2:]), nil
```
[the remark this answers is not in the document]
fbiville: 👍

pkg/fileutils/paths.go
```diff
@@ -0,0 +1,18 @@
+package fileutils
+
+import (
+	"os/user"
+	"path/filepath"
+	"strings"
+)
+
+func ResolveTilde(path string) (string, error) {
+	if !strings.HasPrefix(path, "~/") {
```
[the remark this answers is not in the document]
fbiville: 👍 
We also need to change the default value of the global flag `--kubeconfig`.

pkg/fileutils/paths_test.go
```diff
@@ -0,0 +1,28 @@
...
+		initialPath := "~/some/location"
+
+		path, err := fileutils.ResolveTilde(initialPath)
+
+		Expect(err).NotTo(HaveOccurred())
+		Expect(path).NotTo(ContainSubstring("~"))
+		Expect(path).To(HaveSuffix(initialPath[2:]))
+	})
+
+	It("returns path without tilde as is", func() {
```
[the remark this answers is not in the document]
fbiville: I may be missing something but I just tested`~/~` in Debian and Mac OS and it both resolves to `CURRENT_USER_HOME/~`. `~` in Powershell seems to resolve to the current user home directory as well.

Another case that I discarded maybe too early is the syntax `~another_user/dir` which leads to `ANOTHER_USER_HOME/dir`. Do we want to support this apparently* Unix-only feature?
* gave it a quick run in Powershell/Windows 10, `~anotheruser` is taken literally as a directory called `~anotheruser`.

fbiville:
@glyn Another case that I discarded maybe too early is the syntax `~another_user/dir` which leads to `ANOTHER_USER_HOME/dir`. Do we want to support this apparently* Unix-only feature?
\* gave it a quick run in Powershell/Windows 10, `~anotheruser` is taken literally as a directory called `~anotheruser`.

cmd/commands/wiring.go
```diff
@@ -189,7 +190,7 @@ func installKubeConfigSupport(command *cobra.Command, client *core.Client, kc *c
 	kubeConfigPath := ""
 	masterURL := ""
 
-	command.PersistentFlags().StringVar(&kubeConfigPath, "kubeconfig", "~/.kube/config", "the `path` of a kubeconfig")
+	command.PersistentFlags().StringVar(&kubeConfigPath, "kubeconfig", fmt.Sprintf("~%[1]c.kube%[1]cconfig", filepath.Separator), "the `path` of a kubeconfig")
```
glyn: Actually, this change isn't needed because Windows accepts paths with forward slashes instead of backward slashes. However, when we parse Windows paths, we need to be prepared to accept backslashes which is why we can't parse for strings starting `~/`.

pkg/fileutils/paths_test.go
```diff
@@ -1,14 +1,16 @@
...
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/projectriff/riff/pkg/fileutils"
+	"path/filepath"
 )
 
 var _ = Describe("ResolveTilde", func() {
 	It("resolves ~/ against current user's home directory", func() {
-		initialPath := "~/some/location"
+		initialPath := fmt.Sprintf("~%[1]csome%[1]clocation", filepath.Separator)
```
glyn: Not necessary - see earlier comment.

fbiville:
Paths starting with `~/` and `~\` should be properly resolved on Windows.
Only `~/` is considered for the other OS.

pkg/fileutils/paths_test.go
```diff
@@ -0,0 +1,59 @@
+package fileutils_test
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/projectriff/riff/pkg/fileutils"
+)
+
+var _ = Describe("StartsWithHomeDirAsTilde", func() {
```
glyn: How about adding a test for `returns false when starting with ~\ on macOS`?

fbiville: 👍

pkg/fileutils/paths_test.go
```diff
@@ -14,6 +14,12 @@ var _ = Describe("StartsWithHomeDirAsTilde", func() {
 		Expect(result).To(BeTrue(), "tilde+forward slash should work")
 	})
 
+	It(`returns true when starting with ~\ on Mac OS`, func() {
```
glyn: Do you mean `returns false ...`?

fbiville: 😅 thanks!

fbiville:
Now rebased and ready to be merged.

pkg/fileutils/paths_test.go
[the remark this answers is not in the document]
glyn: Right.

PR #1126, opened 2019-03-04
New go.sum built with Go 1.12

trisberg:


scothis:
If we're going to use go 1.12 we should use it access the board, including with travis.

fbiville:
Merged via 66abf7171591adee0985495e4f8b1e6353a8d629

PR #1128, opened 2019-03-04
Bump nonce annotation for local path function create and update

### LINKED ISSUE #1127: Function update command should trigger new revision even for local path
If I run `riff function update upper --local-path .` on local source directory the configuration isn't modified so no new revision is created. The image is pushed but we need to force a new revision to be created.

trisberg:
- rename `buildAnnotation` to `nonceAnnotation`

Fixes #1127

### DIFF
--- a/pkg/core/annotations.go
+++ b/pkg/core/annotations.go
@@ -29,12 +29,12 @@ func (c *client) bumpNonceAnnotationForRevision(s *v1alpha1.Service) {
 	if annotations == nil {
 		annotations = map[string]string{}
 	}
-	build := annotations[buildAnnotation]
+	build := annotations[nonceAnnotation]
 	i, err := strconv.Atoi(build)
 	if err != nil {
 		i = 0
 	}
-	annotations[buildAnnotation] = strconv.Itoa(i + 1)
+	annotations[nonceAnnotation] = strconv.Itoa(i + 1)
 	s.Spec.RunLatest.Configuration.RevisionTemplate.SetAnnotations(annotations)
 }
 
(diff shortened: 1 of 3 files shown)


pkg/core/function.go
```diff
@@ -550,6 +551,7 @@ func (c *client) UpdateFunction(buildpackBuilder Builder, options UpdateFunction
 		if err != nil {
 			return err
 		}
+		c.bumpNonceAnnotationForRevision(service)
```
ericbottard: Isn't that already taken care of by line 525 (previously 524)?

pkg/core/function.go
```diff
@@ -110,6 +110,7 @@ func (c *client) CreateFunction(buildpackBuilder Builder, options CreateFunction
 				return nil, nil, err
 			}
 		}
+		c.bumpNonceAnnotationForRevision(s)
```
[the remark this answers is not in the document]
trisberg: there are different annotations for builds and revisions, separate objects, we should probably move the `c.bumpNonceAnnotationForBuild(s)` call to the cluster build section

pkg/core/function.go
trisberg: same as previous comment

pkg/core/function.go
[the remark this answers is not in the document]
ericbottard: Huh, sorry I totally misread that.

scothis:
lgtm. Should we add function update to FATS to detect these breakages in the future?

trisberg:
I think it would be nice to add update test to FATS

pkg/core/annotations.go
```diff
@@ -29,12 +29,12 @@ func (c *client) bumpNonceAnnotationForRevision(s *v1alpha1.Service) {
 	if annotations == nil {
 		annotations = map[string]string{}
 	}
-	build := annotations[buildAnnotation]
+	build := annotations[nonceAnnotation]
```
fbiville: Do we want to use more names including `nonce`? I believe @glyn pointed out that this word has a bad connotation in the UK.

glyn: It's not a big problem. The term has been popular in computer security circles, regardless of the dodgy connotations. Just know that any UK readers are likely to wince, or at least smile wryly, as they read this code.

trisberg: I'm fine with changing this to something other than `nonce`, I think we had `build` originally but changed it, @scothis might have some opinion on this

scothis: no strong opinion.

ericbottard:
Merged as 008a528e

PR #1130, opened 2019-03-04
Merge ./pkg/core/{Client,KubectlClient}

### LINKED ISSUE #1129: Merge ./pkg/core/{Client and KubectlClient}
We have two different client interfaces whose methods differ based on the backing technology used to fulfill the request. ./pkg/core/Client uses the programmatic k8s client, while ./pkg/core/KubectlClient shells out to kubectl. In general, we should avoid shelling out to kubectl since we have no ideas what version is installed and instead take responsibility for managing resource ourselves.

This split also make it harder to reuse methods defined on one client from methods on the other client. By merging the two clients, we can define a single public interface and individual methods can use the implementation they need.

scothis:
This eliminates the technical distinction between the two clients and
provides a uniform interface for callers. We will likely want to
decompose this interface in the future into domain focused interfaces,
decoupling from the underpinning implementation is a good first step.

Fixes #1129


go.sum
```diff
@@ -236,6 +236,7 @@ github.com/hashicorp/yamux v0.0.0-20181012175058-2f1d1f20f75d h1:kJCB4vdITiW1eC1
 github.com/hashicorp/yamux v0.0.0-20181012175058-2f1d1f20f75d/go.mod h1:+NfK9FKeTrX5uv1uIXGdwYDTeHna2qgaIlx54MXqjAM=
 github.com/hpcloud/tail v1.0.0 h1:nfCOvKYfkgYP8hkirhJocXT2+zOD8yUNjXaWfTlyFKI=
 github.com/hpcloud/tail v1.0.0/go.mod h1:ab1qPbhIpdTxEkNHXyeSf5vhxWSCs/tWer42PpOxQnU=
+github.com/ijc/Gotty v0.0.0-20170406111628-a8b993ba6abd h1:anPrsicrIi2ColgWTVPk+TrN42hJIWlfPHSBP9S0ZkM=
```
fbiville: there should not be any changes here, what happened?

pkg/core/client.go
```diff
@@ -62,8 +70,17 @@ type client struct {
...
 	serving      serving_cs.Interface
 	clientConfig clientcmd.ClientConfig
+	kubeCtl      kubectl.KubeCtl
+	kustomizer   kustomize.Kustomizer
 }
 
-func NewClient(clientConfig clientcmd.ClientConfig, kubeClient kubernetes.Interface, eventing eventing_cs.Interface, serving serving_cs.Interface) Client {
-	return &client{clientConfig: clientConfig, kubeClient: kubeClient, eventing: eventing, serving: serving}
+func NewClient(clientConfig clientcmd.ClientConfig, kubeClient kubernetes.Interface, eventing eventing_cs.Interface, serving serving_cs.Interface, kubeCtl kubectl.KubeCtl, kustomizer kustomize.Kustomizer) Client {
+	return &client{
```
fbiville: maybe this PR could be an opportunity to rename `client` to a more descriptive, less vague name?

go.sum
fbiville: nvm, got the same behavior locally with another PR. I guess that's the effect of using Go 1.12.

scothis: `make` adds this line, `go mod tidy` with 1.11.5 removes it. I hope go 1.12 stabilizes things.

pkg/core/client.go
scothis: Ultimately, I think we should decompose 'client' into a set of domain specific clients. I don't want to do that right now though.

fbiville: 👍

PR #1132, opened 2019-03-05
Always lookup builder/run images from cluster for off-cluster builds

scothis:
Previously, the pack builder and run images were passed as configuration from the main into the wiring of the root command. This enabled other vendors to provide custom images.

In reality, the images used for local pack builds should be the same as are used for on-cluster builds. Now that we use a ClusterBuildTemplate that is parameterized with these images, we can lookup the images from a central location in a canonical way. Now the pack builder and run images are always defined by the riff-build ClusterBuildTemplate. This is true for both on- and off-cluster builds.

One downside is that this change requires a k8s cluster to have a riff install before riff can be used to build a function. While this is a small loss of functionality, the benefit of consistency outweigh the loss.


pkg/core/function.go
```diff
@@ -534,19 +530,17 @@ func (c *client) UpdateFunction(buildpackBuilder Builder, options UpdateFunction
 	if build == nil {
 		// function was built locally, attempt to reconstruct configuration
 		localBuild := BuildOptions{
-			RunImage:       annotations[buildpackRunImageAnnotation],
-			BuildpackImage: annotations[buildpackBuildImageAnnotation],
-			LocalPath:      appDir,
-			Artifact:       annotations[functionArtifactAnnotation],
-			Handler:        annotations[functionHandlerAnnotation],
-			Invoker:        annotations[functionOverrideAnnotation],
+			LocalPath: appDir,
```
ericbottard: Don't we still want to save which actual image was used?

pkg/core/helpers.go
```diff
@@ -23,13 +23,36 @@ import (
...
 	DefaultImagePrefixKey = "default-image-prefix"
 )
 
+type PackConfig struct {
+	BuilderImage string
+	RunImage     string
+}
+
+func (c *client) FetchPackConfig() (*PackConfig, error) {
+	template, err := c.build.BuildV1alpha1().ClusterBuildTemplates().Get("riff-cnb", v1.GetOptions{})
```
ericbottard: The 3 strings used in this function should be defined as constants I think

pkg/core/helpers.go
```diff
@@ -23,13 +23,36 @@ import (
...
+	if err != nil {
+		return nil, err
+	}
+	config := &PackConfig{}
+	for _, param := range template.Spec.Parameters {
+		switch param.Name {
```
ericbottard: Should this fail if we did not find those 3 keys?

pkg/core/function.go
scothis: I don't think so, that would add drift between the current images and the images when the function was initially created. We don't save those images for on-cluster builds.

If we ever end up with multiple builders then we'll need some way to pick which one to use, but we don't have that problem today.

ericbottard: Fair point

PR #1134, opened 2019-03-05
riff function build

### LINKED ISSUE #972: Allow function builds without creating a Knative service

scothis:
Build a riff function from source producing an container image. This
command currently only operates against local source, however, that may
change in the future. Currently, this requires a Docker daemon and does
not use a Kubernetes cluster.

`riff function build` is equivlent to `riff function create` without
creating a Knative Service to run the built image.

This is the build half of #1080 separated from run and moved directly under the function command.

Fixes #972


cmd/commands/function_test.go
```diff
@@ -372,3 +372,99 @@ var _ = Describe("The riff function update command", func() {
...
+			Expect(err).To(MatchError(ContainSubstring("accepts 0 arg(s), received 1")))
+		})
+		It("should fail without required flags", func() {
+			fc.SetArgs([]string{})
+			err := fc.Execute()
+			Expect(err).To(MatchError(ContainSubstring("required flag(s)")))
+			Expect(err).To(MatchError(ContainSubstring("image")))
+			Expect(err).To(MatchError(ContainSubstring("local-path")))
+		})
+	})
```
fbiville: it would be good to add a case where defaults are empty

cmd/commands/function.go
```diff
@@ -184,3 +188,43 @@ func FunctionUpdate(buildpackBuilder core.Builder, fcTool *core.Client) *cobra.C
...
+	buildFunctionOptions := core.BuildFunctionOptions{
+		BuildOptions: core.BuildOptions{
+			BuildpackImage: defaults.BuilderImage,
+			RunImage:       defaults.RunImage,
+		},
+	}
+
+	command := &cobra.Command{
+		Use:   "build",
+		Short: "Build a function container from local source",
```
fbiville: I think it would be helpful to add a long description with examples
it still takes me some time to adjust to `--artifact` and `--handler`,  I guess it could be the case to first-time users as well

pkg/core/function.go
```diff
@@ -589,18 +589,34 @@ func (c *client) UpdateFunction(buildpackBuilder Builder, options UpdateFunction
 	return nil
 }
 
+type BuildFunctionOptions struct {
+	BuildOptions
+
+	Image string
+}
+
+func (c *client) BuildFunction(buildpackBuilder Builder, options BuildFunctionOptions, log io.Writer) error {
```
fbiville: it seems the log param is unused

pkg/core/function.go
```diff
@@ -589,18 +589,34 @@ func (c *client) UpdateFunction(buildpackBuilder Builder, options UpdateFunction
...
+}
+
+func doLocally(options BuildOptions, doer func() error) error {
 	if err := writeRiffToml(options); err != nil {
 		return err
 	}
 	defer func() { _ = deleteRiffToml(options) }()
 	if options.BuildpackImage == "" {
-		return fmt.Errorf("unable to build function locally: buildpack image not specified")
+		return fmt.Errorf("unable to do locally: buildpack image not specified")
```
fbiville: `"unable to do locally"` does not seem super clear
what about inlining this function and get back to the previous message: `"unable to build function locally[...]"`

cmd/commands/function.go
scothis: I'll centralize the flags between create and build, and we can refine the help in a future PR.

fbiville: 👍

PR #1137, opened 2019-03-06
Add option to prevent service from scaling to zero instances

### LINKED ISSUE #1136: Provide an option for services to opt out of scale to 0
For services that act as a polling source without being triggered by a http request it could be useful to allow them to opt out of getting scaled to 0.

trisberg:
- add `--no-scale-to-zero` flag for `service create`

- all revisions will keep running until service is deleted unless manually removed

Resolves #1136

### DIFF
--- a/cmd/commands/service.go
+++ b/cmd/commands/service.go
@@ -122,6 +122,8 @@ func ServiceCreate(fcTool *core.Client) *cobra.Command {
 	command.Flags().StringVar(&createServiceOptions.Image, "image", "", "the `name[:tag]` reference of an image containing the application/function")
 	command.MarkFlagRequired("image")
 
+	command.Flags().BoolVarP(&createServiceOptions.NoScaleToZero, "no-scale-to-zero", "", false, "the service should not scale to zero instances")
+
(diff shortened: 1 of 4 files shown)


trisberg:
rebased

markfisher:
I do have mixed feelings about this. I assume it's mostly going to be requested due to cold-start latency, which ideally would be improved to the point of being suitable for all but extreme cases. For those cases, as Scott mentions maybe a `Deployment` is a better fit anyways.

All that said, if we do add this feature, should we just expose `--min-scale` instead?

trisberg:
I initially exposed `--min-scale` but changed to only allow 1 since the revisions don't automatically scale down and there could be a lot of orphaned pods running. I think that is a bug in Knative serving, when the revision is no longer routable, then it should just scale to 0 regardless of this annotation.

trisberg:
It is possible to do this with a patch command like:

```
kubectl patch -n demo kservice petclinic --type=json -p='[{"op": "add", "path": "/spec/runLatest/configuration/revisionTemplate/metadata/annotations", "value": {"autoscaling.knative.dev/minScale": "1"}}]'
```

so maybe we should not merge this until the revision scaling is better behaved

PR #1138, opened 2019-03-06
Allow builds from directories within git repos

### LINKED ISSUE #1089: Multiple functions in a single repo
It's sometimes advantageous to store more than one function in the same git repository. Currently, riff doesn't support this case, however, Knative Build does with the SubPath field on the SourceSpec.

We can add a `--sub-path` flag to `riff function create`. When doing an in-cluster build we should set `build.Spec.SubPath` with the value. When doing a local build the flag could either be ignored, forbidden or appended onto the value of `--local-path` depending on what produces the clearest command help text and test cases.

scothis:
Builds can now be triggered from directories nested in a git repo. This
means that a single repo can host multiple functions. Only files within
the path will exist at build time.

```
riff function create <name> --git-repo <repo> --sub-path path/within/repo
```

Fixes #1089


ericbottard:
👍

PR #1148, opened 2019-03-11
Give users more control of --image while using the default prefix

scothis:
Previously, using the default image prefix was all-or-nothing. Now, we
can use the default value while also providing a custom repository path
under the prefix or a custom name within the repository.

- `--image _/custom-name` replaces the `_` with the prefix value
- `--image _:custom-tag` replace the `_` with the current prefix value
  combined with the function name (the current default if --image is not
  set)
- `--image _/custom-name:custom-tag` replaces the `_` with the prefix value

Refs #1105


markfisher:
I like the simplicity of this implementation, but I'm wondering if we would have a need to support also changing the repo name and the tag but not the image name, which would require something more like: `my.repo/_:0.1.2` in which case we might want to require this for just the tag: `_/_:0.1.2`

scothis:
> I'm wondering if we would have a need to support also changing the repo name and the tag but not the image name, which would require something more like: `my.repo/_:0.1.2` in which case we might want to require this for just the tag: `_/_:0.1.2`

I'm leaning on a leading underscore being an invalid repository to never conflict with a valid image name. If we move the underscore into the middle of the repository then I'm not sure if it will be valid, we'd need to make sure we don't clobber a valid image name. 

My main goal for adding a default for --image was to avoid needing to specify the registry every time. I don't think we should get too fancy until we have user demand for something more.

cmd/commands/function.go
```diff
@@ -65,16 +66,31 @@ func FunctionCreate(buildpackBuilder core.Builder, fcTool *core.Client) *cobra.C
...
+				}
+
+				fnName := args[functionCreateFunctionNameIndex]
+				if createFunctionOptions.Image == "" {
+					// combine prefix and function name to provide default image
+					createFunctionOptions.Image = fmt.Sprintf("%s/%s", prefix, fnName)
+				} else if strings.HasPrefix(createFunctionOptions.Image, "_:") {
+					// add the prefix to the specified image tag
+					createFunctionOptions.Image = strings.Replace(createFunctionOptions.Image, "_", fmt.Sprintf("%s/%s", prefix, fnName), 1)
+				} else if strings.HasPrefix(createFunctionOptions.Image, "_/") {
```
fbiville: It makes sense but I find the overloading of `_` a bit confusing at first.
It would be nice to be able to use explicit variables instead.

Here is a short example of what one could do (based on Golang [`text/template`](https://golang.org/pkg/text/template/) package):
```
imageTemplate, err := template.New("image").Parse(createFunctionOptions.Image)
if err != nil {
	return err
}
builder := &strings.Builder{}
err = imageTemplate.Execute(builder, map[string]string{
	"ImagePrefix": prefix,
	"FunctionName":    fnName,
})
if err != nil {
	return err
}
createFunctionOptions.Image = builder.String()
```

The default value for the image option could become: `{{.ImagePrefix}}/{{.FunctionName}}` and users could reference either of the variables (and arguably shoot themselves in the foot).

scothis: I think a template string is generally too complex for users at a terminal and doesn't add enough value for scripts that can do their own substitution. If others have strong feelings, I'm willing to be convinced.

My first though was to do something more like `_prefix/foo` where `_prefix` is substituted, but then I released that 'prefix' isn't adding any value, so I removed it.

fbiville: Even without going to a full-fledged template string, I'd tend to use two different symbols instead of just one. One symbol would be for the prefix, one for the prefix+function. WDYT?

ericbottard:
I agree that we should stay away from complicated, possibly shell-reserved
characters.
No particular opinion about using two different characters, as long as they
are forbidden in regular image names

Le mer. 13 mars 2019 à 15:58, Florent Biville <notifications@github.com> a
écrit :

cmd/commands/function.go
scothis: I think adding two different symbols make the replacement more arcane and increases cognitive load as you have to think about which symbol to use. Using a single symbol to indicate replacement and using the secondary symbol (`/` or `:`) as part of the image in its natural form.

As is, the leading `_` is replaced with a value that will make the rest of the image valid. If we think the behavior is confusing, I'm open to dropping the `_:` form, re-typing the function name shouldn't add very much of a cognitive load, it was also a late addition once I saw how easy it would be to implement.

scothis:
Mark had the idea of supporting multiple image prefix values if/when we have multiple credentials to different registries. In that case we could use the name of the credentials after the underscore to indicate which prefix to apply.

For example, with GCR and DockerHub configured within the same namespace.

- `_` default prefix (equivalent to not specifying `--image`)
- `_gcr`
- `_gcr/custom-name`
- `_hub`
- `_hub:custom-tag`

`gcr` and `hub` are named defined by the user when adding the credentials.

cmd/commands/function.go
fbiville: 👍

pkg/riff/commands/function.go
```diff
@@ -65,16 +66,28 @@ func FunctionCreate(buildpackBuilder core.Builder, fcTool *core.Client) *cobra.C
 				return err
 			}
 
-			if createFunctionOptions.Image == "" {
+			if createFunctionOptions.Image == "" || strings.HasPrefix(createFunctionOptions.Image, "_") {
```
fbiville: small nit, now that there is no `_:`
```suggestion
			if createFunctionOptions.Image == "" || strings.HasPrefix(createFunctionOptions.Image, "_/") {
```

scothis: I'd rather keep it as is

PR #1153, opened 2019-03-13
Update headers

fbiville:


fbiville:
Shall I wait for other tasks to be complete before merging that one @glyn?

glyn:
> Shall I wait for other 0.3.0 tasks to be complete before merging that one @glyn?

No, I don't think so. Any clashes you cause will be trivial to resolve and it's important to keep the copyright dates correct.

However, I guess you might want to re-run headache with the recent fix and update this PR first.

Issue #1155, opened 2019-03-14
Build fails on Windows


trisberg:
Latest master fails with this:

```
$ make build
warning: LF will be replaced by CRLF in go.mod.
The file will have its original line endings in your working directory
warning: LF will be replaced by CRLF in go.sum.
The file will have its original line endings in your working directory
GO111MODULE=on go build -o ./riff -ldflags "-X github.com/projectriff/riff/pkg/env.cli_name=riff -X github.com/projectriff/riff/pkg/env.cli_version=0.3.0-snapshot -X github.com/projectriff/riff/pkg/env.cli_gitsha=025c79d36c27689dc5ea0d988a521ceae6b0f4ce -X github.com/projectriff/riff/pkg/env.cli_gitdirty="
can't load package: package github.com/projectriff/riff:
main.go:1:1: expected 'package', found cmd
make: *** [Makefile:52: riff] Error 1
```

fbiville:
Is it because `main.go` is now a symlink?

trisberg:
That's my hunch.

I'm using `go version go1.11.4 windows/amd64 `

scothis:
Proposed fix in https://github.com/projectriff/riff/pull/1157, but I have not tested on Windows

PR #1157, opened 2019-03-14
Use proper main location for building riff

scothis:
./main.go is a symlink to ./cmd/riff/main.go, it's intended as a fall
back, but not the main entry point. We should use the proper entry point
when building riff.

Refs #1155


ericbottard:
LGTM

PR #1173, opened 2019-03-19
Modify the way files paths/URLs/schemes are handled

### LINKED ISSUE #1036: riff system install fails to install using manifest from local directory on Windows

trisberg:
- on Windows using net/url.Parse with an absolute file path without a scheme but with a drive letter will cause the URL to report the drive letter as the scheme

Fixes #1036

### DIFF
--- a/pkg/fileutils/read.go
+++ b/pkg/fileutils/read.go
@@ -38,17 +38,21 @@ func Read(file string, base string) ([]byte, error) {
 }
 
 func readAbsFile(file string) ([]byte, error) {
+	// The file argument could be an abs file reference or a URL
 	u, err := url.Parse(file)
 	if err != nil {
 		return nil, err
 	}
-	if u.Scheme == "" {
+	if u.Scheme == "http" || u.Scheme == "https" || u.Scheme == "file" {
+		return ReadUrl(u, 0)
+	}
+	if EmptyScheme(u, runtime.GOOS) {
 		if !filepath.IsAbs(file) {
 			return nil, fmt.Errorf("absolute path expected instead of relative path: %s", file)
 		}
 		return ioutil.ReadFile(file)
 	}
-	return ReadUrl(u, 0)
+	return nil, fmt.Errorf("invalid scheme '%s' for file path: %s", u.Scheme, file)
 }
 
 func ReadUrl(url *url.URL, httpTimeout time.Duration) ([]byte, error) {
@@ -93,4 +97,12 @@ func downloadFile(url string, timeout time.Duration) ([]byte, error) {
 		return nil, err
 	}
 	return buf.Bytes(), nil
+}
+
+func EmptyScheme(u *url.URL, os string) bool {
+	// On Windows the URL potentially ends up with the drive letter as the scheme
+	if u.Scheme == "" || (os == "windows" && len(u.Scheme) == 1) {
+		return true
+	}
+	return false
 }
--- a/pkg/fileutils/read_test.go
+++ b/pkg/fileutils/read_test.go
@@ -201,6 +201,135 @@ var _ = Describe("ReadUrl", func() {
 	})
 })
 
+var _ = Describe("EmptyScheme", func() {
+
+	var (
+		os      string
+		file	string
+		u       *url.URL
+		err     error
+		empty   bool
+	)
+
+	JustBeforeEach(func() {
+		u, err = url.Parse(file)
+		empty = fileutils.EmptyScheme(u, os)
+	})
+
+	Context("on Windows", func() {
+
+		BeforeEach(func() {
+			os = "windows"
+		})
+
+		Context("when file: is the scheme and using unix separators", func() {
+			BeforeEach(func() {
+				file = "file:///my/file"
+			})
+
+			It("scheme should not be empty", func() {
+				Expect(err).NotTo(HaveOccurred())
+				Expect(empty).To(BeFalse())
+			})
+		})
+
+		Context("when no scheme and using windows separators and no drive letter", func() {
+			BeforeEach(func() {
+				file = "\\my\\file"
+			})
+
+			It("scheme should be empty", func() {
+				Expect(err).NotTo(HaveOccurred())
+				Expect(empty).To(BeTrue())
+			})
+		})
+
+		Context("when no scheme and using windows separators and drive letter", func() {
+			BeforeEach(func() {
+				file = "C:\\my\\file"
+			})
+
+			It("scheme should be empty", func() {
+				Expect(err).NotTo(HaveOccurred())
+				Expect(empty).To(BeTrue())
+			})
+		})
+
+		Context("when http: is the scheme", func() {
+			BeforeEach(func() {
+				file = "http://foo.com/bar"
+			})
+
+			It("scheme should not be empty", func() {
+				Expect(err).NotTo(HaveOccurred())
+				Expect(empty).To(BeFalse())
+			})
+		})
+
+		Context("when https: is the scheme", func() {
+			BeforeEach(func() {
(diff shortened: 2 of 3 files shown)


pkg/fileutils/read.go
```diff
@@ -38,17 +39,22 @@ func Read(file string, base string) ([]byte, error) {
...
 	u, err := url.Parse(file)
 	if err != nil {
 		return nil, err
 	}
-	if u.Scheme == "" {
+	if u.Scheme == "http" || u.Scheme == "https" || u.Scheme == "file" {
+		return ReadUrl(u, 0)
+	}
+	// On Windows the URL potentially ends up with the drive letter as the scheme so compare with the URL as a string
+	if u.Scheme == "" || (runtime.GOOS == "windows" && len(u.Scheme) == 1 && strings.HasPrefix(u.String(), u.Scheme + ":")) {
```
fbiville: Nit: maybe extract the Windows-specific check in a small aptly-named function?

I was initially tempted to check that `u.Scheme` belongs to [a-z], but today I learnt that Windows XP (and later versions) supports mounting non-letter drives ([source](https://en.wikipedia.org/wiki/Drive_letter_assignment#Common_assignments)).

pkg/fileutils/read.go
```diff
@@ -38,17 +39,21 @@ func Read(file string, base string) ([]byte, error) {
...
+	// The file argument could be an abs file reference or a URL
 	u, err := url.Parse(file)
 	if err != nil {
 		return nil, err
 	}
-	if u.Scheme == "" {
+	if u.Scheme == "http" || u.Scheme == "https" || u.Scheme == "file" {
+		return ReadUrl(u, 0)
+	}
+	if u.Scheme == "" || onWindowsSchemeIsDriveLetter(u) {
```
glyn: I think this disjunction should be factored out into a helper function called something like `emptyScheme` because that's really what we are looking for. That will also make the caller more self-documenting.

pkg/fileutils/read.go
```diff
@@ -94,3 +99,12 @@ func downloadFile(url string, timeout time.Duration) ([]byte, error) {
 	}
 	return buf.Bytes(), nil
 }
+
+func onWindowsSchemeIsDriveLetter(u *url.URL) bool {
+	// On Windows the URL potentially ends up with the drive letter as the scheme so compare with the URL as a string
+	if runtime.GOOS == "windows" && len(u.Scheme) == 1 && strings.HasPrefix(u.String(), u.Scheme + ":") {
```
glyn: Doesn't `len(u.Scheme) == 1` _imply_ that `strings.HasPrefix(u.String(), u.Scheme + ":")`? If not, we need some testcases which show why this extra conjunct is needed.

Actually, more tests in this area, to cover the Windows novelties are probably advisable anyway, like Florian said.

ericbottard: Even, isn't `strings.HasPrefix(u.String(), u.Scheme + ":")` ALWAYS true? This is just testing that a URL starts with its scheme followed by a colon. I don't think there is a 100% sure way to test this (there could always be a fancy scheme named "d"), but testing Windows + 1 letter scheme seems to be the best we can do for now

glyn:
@trisberg BTW I really appreciate you coming to help with the delicate area of URL and file path handling in Go. The more eyes on this code, the better!

pkg/fileutils/read.go
trisberg: Looking at the whole chain the manifest file reference goes through, it seems we massage the path and strip out the scheme from the file reference, then we parse it into a URL again in this function, so now something like `C:/foo/bar.yaml` gets parsed as a scheme of `c` and a path of `/foo/bar.yaml` and a file like `C:\foo\bar.yaml` gets a scheme of `c` and and empty path. Lots of weird issues but this fix fixes some. And, yeah, we probably don't need the prefix test as long as the scheme is 1 character.

pkg/fileutils/read_test.go
```diff
@@ -204,67 +204,76 @@ var _ = Describe("ReadUrl", func() {
...
 	var (
+		file	string
 		u       *url.URL
+		os      string
 		err     error
 		empty   bool
 	)
 
+	BeforeEach(func() {
+		os = "linux"
```
glyn: It would be good to put this in a "linux" Context and have a corresponding "windows" context. Then it will be easier to compare which tests are run for which OS's.

pkg/fileutils/read_test.go
```diff
@@ -275,7 +284,7 @@ var _ = Describe("EmptyScheme", func() {
 
 		It("scheme should not be empty", func() {
 			Expect(err).NotTo(HaveOccurred())
-			Expect(!empty)
+			Expect(empty).NotTo(BeTrue())
```
glyn: `Expect(empty).To(BeFalse())` is easier to read.

pkg/fileutils/read_test.go
```diff
@@ -204,67 +204,76 @@ var _ = Describe("ReadUrl", func() {
...
 	Context("when http: is the scheme", func() {
 		BeforeEach(func() {
-			u, err = url.Parse("http://foo.com/bar")
+			file = "http://foo.com/bar"
 		})
 
 		It("scheme should not be empty", func() {
 			Expect(err).NotTo(HaveOccurred())
-			Expect(!empty)
+			Expect(empty).NotTo(BeTrue())
```
glyn: `Expect(empty).To(BeFalse())` is easier to read.

pkg/fileutils/read_test.go
```diff
@@ -204,67 +204,76 @@ var _ = Describe("ReadUrl", func() {
...
 	Context("when file: is the scheme and using unix separators", func() {
 		BeforeEach(func() {
-			u, err = url.Parse("file:///my/file")
+			file = "file:///my/file"
 		})
 
 		It("scheme should not be empty", func() {
 			Expect(err).NotTo(HaveOccurred())
-			Expect(!empty)
+			Expect(empty).NotTo(BeTrue())
```
glyn: `Expect(empty).To(BeFalse())` is easier to read.

pkg/fileutils/read_test.go
```diff
@@ -254,10 +248,57 @@ var _ = Describe("EmptyScheme", func() {
...
+		})
+
+		It("scheme should not be empty", func() {
+			Expect(err).NotTo(HaveOccurred())
+			Expect(empty).To(BeFalse())
+		})
+	})
+})
+
+var _ = Describe("EmptyScheme on Unix-style OS", func() {
```
glyn: This should be part of a single "EmptyScheme" Describe block with two Context blocks for macOS and linux, respectively. That will stop the two Describe blocks "parting company".

trisberg:
rebased and squashed

Issue #1192, opened 2019-03-26
Allow, but do not require riff.toml


scothis:
We currently use riff.toml to pass configuration from the cli to the function buildpacks about the function that should be built. This includes the artifact, handler and invoker override. The presence of the riff.toml file is what indicates to the buildpack that the source is for a function and not an application.

The riff.toml file is created by the build template for cluster builds and by the riff cli for local builds. If there is an existing riff.toml, it is either overwritten or the build is terminated.

We should allow users to commit a riff.toml into a source repo as an option. The user should also be able to override the values stored in riff.toml by runtime configuration values passed in the CLI.

pack is now able to receive individual environmental variables that are exposed to the buildpacks. We can use the environment in addition to the values in riff.toml (if any) to configure the build. Detecting a function build can be triggered by either the presence of riff.toml or a special environment variable.

- riff.toml -> `RIFF=true`
- `artifact` -> `RIFF_ARTIFACT`
- `handler` -> `RIFF_HANDLER`
- `override` -> `RIFF_OVERRIDE`

This change will require:
- [x] updating the Metadata constructor in riff-buildpack (https://github.com/projectriff/riff-buildpack/pull/55)
- [x] update all function buildpacks with the new riff-buildpack
  - [x] java (https://github.com/projectriff/java-function-buildpack/pull/2)
  - [x] node (https://github.com/projectriff/node-function-buildpack/pull/3)
  - [x] command (https://github.com/projectriff/command-function-buildpack/pull/3)
- [x] update the riff-buildpack-group (https://github.com/projectriff/riff-buildpack-group/pull/42)
  - [x] consume each updated buildpack
  - [x] update build template to not create riff.toml but pass values as envvars to detect and build containers
- [x] update riff cli (https://github.com/projectriff/riff/pull/1195)
  - [x] import latest pack
  - [x] no longer create riff.toml for local builds, passing values as env vars to pack
- [x] update FATS (https://github.com/projectriff/fats/pull/128)
  - [x] replace `.fats/create` config with a riff.toml for each function

scothis:
9 PRs later, we're done

PR #1195, opened 2019-03-26
Allow riff.toml to exist

scothis:
Function authors may define and commit their own riff.toml file which
will be used for both local and on-cluster builds. The riff cli allows
users to override values in the riff.toml when creating a function.

packs now allows environment variables to be passed to the detect and
build phases. The riff cli no longer writes its own riff.toml instead
defering to the user's riff.toml and instead passes it's custom config
as env vars. Function buildpacks understand both riff.toml and the env
var, favoring the environment when there is a conflict.

Refs #1192

### DIFF
--- a/pkg/core/client.go
+++ b/pkg/core/client.go
@@ -65,7 +65,7 @@ type Client interface {
 }
 
 type Builder interface {
-	Build(appDir, buildImage, runImage, repoName string, log io.Writer) error
+	Build(repoName string, options BuildOptions, log io.Writer) error
 }
 
 type client struct {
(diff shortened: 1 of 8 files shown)


scothis:
Note: pack does not currently expose the ability to set env vars via a clean API. We're duplicating the current content of `github.com/buildpack/pack.Build` inside riff, augmenting it to accept env vars. Once pack offers a proper api we will switch.

pkg/core/client.go
```diff
@@ -65,7 +65,7 @@ type Client interface {
 }
 
 type Builder interface {
-	Build(appDir, buildImage, runImage, repoName string, log io.Writer) error
+	Build(repoName string, options BuildOptions, log io.Writer) error
```
ericbottard: Builder is used to abstract and easily mock the capability of building. It should not know about the command/service type `BuildOptions` IMO

scothis: The Builder interface is going away once we have a properly pack client API to consume. I'm not worried about it.

PR #1200, opened 2019-03-28
Add contributor covenant

scothis:
We're big enough that we should have a contributor code of conduct. It's better to have this in place before something happens.

This is a direct lift from spring. We may want to adjust the reporting email address to be something riff specific down the road, but should be good enough for now.


jldec:
Thanks for this @scothis

PR #1202, opened 2019-03-28
Apply the Istio patch to istio.yaml

scothis:
The istio.yaml published as part of the Knative Serving release is too
heavy in both the number of replicas and reservations. We previously
scaled it back using a "patch" yaml that duplicates the resources that
needed to change. Now we use a single yaml file with the patch applied.

The diff from the Knative version is also published.

### DIFF
--- a/cmd/riff/main.go
+++ b/cmd/riff/main.go
@@ -30,8 +30,8 @@ var (
 		"stable": {
 			ManifestVersion: "0.1",
 			Istio: []string{
-				"https://storage.googleapis.com/knative-releases/serving/previous/v0.4.1/istio.yaml",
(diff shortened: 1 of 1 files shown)


fbiville:
Could we use kustomize to patch the istio resource or is the approach too fragile?

scothis:
@glyn the rational is threefold:
1. applying one file is more efficient than applying two files for the same result. Within k8s we no longer mutate a resource that was just created
2. we can avoid images published using the registry v1 api as their digests will mutate in certain conditions making the digest not very useful as a digest.
3. the "patch" file is actually more work to maintain since it must contain a full copy of the resource being modified. We needed to make sure that any other changes are also applied in the patch file.

@fbiville you're welcome to try using kustomize to generate the file.

PR #1204, opened 2019-04-01
Print built image name at the end of function create

### LINKED ISSUE #1201: Make image name easier to figure out after build
Since #1105 developers can call `function create` without specifying `--image`

Developers should have an easy way to see the actual image name produced from the `--verbose` output of the `function create`, and maybe even from the non-verbose output or perhaps from a variant of `service status` or `service list`.

As a starting point, I suggest adding the image name to the last line of `function create` e.g.

```
riff function create: built image xxxx/xx for service xx. 
```

ericbottard:
Fixes #1201

### DIFF
--- a/pkg/riff/commands/function.go
+++ b/pkg/riff/commands/function.go
@@ -114,7 +114,9 @@ func FunctionCreate(buildpackBuilder core.Builder, fcTool *core.Client) *cobra.C
 					return err
 				}
 			} else {
+				_, _ = fmt.Fprintf(cmd.OutOrStdout(), "\nriff function create: built image %s for service %s.\n", f.Spec.RunLatest.Configuration.RevisionTemplate.Spec.Container.Image, f.Name)
 				PrintSuccessfulCompletion(cmd)
+
 				if !createFunctionOptions.Verbose && !createFunctionOptions.Wait {
 					namespaceOption := ""
 					if createFunctionOptions.Namespace != "" {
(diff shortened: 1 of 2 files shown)


pkg/riff/commands/function_test.go
```diff
@@ -462,3 +462,21 @@ var _ = Describe("The riff function build command", func() {
...
+							Container: corev1.Container{
+								Image: img,
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
```
fbiville: nice

fbiville:
Merged in 022e09db1895c8d99ec5bce052560a50b11f3cac

pkg/riff/commands/function.go
```diff
@@ -114,7 +114,9 @@ func FunctionCreate(buildpackBuilder core.Builder, fcTool *core.Client) *cobra.C
 					return err
 				}
 			} else {
+				_, _ = fmt.Fprintf(cmd.OutOrStdout(), "\nriff function create: built image %s for service %s.\n", f.Spec.RunLatest.Configuration.RevisionTemplate.Spec.Container.Image, f.Name)
```
trisberg: should we really hardcode `riff` here?

fbiville: oops, merged too hastily. A fix is here: https://github.com/projectriff/riff/pull/1206.

ericbottard: Good catch @trisberg !

PR #1205, opened 2019-04-01
 Delete existing image prefix when none set

### LINKED ISSUE #1199: We should reset the 'default-image-prefix' when using '--secret' without '--image-prefix'
If we don't do this we can end up with a `build-credentials` secret that do not match the `default-image-prefix`

fbiville:
Fixes #1199.


trisberg:
The CI passes now so I'll merge this and will open an additional PR to remove the default-image-prefix

PR #1206, opened 2019-04-01
Do not hardcode CLI name

### LINKED ISSUE #1201: Make image name easier to figure out after build

fbiville:
Fixes #1201

### DIFF
--- a/pkg/riff/commands/function.go
+++ b/pkg/riff/commands/function.go
@@ -114,7 +114,7 @@ func FunctionCreate(buildpackBuilder core.Builder, fcTool *core.Client) *cobra.C
 					return err
 				}
 			} else {
-				_, _ = fmt.Fprintf(cmd.OutOrStdout(), "\nriff function create: built image %s for service %s.\n", f.Spec.RunLatest.Configuration.RevisionTemplate.Spec.Container.Image, f.Name)
+				_, _ = fmt.Fprintf(cmd.OutOrStdout(), "\n%s function create: built image %s for service %s.\n", env.Cli.Name, f.Spec.RunLatest.Configuration.RevisionTemplate.Spec.Container.Image, f.Name)
 				PrintSuccessfulCompletion(cmd)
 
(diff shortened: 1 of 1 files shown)


pkg/riff/commands/function.go
```diff
@@ -114,7 +114,7 @@ func FunctionCreate(buildpackBuilder core.Builder, fcTool *core.Client) *cobra.C
 					return err
 				}
 			} else {
-				_, _ = fmt.Fprintf(cmd.OutOrStdout(), "\nriff function create: built image %s for service %s.\n", f.Spec.RunLatest.Configuration.RevisionTemplate.Spec.Container.Image, f.Name)
+				_, _ = fmt.Fprintf(cmd.OutOrStdout(), "\n%s function create: built image %s for service %s.\n", env.Cli.Name, f.Spec.RunLatest.Configuration.RevisionTemplate.Spec.Container.Image, f.Name)
```
sbawaska: <nit>
Since this is going to be printed while running the `riff function create` command, I don't think the prefix is needed.
</nit>

ericbottard:
I'm not sure I understand why we did not derive the cli name from `os.Args[0]` btw...

pkg/riff/commands/function.go
scothis: Agreed. We should only include the CLI command when we are giving the user a subsequent command they can run.

fbiville:
Why not fix it in this PR? This has not been merged yet.

fbiville:
I suppose we can close this as the replacement PR https://github.com/projectriff/riff/pull/1213 is merged? cc @ericbottard @scothis

ericbottard:
Closing, covered by #1213

PR #1210, opened 2019-04-02
Upgrade Knative Serving to 0.5.0

### LINKED ISSUE #1121: Functions sometimes do not become ready causing riff function create to hang
FATS has bumped into a couple failures since upgrading to Knative 0.4 where functions are never detected as ready. This causes `riff function create` to not return when using `--wait` or `--verbose`, and eventually the job times out. Before the timeout, the build pod runs to completion, the revision pod starts up normally and then shuts down normally after a minute because the autoscaler saw no traffic.

I've only seen this happen with command functions, but it could be more widespread. I have not been able to reproduce this outside of FATS.

https://travis-ci.com/projectriff/riff/jobs/181726274#L2938

scothis:
Fixes #1121

### DIFF
--- a/cmd/riff/main.go
+++ b/cmd/riff/main.go
@@ -30,13 +30,13 @@ var (
 		"stable": {
 			ManifestVersion: "0.1",
 			Istio: []string{
-				// diff https://storage.googleapis.com/projectriff/istio/istio-riff-knative-serving-v0-4-1.diff
-				"https://storage.googleapis.com/projectriff/istio/istio-riff-knative-serving-v0-4-1.yaml",
+				// diff https://storage.googleapis.com/projectriff/istio/istio-riff-knative-serving-v0-5-0.diff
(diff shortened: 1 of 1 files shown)


scothis:
This PR does not change the compile target for Knative serving to 0.5.0 because there is a breaking change in knative/pkg that blocks compiling serving@0.5.0 and eventing@0.4.0 in the same project.

scothis:
@trisberg We're not going to upgrade eventing to 0.5, so we can't wait. Unless you're advocating that we remove Knative eventing instead of deprecate it for riff 0.3?

trisberg:
I meant wait on upgrading the serving yaml until we remove eventing or eventing is compatible

scothis:
@trisberg there are features (bug fixes) in serving 0.5 that we want.

trisberg:
OK, let me see what Domino says ...

PR #1211, opened 2019-04-02
Deprecate Knative Eventing

### LINKED ISSUE #1209: Add deprecation warnings for channel and subscription commands
We don't believe that Knative eventing is the future for the core stream processing model in riff. We should warn users that these commands will be going away in the future once we have a replacement.

Ref #1159

scothis:
We will remove Knative Eventing from riff after the 0.3.0 release branch
is forked from master. A replacement for Eventing is being developed.

Fixes #1209


pkg/riff/commands/channel.go
```diff
@@ -18,20 +18,23 @@ package commands
...
+	v1 "k8s.io/api/core/v1"
 )
 
 func Channel() *cobra.Command {
 	return &cobra.Command{
 		Use:   "channel",
-		Short: "Interact with channel related resources",
+		Short: "[DEPRECATED] Interact with channel related resources",
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			cmd.OutOrStderr().Write([]byte("Knative eventing is deprecated, and will be removed in a future release\n"))
```
markfisher: maybe this should just say "channel is deprecated"

also we might want to say "will be replaced"... could even mention "streams" (worst case, we change names) so that it's clear that the same general use-cases will still be supported

pkg/riff/commands/subscription.go
```diff
@@ -45,7 +46,11 @@ const (
 func Subscription() *Command {
 	return &Command{
 		Use:   "subscription",
-		Short: "Interact with subscription-related resources",
+		Short: "[DEPRECATED] Interact with subscription-related resources",
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			cmd.OutOrStderr().Write([]byte("Knative eventing is deprecated, and will be removed in a future release\n"))
```
markfisher: ditto (but mention "processor"?)

Issue #1215, opened 2019-04-03
flakey --verbose output during function create


jldec:
installed riff from master `riff cli: 0.3.0-snapshot (522852c0567cc1c10914710eb658ff7c71f66679`

The output of `function create` with `--verbose` is inconsistent, and sometimes missing important information e.g. in the example below the detection stage fails - the output shows this on the first run, but not on the 2nd. 

```
~/riff/riff (master)$ riff service delete square

riff service delete completed successfully

~/riff/riff (master)$ riff function create square --git-repo https://github.com/jldec/node-fun-square --artifact square-noway.js --verbose
Waiting for LatestCreatedRevisionName
Waiting on function creation: checkService failed to obtain service status for observedGeneration 1
LatestCreatedRevisionName available: square-xc5fh
Waiting on function creation: function creation incomplete: service status unknown: RevisionMissing
default/square-cp7ff-pod-264625[build-step-git-source-0]: {"level":"warn","ts":1554291888.3841095,"logger":"fallback-logger","caller":"logging/config.go:65","msg":"Fetch GitHub commit ID from kodata failed: \"ref: refs/heads/upstream/release-0.5\" is not a valid GitHub commit ID"}
default/square-cp7ff-pod-264625[build-step-git-source-0]: {"level":"info","ts":1554291889.9198723,"logger":"fallback-logger","caller":"git-init/main.go:101","msg":"Successfully cloned \"https://github.com/jldec/node-fun-square\" @ \"master\" in path \"/workspace\""}
default/square-cp7ff-pod-264625[build-step-detect]: 2019/04/03 11:44:56 Trying group of 3...
default/square-cp7ff-pod-264625[build-step-detect]: 2019/04/03 11:44:56 ======== Results ========
default/square-cp7ff-pod-264625[build-step-detect]: 2019/04/03 11:44:56 Cloud Foundry OpenJDK Buildpack: pass
default/square-cp7ff-pod-264625[build-step-detect]: 2019/04/03 11:44:56 Cloud Foundry Build System Buildpack: skip
default/square-cp7ff-pod-264625[build-step-detect]: 2019/04/03 11:44:56 Java Function Buildpack: fail
default/square-cp7ff-pod-264625[build-step-detect]: 2019/04/03 11:44:56 Trying group of 3...
default/square-cp7ff-pod-264625[build-step-detect]: 2019/04/03 11:44:56 ======== Output: NPM Buildpack ========
[...]

PR #1217, opened 2019-04-03
Verbose function create wait for logs before exiting

glyn:
Fixes https://github.com/projectriff/riff/issues/1215


glyn:
> Generally looks good, but feels like we are abusing the errChan. Maybe the channel just needs to be renamed to something more descriptive of its role.

Yes, I was conscious of that problem, but I'll give it more thought. Thanks for raising it!

scothis:
One of the `function create` commands stalled. I can't say that it's because of this change, but it's certainly in the same area of the codebase.

https://travis-ci.com/projectriff/riff/jobs/190046137#L3722

glyn:
> One of the `function create` commands stalled. I can't say that it's because of this change, but it's certainly in the same area of the codebase.
> 
> https://travis-ci.com/projectriff/riff/jobs/190046137#L3722

Hmmm. I can't immediately see why my code change would have caused this. I wonder if it's our old friend https://github.com/projectriff/riff/issues/1121 recurring?

pkg/core/function.go
```diff
@@ -242,6 +248,7 @@ func (c *client) makeBuildSourceSpec(options CreateFunctionOptions) *build.Sourc
 }
 
 func (c *client) displayFunctionCreationProgress(serviceNamespace string, serviceName string, logWriter io.Writer, stopChan <-chan struct{}, errChan chan<- error) {
+	defer close(errChan)
```
glyn: This avoids a deadlock in the `--local-path` case where `displayFunctionCreationProgress` was returning **without** closing `errChan`.

glyn:
> > One of the `function create` commands stalled. I can't say that it's because of this change, but it's certainly in the same area of the codebase.
> > https://travis-ci.com/projectriff/riff/jobs/190046137#L3722
> 
> Hmmm. I can't immediately see why my code change would have caused this. I wonder if it's our old friend #1121 recurring?

My code changes were the cause. A `--local-path` build had a return path that did not close `loggingTerminationChan`.

scothis:
@glyn can you merge/rebase master, there's a AZP patch that should help unblock things

scothis:
These test failures are real

glyn:
> These test failures are real

Not sure how to interpret this. Are the failures real at master too or are they specific to this PR? Re-merged master to try to see the wood for the trees...

glyn:
Failed again with:
```
Function did not produce expected result:
   expected: RIFF
   actual: curl http://10.30.0.245:32202/ -H 'Host: fats-local-uppercase-node.riff-1771-2.example.com' -H 'Content-Type: text/plain' -d riff -v
The command "ci/fats.sh lite" exited with 1.
```
A test PR (with just a README change) based on HEAD of the master branch passes all the checks, so it seems the above failure is caused by this PR's code changes.

pkg/core/function.go
```diff
@@ -185,21 +185,27 @@ func (c *client) CreateFunction(buildpackBuilder Builder, options CreateFunction
...
+				}
 			}
-			err := c.waitForSuccessOrFailure(ns, s.Name, 1, stopChan, errChan, options.Verbose)
 			if err != nil {
 				return nil, nil, nil, err
 			}
 
 			// fetch latest Service and Revision, ignore errors
-			s1, err := c.serving.ServingV1alpha1().Services(s.Namespace).Get(s.Name, metav1.GetOptions{})
+			s1, err := c.serving.ServingV1alpha1().Services(s.Namespace).Get(s.Name, v1.GetOptions{})
```
scothis: please revert to `metav1`, the name `v1` is too generic and leads to conflicts with other `v1` packages.

glyn: Fixed.

PR #1218, opened 2019-04-03
Update registries doc

### LINKED ISSUE #1094: Document how to authenticate local builds to container registries
For each supported environment on projectriff.io, we document how to configure in-cluster builds to the container registry, but never show how to authenticate local builds to the registry.

Local (Minikube, Docker for Desktop): none (it's open to all)
DockerHub: `docker login`
GCR: `gcloud auth configure-docker`

We should also document that local builds require a Docker daemon.

Follow up to #1054

trisberg:
Fixes #1094

### DIFF
--- a/Using-Registries.md
+++ b/Using-Registries.md
@@ -1,4 +1,68 @@
-## Using a Private Registry for Images
+## Using Registries
+
+### Using Docker Hub with builds
+
+#### Initialize a namespace
+
+When you initialize the namespace use the `--dockerhub` flag together with your Docker ID.
+
+If you are initializing the `default` namespace with Docker Hub, then you can use:
+
+```
+riff namespace init default --dockerhub <docker-id>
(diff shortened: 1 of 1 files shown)


Using-Registries.md
```diff
@@ -1,4 +1,68 @@
-## Using a Private Registry for Images
+## Using Registries
+
+### Using Docker Hub with builds
+
+#### Initialize a namespace
+
+When you initialize the namespace use the `--dockerhub` flag together with your Docker ID.
```
scothis: :cough:

trisberg:
I would agree, once we have something other than getting-started guides there. I would prefer to have something documented anywhere before that happens though.

trisberg:
Created a projectriff.io issue - https://github.com/projectriff/projectriff.io/issues/87

PR #1220, opened 2019-04-03
replace fileutils

### LINKED ISSUE #1216: Replace fileutils package with external module
riff should obtain this package from github.com/pivotal/go-ape instead of duplicating it.

sbawaska:
replace fileutils which is now extracted into its own repo at github.com/pivotal/go-ape

closes #1216


pkg/core/system.go
```diff
@@ -21,12 +21,11 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"github.com/pivotal/go-ape"
```
scothis: I find it odd that we can change the import and everything else continues to work. Also it's common to separate core and third party import groups. I'd expect this line to be down by line 31.

^ and for every other file modified.

sbawaska: The package name was not changed when the code was extracted, so everything continues to work. 

I will group the imports.

Issue #1232, opened 2019-04-05
Upgrade to Istio 1.0.7


scothis:
It contains patches for critical Envoy vulnerabilities.

Refs https://github.com/knative/serving/pull/3668

scothis:
We upgrading to Istio 1.0.7 independent of Knative Serving

PR #1235, opened 2019-04-08
Uprade Istio to 1.0.7

scothis:
This is the version of Istio that will be included with Knative Serving
0.5.1 and fixes a couple CVEs in Envoy.

Refs #1232

### DIFF
--- a/cmd/riff/main.go
+++ b/cmd/riff/main.go
@@ -30,8 +30,8 @@ var (
 		"stable": {
 			ManifestVersion: "0.1",
 			Istio: []string{
-				// diff https://storage.googleapis.com/projectriff/istio/istio-riff-knative-serving-v0-5-1.diff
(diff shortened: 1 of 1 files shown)


scothis:
I'm jumping the gun on Knative 0.5.1, this is pulled from the release branch but won't change between now and 0.5.1. It's why the version of Istio is different than the serving version.

In the nearish future, Knative is going to stop distributing Istio, so we'll need to create our own distributions.

trisberg:
since there is no knative-0.5.1 artifacts yet, maybe just rename the curated yaml to `riff-istio-v1-0-7-.yaml` or `istio-v1-0-7-riff-knative-serving-v0-5-0.yaml`

scothis:
Renamed the file to `istio-v1.0.7-riff.yaml`

Issue #1159, opened 2019-03-15
Stream driven Functions


scothis:
In riff 0.0.x we had support for streaming messages between Functions mediated by Topics backed by Kafka. With the transition to Knative, we lost the ability to stream messages from a Channel to a function and were limited to item-at-a-time request-reply semantics.

We still care deeply about the potential of stateful stream processing in a polyglot, FaaS environment.

Knative serving is no longer restricted to http/1.1, which unlocks support for streaming protocols like h2, WebSockets, gRPC and RSocket. Knative eventing has had no movement towards supporting any protocol other than http/1.1.

High level things to do:
- bless a baseline protocol for channels and invokers to support
  - define additional stream semantics as needed
- create stream enabled channels
- update all invokers to support streams
- ensure proper autoscaling for stream based requests
- bridge http/1.1 into a stream based function

markfisher:
We have started working on a PoC that includes the following components:

1. Stream CRD with "name" and "provider" (other properties will be added, e.g. "partitions")
2. Processor CRD with "inputs", "outputs", and "function"
3. Stream Gateway providing a generic API that can be backed by different event-log based messaging systems (such as Kafka or Kinesis); initially exploring [liiklus](https://github.com/bsideup/liiklus)
4. Processor will connect to the Stream Gateway(s) for the input/output Streams and interact with the target Function (with which it's colocated in a Pod) via its Invoker layer using gRPC for bidirectional streaming

scothis:
Progress continues within https://github.com/projectriff/system and https://github.com/projectriff/kafka-gateway. The CLI is now able to create Stream and Processor resources.

ericbottard:
Closing as the initial implementation largely covers the bullet points listed here

PR #1237, opened 2019-04-08
Drop Knative Eventing

scothis:
We'll replace the capability with a new streamming model.

This also upgrade Knative Serving and Build to 0.5 which was previously
blocked by also depending on Knative Eventing 0.4.

Refs: #1159


scothis:
Azure burped, and retriggering doesn't update the check for some reason. The build actually passed here https://dev.azure.com/projectriff/projectriff/_build/results?buildId=551

PR #1247, opened 2019-04-10
Fix kustomize integration

### LINKED ISSUE #1226: Figure out how to run the kustomize tests on Windows
This is the only remaining issue from #1174

fbiville:
This upgrades Kustomize to the highest stable version and includes
a simple workaround that can work now because of kustomize
improved normalization logic.

There is an open issue against kustomize to get rid of
normalization entirely when one uses their in-memory FS impl like
riff does: https://github.com/kubernetes-sigs/kustomize/issues/963

Fixes #1226


scothis:
Merged master into the PR to resolve the go.mod conflict

PR #1254, opened 2019-04-17
Introduce `credentials` command

### LINKED ISSUE #817: Allow for multiple push credentials

fbiville:
Refactor secret management to leverage it in this new operation.

Fixes #817.

### DIFF
--- a/pkg/core/credentials.go
+++ b/pkg/core/credentials.go
@@ -23,6 +23,10 @@ type SetCredentialsOptions struct {
 	RegistryUser string
 }
 
+type ListCredentialsOptions struct {
+	NamespaceName string
+}
+
 func (o *SetCredentialsOptions) secretType() secretType {
 	switch {
 	case o.DockerHubId != "":
@@ -62,6 +66,16 @@ func (c *client) SetCredentials(options SetCredentialsOptions) error {
 		return err
 	}
 	return c.updateServiceAccount(namespace, secret.Name, serviceAccount)
+}
+
+func (c *client) ListCredentials(options ListCredentialsOptions) (*corev1.SecretList, error) {
+	namespace := c.explicitOrConfigNamespace(options.NamespaceName)
+	initLabelKeys := sortedKeysOf(getInitLabels())
+
+	fmt.Printf("Listing secrets matching label keys %v in namespace %q\n", initLabelKeys, namespace)
+	return c.kubeClient.CoreV1().Secrets(namespace).List(v1.ListOptions{
+		LabelSelector: existsSelectors(initLabelKeys),
+	})
 }
 
 func (c *client) convertDockerHubSecret(namespace, secret, username string, labels map[string]string) (*corev1.Secret, error) {
--- a/pkg/core/credentials.go
+++ b/pkg/core/credentials.go
@@ -46,12 +46,12 @@ func (o *SetCredentialsOptions) secretType() secretType {
 }
 
 func (c *client) SetCredentials(options SetCredentialsOptions) error {
-	namespace := options.NamespaceName
+	namespace := c.explicitOrConfigNamespace(options.NamespaceName)
 	secret, err := c.kubeClient.CoreV1().Secrets(namespace).Get(options.SecretName, v1.GetOptions{})
 	if err != nil && !errors.IsNotFound(err) {
 		return err
 	}
-	if secret != nil {
+	if err == nil {
 		fmt.Printf("Deleting existing secret %q in namespace %q\n", secret.ObjectMeta.Name, namespace)
 		if err = c.kubeClient.CoreV1().Secrets(namespace).Delete(secret.Name, &v1.DeleteOptions{}); err != nil {
 			return err
--- a/pkg/core/credentials.go
+++ b/pkg/core/credentials.go
@@ -50,31 +50,21 @@ func (o *SetCredentialsOptions) secretType() secretType {
 
 func (c *client) SetCredentials(options SetCredentialsOptions) error {
 	namespace := c.explicitOrConfigNamespace(options.NamespaceName)
-	secret, err := c.kubeClient.CoreV1().Secrets(namespace).Get(options.SecretName, v1.GetOptions{})
-	if err != nil && !errors.IsNotFound(err) {
-		return err
-	}
-	if err == nil {
-		fmt.Printf("Deleting existing secret %q in namespace %q\n", secret.ObjectMeta.Name, namespace)
-		if err = c.kubeClient.CoreV1().Secrets(namespace).Delete(secret.Name, &v1.DeleteOptions{}); err != nil {
-			return err
-		}
-	}
 	initLabels := getInitLabels()
-	secret, err = c.convertAdditionalSecret(&options, initLabels)
-	if err = c.createBasicSecret(namespace, secret); err != nil {
+	if err := c.createOrUpdateSecret(namespace, initLabels, options); err != nil {
 		return err
 	}
 
+	secretName := options.SecretName
 	serviceAccount, err := c.kubeClient.CoreV1().ServiceAccounts(namespace).Get(BuildServiceAccountName, v1.GetOptions{});
 	if errors.IsNotFound(err) {
-		if err = c.createServiceAccount(namespace, secret.Name, initLabels); err != nil {
+		if err = c.createServiceAccount(namespace, secretName, initLabels); err != nil {
 			return err
 		}
 	} else if err != nil {
 		return err
 	} else {
-		err = c.updateServiceAccount(namespace, secret.Name, serviceAccount)
+		err = c.updateServiceAccount(namespace, secretName, serviceAccount)
 		if err != nil {
 			return err
 		}
@@ -110,6 +100,23 @@ func (c *client) ListCredentials(options ListCredentialsOptions) (*corev1.Secret
 func (c *client) DeleteCredentials(options DeleteCredentialsOptions) error {
 	namespace := c.explicitOrConfigNamespace(options.NamespaceName)
 	return c.kubeClient.CoreV1().Secrets(namespace).Delete(options.Name, &v1.DeleteOptions{})
+}
+
+func (c *client) createOrUpdateSecret(namespace string, initLabels map[string]string, options SetCredentialsOptions) error {
+	secret, secretGetErr := c.kubeClient.CoreV1().Secrets(namespace).Get(options.SecretName, v1.GetOptions{})
+	if secretGetErr != nil && !errors.IsNotFound(secretGetErr) {
+		return secretGetErr
+	}
+
+	secret, err := c.convertAdditionalSecret(&options, initLabels)
+	if err != nil {
+		return err
+	}
+
+	if errors.IsNotFound(secretGetErr) {
+		return c.createBasicSecret(namespace, secret)
+	}
+	return c.updateBasicSecret(namespace, secret)
 }
 
 func (c *client) convertDockerHubSecret(namespace, secret, username string, labels map[string]string) (*corev1.Secret, error) {
@@ -165,6 +172,16 @@ func (c *client) createBasicSecret(namespace string, secret *corev1.Secret) erro
 		secret.StringData["username"])
 
 	_, err := c.kubeClient.CoreV1().Secrets(namespace).Create(secret)
+	return err
+}
+
+func (c *client) updateBasicSecret(namespace string, secret *corev1.Secret) error {
+	fmt.Printf("Updating secret %q with basic authentication to server %q for user %q\n",
+		secret.ObjectMeta.Name,
+		secret.Annotations["build.knative.dev/docker-0"],
+		secret.StringData["username"])
+
+	_, err := c.kubeClient.CoreV1().Secrets(namespace).Update(secret)
 	return err
 }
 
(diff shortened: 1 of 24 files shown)


trisberg:
I would prefer to follow the `namespace init` command and have the namespace be a required argument like:

```
riff namespace add-credentials default --secret new-creds ...
```

fbiville:
I thought about it and concluded that allowing an argument would be ambiguous. 

In the case of `namespace init`, it is clear that the main argument is the namespace.
In the case of `namespace add-credentials`, the main argument could either be the credentials to add or the target namespace.

Because of the ambiguity, I tried to avoid the problem altogether by not allowing any argument and passing everything as an option.

fbiville:
@trisberg should we allow the equivalent of `namespace init`'s `--no-secret`? That would allow users to create secrets on their own and bind it to the riff build service account.

pkg/riff/commands/namespace.go
```diff
@@ -153,3 +154,53 @@ func NamespaceCleanup(c *core.Client) *cobra.Command {
 
 	return command
 }
+
+func NamespaceAddCredentials(c *core.Client) *cobra.Command {
+	options := core.NamespaceAddCredentialsOptions{}
+
+	command := &cobra.Command{
+		Use:     "add-credentials",
```
scothis: Consider renaming this to `set-credentials`, not all operations are adding a new secret, sometimes a user will want to update existing credentials. In HTTP semantics, this should be a PUT, not a POST.

Or better yet to avoid the hyphenation which we don't have anywhere else in the CLI `credentials set`

pkg/riff/commands/namespace.go
```diff
@@ -153,3 +154,53 @@ func NamespaceCleanup(c *core.Client) *cobra.Command {
...
+			if err := (*c).NamespaceAddCredentials(options); err != nil {
+				return err
+			}
+			PrintSuccessfulCompletion(cmd)
+			return nil
+		},
+	}
+
+	command.Flags().StringVar(&options.NamespaceName, "namespace", "", "the `namespace` of the credentials to be added")
+	command.Flags().StringVarP(&options.SecretName, "secret", "s", "", "the name of a `secret` containing credentials for the image registry")
```
scothis: What if we default the secret to `push-credentials` to update the credentials when the namespace is first inited?

pkg/riff/commands/namespace.go
```diff
@@ -153,3 +154,53 @@ func NamespaceCleanup(c *core.Client) *cobra.Command {
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if err := (*c).NamespaceAddCredentials(options); err != nil {
+				return err
+			}
+			PrintSuccessfulCompletion(cmd)
+			return nil
+		},
+	}
+
+	command.Flags().StringVar(&options.NamespaceName, "namespace", "", "the `namespace` of the credentials to be added")
```
scothis: It seems odd that this uses a flag to pick the namespace, while `namespace init` uses an argument. I'd expect them to use the same pattern. (although I personally prefer using a flag)

pkg/core/namespace.go
```diff
@@ -210,30 +299,42 @@ func (c *client) initServiceAccount(options *NamespaceInitOptions, initLabels ma
...
+func (c *client) updateServiceAccount(namespace, secretName string, serviceAccount *corev1.ServiceAccount) error {
+	secretAlreadyPresent := false
+	for _, s := range serviceAccount.Secrets {
+		if s.Name == secretName {
+			secretAlreadyPresent = true
+			break
 		}
 	}
+	if secretAlreadyPresent {
+		fmt.Printf("Secret %q is already bound to serviceaccount %q in namespace %q. Skipping.\n", secretName, serviceAccount.Name, namespace)
```
scothis: Do we need this? The message is more likely to confuse users than inform them.

pkg/core/namespace.go
```diff
@@ -143,6 +167,33 @@ func (c *client) NamespaceCleanup(options NamespaceCleanupOptions) error {
...
 
+func (c *client) NamespaceAddCredentials(options NamespaceAddCredentialsOptions) error {
+	namespace := options.NamespaceName
+	secretName := options.SecretName
+	exists, err := c.checkSecretExists(namespace, secretName)
+	if err != nil {
+		return err
+	}
+	if exists {
+		return fmt.Errorf("secret %q already exists", secretName)
```
scothis: rather than error out, update the secret

pkg/riff/commands/namespace.go
scothis: Another approach would be to move this into a top level command `riff credentials`

pkg/riff/commands/namespace.go
fbiville: A `riff credentials` command is what I was arguing for :)

fbiville: Not sure about the `set`. For now, this command adds a secret only if there is none matching.

pkg/riff/commands/namespace.go
fbiville: As mentioned above, this command adds secrets, but does not update them.

pkg/core/namespace.go
fbiville: It's the same logic as for `namespace init`, I just rephrased it to make it clearer.

pkg/riff/commands/namespace.go
fbiville: Yep, there is a similar discussion going on in the PR. I'd argue for a separate top-level command.

fbiville:
Answering other comments here.

 - how do I see what secrets are installed?
I would say this is `kubectl` job. If we really want to and choose to move the current logic to `riff credentials add/set`, then we could have `riff credentials list`. Should be done in another PR, I think.
 - how do I remove a secret?
That should be done via a separate issue and another command. This will be easier if we opt for a `riff credentials` top-level command.
 - what if there is more than one secret for the same registry?
Hmmm, I do not see the issue with that in the current impl. What am I missing?
 - why can't I update a secret
We could repurpose the current add-only command.
 - which secret is used to default the image prefix? How do I change the default?
The one specified when running `namespace init`. Changing the default means re-running `namespace init`. We could introduce a separate standalone command, but that may be confusing to users and should be done in another PR, imo.

pkg/core/namespace.go
fbiville: Do we want this for this command only or for `namespace init` as well? This piece of code is currently shared by both.

pkg/riff/commands/namespace.go
fbiville: Updated. This command now allows secret updates. `riff namespace init` semantics remains unchanged.

pkg/riff/commands/namespace.go
fbiville: Now the command allows secret update. While I see the proposed symmetry with `namespace init`, it seems a bit strange to me to allow to write `credentials set` without naming them at all. Happy to revisit if you think it's more consistent that way.

pkg/riff/commands/namespace.go
fbiville: Done in newest commit.

pkg/core/namespace.go
fbiville: Fixed in newest commit.

pkg/core/namespace.go
fbiville: The newest commit does not change `namespace init` behavior but allows secret updates in `credentials set`.

fbiville:
If we could preserve @badrfoot commit before merging that PR, it would be nice!

pkg/riff/commands/credentials.go
```diff
@@ -0,0 +1,165 @@
...
+		Short: "Interact with credentials related resources",
+	}
+}
+
+func CredentialsSet(c *core.Client) *cobra.Command {
+	options := core.SetCredentialsOptions{}
+
+	command := &cobra.Command{
+		Use:     "set",
+		Short:   "create or update secret and bind it to the " + env.Cli.Name + " service account (created if not found)",
```
scothis: The name of the service account is always riff-build, the name of the cli is irrelevant here

pkg/riff/commands/credentials.go
```diff
@@ -0,0 +1,165 @@
...
+		Example: `  ` + env.Cli.Name + ` credentials set --secret mysecret --namespace default --docker-hub johndoe`,
+		Args:    cobra.ExactArgs(credentialsSetNumberOfArgs),
+		PreRunE: FlagsValidatorAsCobraRunE(
+			FlagsValidationConjunction(
+				FlagsDependency(Set("namespace"), ValidDnsSubdomain("namespace")),
```
scothis: We're not creating a namespace, that the namespace exists is all the validation we should care about.

pkg/riff/commands/credentials.go
```diff
@@ -0,0 +1,165 @@
...
+				NotBlank("secret"),
+				AtMostOneOf("gcr", "docker-hub", "registry-user"),
+				FlagsDependency(Set("image-prefix"), NotBlank("image-prefix")),
+				FlagsDependency(Set("image-prefix"), IsTrue("enable-image-prefix")),
```
scothis: Use of `--image-prefix` should imply `--enable-image-prefix`, not require it

pkg/riff/commands/credentials.go
```diff
@@ -0,0 +1,165 @@
...
+		},
+	}
+
+	command.Flags().StringVar(&options.NamespaceName, "namespace", "", "the `namespace` of the credentials to be added")
+	command.Flags().StringVarP(&options.SecretName, "secret", "s", "push-credentials", "the name of a `secret` containing credentials for the image registry")
+	command.Flags().StringVar(&options.GcrTokenPath, "gcr", "", "path to a file containing Google Container Registry credentials")
+	command.Flags().StringVar(&options.DockerHubId, "docker-hub", "", "Docker ID for authenticating with Docker Hub; password will be read from stdin")
+	command.Flags().StringVar(&options.Registry, "registry", "", "registry server host, scheme must be \"http\" or \"https\" (default \"https\")")
+	command.Flags().StringVar(&options.RegistryUser, "registry-user", "", "registry username; password will be read from stdin")
+	command.Flags().BoolVar(&options.EnableImagePrefix, "enable-image-prefix", false, "allow image prefix creation/update")
```
scothis: rename `--enable-image-prefix` -> `--set-image-prefix`

pkg/core/image_prefix.go
```diff
@@ -0,0 +1,37 @@
+package core
+
+import (
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+)
+
+func DetermineImagePrefix(userDefinedPrefix string, dockerHubId string, gcrTokenPath string) (string, error) {
```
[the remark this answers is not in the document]
fbiville: I have been hesitant to create a secret-related struct that others (all the `XxxOptions`) would bring in and change existing code. 
You are right though, the image prefix logic should remain unchanged no matter how many new registries we decide to support 👍

pkg/riff/commands/credentials.go
fbiville: `--allow-image-prefix-change`?

pkg/core/credentials.go
```diff
@@ -0,0 +1,228 @@
...
+
+func (c *client) SetCredentials(options SetCredentialsOptions) error {
+	namespace := c.explicitOrConfigNamespace(options.NamespaceName)
+	secret, err := c.kubeClient.CoreV1().Secrets(namespace).Get(options.SecretName, v1.GetOptions{})
+	if err != nil && !errors.IsNotFound(err) {
+		return err
+	}
+	if err == nil {
+		fmt.Printf("Deleting existing secret %q in namespace %q\n", secret.ObjectMeta.Name, namespace)
+		if err = c.kubeClient.CoreV1().Secrets(namespace).Delete(secret.Name, &v1.DeleteOptions{}); err != nil {
```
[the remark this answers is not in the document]
fbiville: Thanks, that's a good point. I noticed the delete-then-create also happens during `namespace init`. As the semantics of the latter are slightly different, I assume this is OK? Or is it?

pkg/core/credentials.go
```diff
@@ -0,0 +1,228 @@
...
+
+	fmt.Printf("Listing secrets matching label keys %v in namespace %q\n", initLabelKeys, namespace)
+	return c.kubeClient.CoreV1().Secrets(namespace).List(v1.ListOptions{
+		LabelSelector: existsSelectors(initLabelKeys),
+	})
+}
+
+func (c *client) DeleteCredentials(options DeleteCredentialsOptions) error {
+	namespace := c.explicitOrConfigNamespace(options.NamespaceName)
+	return c.kubeClient.CoreV1().Secrets(namespace).Delete(options.Name, &v1.DeleteOptions{})
```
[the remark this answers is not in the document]
fbiville: yep, it was the end of a long day 😅

scothis:
Ideas have been folded into #1266

PR #1265, opened 2019-05-06
Rethinking the riff CLI

scothis:
Reimaginging the riff CLI as a client for [projectriff/system](https://github.com/projectriff/system) and [projectriff/cnab-riff](https://github.com/projectriff/cnab-riff).

While the bones of many commands have been laid out, the exact shape of the commands are still in flux.

- `riff function` -> `riff function`
- `riff namespace` replaced by cnab 
- `riff service` -> `riff requestprocessor`
- `riff system` replaced by cnab

We previously removed Knative eventing types, but those would have mapped as:
- `riff channel` -> `riff stream`
- `riff subscription` -> `riff streamprocessor`

Tests are modeled after the Knative table test pattern for commands instead of ginkgo. The standard k8s fake clients are used instead of mocks. Assertions are written by hand instead of gomega, but any assertion library could be used. Each command exists in a single file, all helpers live in another package.


ericbottard:
I think there are certain patterns in `knctl` which we may find interesting. This would be a good opportunity to adopt them

docs/riff.md
```diff
@@ -1,26 +1,28 @@
...
+* [riff application](riff_application.md)	 - 
+* [riff completion](riff_completion.md)	 - 
+* [riff credential](riff_credential.md)	 - 
+* [riff function](riff_function.md)	 - 
+* [riff requestprocessor](riff_requestprocessor.md)	 - 
+* [riff stream](riff_stream.md)	 - 
+* [riff streamprocessor](riff_streamprocessor.md)	 - 
+* [riff version](riff_version.md)	 - 
 
+###### Auto generated by spf13/cobra on 6-May-2019
```
fbiville: do we want this mention? how come it was not generated before?

pkg/client/client.go
```diff
@@ -0,0 +1,98 @@
...
+	buildv1alpha1 "github.com/projectriff/system/pkg/client/clientset/versioned/typed/build/v1alpha1"
+	runv1alpha1 "github.com/projectriff/system/pkg/client/clientset/versioned/typed/run/v1alpha1"
+	streamv1alpha1 "github.com/projectriff/system/pkg/client/clientset/versioned/typed/stream/v1alpha1"
+	"k8s.io/client-go/kubernetes"
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+type Client interface {
```
fbiville: this is not new, but I think the name is too vague and this interface may suffer from undesired growth because its name (and therefore its intent) is not clear

pkg/client/client.go
```diff
@@ -0,0 +1,98 @@
...
+	DefaultNamespace() string
+	Core() corev1.CoreV1Interface
```
fbiville: I think I get what the rest of the operations do, but what about `Core`? what does it do?

pkg/riff/commands/credential_list_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+	"github.com/projectriff/riff/pkg/riff/commands"
+	rifftesting "github.com/projectriff/riff/pkg/testing"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+func TestCredentialListCommand(t *testing.T) {
+	rifftesting.Table{{
```
fbiville: it's subjective but I find this particular structure of tests to be be a bit hard to read (especially with the actual invocation is being at the bottom of the file which breaks the reading linearity)
also: how do you handle error cases with that structure?

pkg/riff/commands/application.go
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package commands
+
+import "github.com/spf13/cobra"
+import "github.com/projectriff/riff/pkg/riff"
+
+func NewApplicationCommand(p *riff.Params) *cobra.Command {
```
fbiville: having the main subcommands compose their own subcommands is a great idea 👍

pkg/riff/commands/root.go
```diff
@@ -0,0 +1,45 @@
...
+
+package commands
+
+import (
+	"github.com/projectriff/riff/pkg/env"
+	"github.com/projectriff/riff/pkg/riff"
+	"github.com/spf13/cobra"
+)
+
+func NewRootCommand(p *riff.Params) *cobra.Command {
```
fbiville: nice idea again, this is better structured than the former `wiring.go`

pkg/riff/commands/application.go
ericbottard: This is one of the areas where we should definitely borrow from `knctl` IMO

trisberg: Or knative/client, is `knctl` going to be around once the `client` gains more traction?

ericbottard:
Some of the aspects of knctl I like:
- node commands automatically generate their help / behavior, based on the leaf commands they host
- the `*Options` classes basically end up implementing the (GoF) Command pattern, which is better than capturing references to fields of them in a closure IMO
- the UI / Table / coloring package is not too shabby
-

AUTHORS
```diff
@@ -3,4 +3,5 @@
 # This does not necessarily list everyone who has contributed code, since in
 # some cases, their employer may be the copyright holder.  To see the full list
 # of contributors, see the revision history in source control.
+
```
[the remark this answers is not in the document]
scothis: I like a newline between header blocks and content

docs/riff.md
scothis: We'll need to kill that otherwise CI will fail. I'm not sure why it's there because we're using the same version of cobra and the same code to generate docs, maybe it's an internal library that version bumped.

pkg/client/client.go
scothis: how about renaming the package to `k8s`?

pkg/client/client.go
scothis: it returns a k8s clientset for the 'core' api group.

docs/riff.md
ericbottard: `cmd.DisableAutoGenTag = true`

scothis: done

ericbottard:
While we're at it, we **may** want to have a look at https://github.com/google/wire

pkg/riff/commands/credential_list_test.go
fbiville: I'm wondering if that approach does not end up sharing too much. Can all the test cases fit in this structure? Having an error case in the PR would help.

pkg/client/client.go
scothis: renamed

pkg/riff/commands/credential_list_test.go
scothis: With this model of testing we can still create one-off go test cases. There is nothing forcing all tests to use the table model. With Ginkgo, it's odd if some of your tests are not managed by ginkgo.

As for error cases, the fake clients have reactors that are able to receive and mutate all requests and also return custom errors. I'll work up a test case with a network failure as an example.

scothis: @fbiville check out https://github.com/projectriff/riff/blob/84bce9fd5ded95cb72ad85e3b6df0157f90a0621/pkg/riff/commands/credential_list_test.go#L98-L103

glyn:
> * `riff namespace` replaced by cnab

Don't we still need `riff namespace init` for setting up any additional namespaces after install?

@sbawaska Does the base invocation image support `riff namespace init` of any namespaces?

pkg/testing/table.go
```diff
@@ -0,0 +1,281 @@
...
+	"github.com/projectriff/riff/pkg/riff"
+	"github.com/spf13/cobra"
+	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/runtime"
+	clientgotesting "k8s.io/client-go/testing"
+)
+
+type CommandTable []CommandTableRecord
+
+type CommandTableRecord struct {
```
glyn: It's not immediately obvious how to use this struct properly, so developers might be reduced to copying and hacking other tests without necessarily understanding any subtleties. Compare this with ginkgo/gomega which has good documentation and is familiar to most of the riff team. If we decide to switch riff to this struct instead of ginkgo, I think we'll need documentation for developers.

pkg/riff/commands/credential_delete_test.go
```diff
@@ -0,0 +1,144 @@
...
+
+func TestCredentialDeleteCommand(t *testing.T) {
+	t.Parallel()
+
+	credentialName := "test-credential"
+	credentialAltName := "test-alt-credential"
+	defaultNamespace := "default"
+	credentialLabel := "projectriff.io/credential"
+
+	table := testing.CommandTable{
```
glyn: A general observation about this style of test versus ginkgo is that the table is flat whereas a ginkgo test is structured as a tree which makes it easy to add test variations without much repetition. The tree structure also helps me, and hopefully others, to think through the cases and sub-cases of the test. When I want to work on a particular node of the ginkgo tree, I can focus it, add extra tests to that node, etc.

One particular use case is TDD. I add a failing test, focus on it, implement the code to make the test pass and then, if there's a problem with the test, I can immediately use the debugger to see what's going on (without hitting the same breakpoint courtesy of other tests). I can then make code changes and ultimately remove the focus (and if I forget to do this, the build will break).

glyn:
A timely article on table-driven tests: https://dave.cheney.net/2019/05/07/prefer-table-driven-tests.

sbawaska:
> > * `riff namespace` replaced by cnab
> 
> Don't we still need `riff namespace init` for setting up any additional namespaces after install?
> 
> @sbawaska Does the base invocation image support `riff namespace init` of any namespaces?

@glyn I think the plan is the `namespace init`'s responsibility is going to be taken over by the `credentials` command. Do I have that right @scothis?

pkg/testing/table.go
scothis: Yes, it needs docs. I've been avoiding docs while things are in flux.

pkg/riff/commands/credential_delete_test.go
scothis: Your usage of Ginkgo is the cleanest I've seen. And I see how it can be easier to add an incremental test. That said, I find a hierarchal structure harder to understand unless you created it (this is true for anything with a custom spatial dimension that doesn't exist in the physical world). It's much more complicated to refactor since you need to have a deep understanding of how any particular state is shared between any other tests. For testing, I find being explicit to always be preferable to being dry. Denormalization is good in this case.

I do agree that being able to focus on a single record under test is handy. I added `Skip` and `Focus` bool fields to the table record struct. I'm still not sold on this focus being better than a filter on the `go test` command. Let me know what you think and we can keep or remove it.

glyn: > Your usage of Ginkgo is the cleanest I've seen. And I see how it can be easier to add an incremental test. That said, I find a hierarchal structure harder to understand unless you created it (this is true for anything with a custom spatial dimension that doesn't exist in the physical world). It's much more complicated to refactor since you need to have a deep understanding of how any particular state is shared between any other tests. For testing, I find being explicit to always be preferable to being dry. Denormalization is good in this case.

Yes, I can see there are some advantages of using a "flatter" and more explicit testing approach, even if it is quite a bit more verbose. I'm personally happy to try other approaches so long as they are reasonably stable and well documented. But I would say that we need to agree to this as a team and take time to spread the testing work around so we can all become familiar with the new approach. It would be all too easy for a couple of us to create a silo here.

> 
> I do agree that being able to focus on a single record under test is handy. I added `Skip` and `Focus` bool fields to the table record struct. I'm still not sold on this focus being better than a filter on the `go test` command. Let me know what you think and we can keep or remove it.

I think the `Focus` field will be quicker to set than writing a filter, so I'd like to keep it and see how it works out.

.gitignore
```diff
@@ -1,32 +1,4 @@
-debug
```
fbiville: I think we should keep the `.gitignore` as is

pkg/riff/commands/credential_delete_test.go
fbiville: Maybe this work in progress should be pushed to a branch on this repo, so we can start to split work amongst us and spread the work that way (requires some coordination though, but I think code review is not necessarily enough to fully grasp the pros and cons).

scothis:
Moved work to https://github.com/projectriff/riff/tree/reset to enable more collaboration.

PR #1267, opened 2019-05-10
Test option validation

scothis:
- created structure for options table tests
- added option validation tests for function commands
- aliased FieldError types with projectriff to avoid a foreign import


pkg/cli/options.go
```diff
@@ -33,3 +34,67 @@ func ValidateOptions(opts apis.Validatable) func(cmd *cobra.Command, args []stri
...
+
+type ListOptions struct {
+	Namespace     string
+	AllNamespaces bool
+}
+
+func (opts *ListOptions) Validate(ctx context.Context) *FieldError {
+	errs := &FieldError{}
+
+	if opts.Namespace == "" && !opts.AllNamespaces {
```
fbiville: my initial reaction when reading this: shouldn't we allow an empty namespace for the default namespace case?
after examining the codebase further, it may be worth documenting that this function is typically called in commands' PreRunE block and the default namespace is usually already injected prior to that PreRunE block (typically via the `NamespaceFlag` function)

pkg/cli/options.go
```diff
@@ -33,3 +34,67 @@ func ValidateOptions(opts apis.Validatable) func(cmd *cobra.Command, args []stri
...
+}
+
+func (opts *DeleteOptions) Validate(ctx context.Context) *FieldError {
+	errs := &FieldError{}
+
+	if opts.Namespace == "" {
+		errs = errs.Also(ErrMissingField("namespace"))
+	}
+
+	if opts.All && len(opts.Names) != 0 {
```
fbiville: maybe for later (or never :D), some kind of `xorValidation(condition, errorWhenAllFalse, errorWhenAllTrue)` function could be introduced (it's the second occurence of this kind in the PR)

pkg/riff/commands/credential_delete_test.go
```diff
@@ -24,6 +24,27 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
+func TestCredentialDeleteOptions(t *testing.T) {
```
fbiville: where/how would you test that the validation is properly "plugged" to its related command? I do not see that integration bit being tested here (which is not extremely important after the fact but is quite natural when test-driving the impl)

pkg/testing/options_table.go
```diff
@@ -0,0 +1,183 @@
...
+		if i != "" {
+			r = append(r, i)
+		}
+	}
+	return r
+}
+
+type OverrideOptionsFunc func(cli.Validatable)
+
+func isOverideOptionsFunc(t reflect.Type) bool {
```
fbiville: s/Overide/Override/

pkg/cli/options.go
scothis: This is options validation, not flag or args validation. The flags and args have already been bound to the options before this code runs. I'll add a note about the options lifecycle.

pkg/cli/options.go
scothis: I've had the same thought. It gets even more fun when there are three or more values.

pkg/riff/commands/credential_delete_test.go
scothis: That test would need to be defined for the command and not the options. In essence, the command would define arguments that would fail validation and then validate that the command failed. Many of the command tests were written before we had option validation, I'll add a test to ensure the options validator is wired properly for each implemented command.

pkg/testing/options_table.go
scothis: rm'ed

pkg/riff/commands/credential_delete_test.go
scothis: done

pkg/cli/options.go
scothis: done

fbiville: 👍

PR #1269, opened 2019-05-13
Implement invoke command

scothis:
also adds support for:
- optional arguments
- base double dash arguments
- os/exec testing


scothis:
@markfisher raised a question as to whether invoke should be a command. Even if we don't want the command, dealing with option arguments and testing code that shells out to another executable is interesting.

pkg/riff/commands/requestprocessor_invoke_test.go
```diff
@@ -0,0 +1,408 @@
...
+	table.Run(t, commands.NewRequestProcessorInvokeCommand)
+}
+
+func TestHelperProcess_RequestProcessorInvoke(t *testing.T) {
+	if os.Getenv("GO_WANT_HELPER_PROCESS") != "1" {
+		return
+	}
+	// TODO assert arguments
+	fmt.Fprintf(os.Stderr, "Command executed: %s\n", shellquote.Join(argsAfterBareDoubleDash(os.Args)...))
+	os.Exit(0)
```
glyn: Won't this terminate the test executable?

pkg/riff/commands/requestprocessor_invoke_test.go
```diff
@@ -0,0 +1,408 @@
...
+	}
+
+	table.Run(t, commands.NewRequestProcessorInvokeCommand)
+}
+
+func TestHelperProcess_RequestProcessorInvoke(t *testing.T) {
+	if os.Getenv("GO_WANT_HELPER_PROCESS") != "1" {
+		return
+	}
+	// TODO assert arguments
```
glyn: Is this TODO to be done in this PR?

pkg/riff/commands/requestprocessor_invoke_test.go
```diff
@@ -0,0 +1,408 @@
...
+	fmt.Fprintf(os.Stderr, "Command executed: %s\n", shellquote.Join(argsAfterBareDoubleDash(os.Args)...))
+	os.Exit(0)
+}
+
+func TestHelperProcess_RequestProcessorInvokeError(t *testing.T) {
+	if os.Getenv("GO_WANT_HELPER_PROCESS") != "1" {
+		return
+	}
+	// TODO assert arguments
```
glyn: Is this TODO to be done in this PR?

pkg/riff/commands/requestprocessor_invoke_test.go
```diff
@@ -0,0 +1,408 @@
...
+	fmt.Fprintf(os.Stderr, "Command executed: %s\n", shellquote.Join(argsAfterBareDoubleDash(os.Args)...))
+	os.Exit(-1)
```
glyn: Won't this terminate the test executable?

docs/riff_requestprocessor_invoke.md
```diff
@@ -20,7 +20,9 @@ riff requestprocessor invoke [flags]
 
 ```
   -h, --help               help for invoke
+      --json               <todo>
```
glyn: Missing help text.

pkg/cli/arguments.go
```diff
@@ -38,12 +41,18 @@ func Args(argDefs ...Arg) cobra.PositionalArgs {
 				// consume all remaining args
 				arity = len(args) - offset
 			}
-			if len(args) < arity {
+			if len(args)-offset < arity {
+				if argDef.Optional {
+					continue
+				}
 				// TODO create a better message saying what is missing
```
glyn: Does this TODO need addressing now?

pkg/riff/commands/requestprocessor_invoke_test.go
scothis: yes, and it's intentional. Vanilla test runs will be caught by `if os.Getenv("GO_WANT_HELPER_PROCESS") != "1" {` and return before hitting the exit. When the test is run from within the exec call, it will hit the line and exit as normal.

The gotcha is that there are two levels of `go test` processes, this is the inner process. See the ExecHelper docs https://github.com/projectriff/riff/pull/1269/files#diff-ca1a901d21e48e7cd3b09c67650a643dR110 for detail and https://npf.io/2015/06/testing-exec-command/ for the inspiration.

pkg/riff/commands/requestprocessor_invoke_test.go
scothis: The TODO can be removed. Because the args are logged on the next line, they can be asserted in the mainline test. For example https://github.com/projectriff/riff/pull/1269/files#diff-97f91502cd33a30cf14d894726f823c5R369

scothis:
Marking this command as hidden until we figure out what to do with it. It can be called by clients that know it exists, but is otherwise invisible.

PR #1270, opened 2019-05-14
Set credentials implementation

scothis:
- implement credential set command for DockerHub, GCR and basic auth
- optionally set default image prefix when setting credentials

Add `Stdin` to the CommandTable to make content available to be read.

Refs #1254


pkg/riff/commands/credential_set.go
```diff
@@ -49,42 +97,198 @@ func NewCredentialSetCommand(c *cli.Config) *cobra.Command {
...
+				}
+			}
 
-			return err
+			return nil
 		},
 	}
 
 	cli.NamespaceFlag(cmd, c, &opts.Namespace)
+	cmd.Flags().StringVar(&opts.DockerHubId, "docker-hub", "", "<todo>")
```
glyn: Missing help text.

glyn:
I have a general concern about "riff credentials set" - it sounds like a noun and is a bit confusing. Also, it's not clear to me why image prefix has anything at all to do with credentials.

pkg/riff/commands/credential_set.go
scothis: Yes, All help text is missing. I've intentionally deferred writing help so that we can address it comprehensively and consistently for all commands. There are 99 `<todo>`s currently.

scothis:
> Please could you supply the missing help text as that will actually help me review the code and tests.

Which command/flag is vague? Perhaps we can find names that are more self descriptive.

> I have a general concern about "riff credentials set" - it sounds like a noun and is a bit confusing.

Do you have another suggestion for a command that will create/update a credential? While we can, I tried to avoid creating separate insert and update commands since the update is nearly identical to a delete and create.

This command was originally defined in #1254

> Also, it's not clear to me why image prefix has anything at all to do with credentials.

The credentials are for an image registry. For DockerHub and GCR we are able to infer the image prefix based solely on the credentials. It's a connivence for the user. The default image prefix was originally specified as part of `namespace init` being derived from the credentials.

I considered adding an explicit `--default-image-prefix` to allow a user to set a custom value, but that made more sense as part of a `riff config` command (which doesn't currently exist).

markfisher:
> Do you have another suggestion for a command that will create/update a credential? While we can, I tried to avoid creating separate insert and update commands since the update is nearly identical to a delete and create.

If we have explicit "create" and "update" then the former would fail if a credential with that name already exists and the latter would fail if it doesn't yet exist. Maybe that's a good thing since, for example, "set" might inadvertently apply an update?

fbiville:
@glyn the last updated specs for the command were: https://github.com/projectriff/riff/issues/817#issuecomment-489078485.

pkg/riff/commands/credential_set.go
```diff
@@ -17,25 +17,73 @@
...
 
 type CredentialSetOptions struct {
 	cli.ResourceOptions
+	DockerHubId       string
+	DockerHubPassword string
+	GcrTokenPath      string
+	Registry          string
+	RegistryUser      string
+	RegistryPassword  string
+	SetAsDefault      bool
```
glyn: Should the field name incorporate something to indicate it is about the default image prefix rather than some kind of default credential?

glyn: Also, this isn't tested.

scothis: Renamed to `--set-default-image-prefix`

scothis:
> If we have explicit "create" and "update" then the former would fail if a credential with that name already exists and the latter would fail if it doesn't yet exist. Maybe that's a good thing since, for example, "set" might inadvertently apply an update?

I tried [splitting out set into independent create and update commands](https://github.com/projectriff/riff/commit/c9ef076f5c597480b614e3b52974c6da606b613a). It doesn't end up simplifying either command and now there are two commands where changes need to be kept in sync (I kept as much code in common as practical). If we really think the UX around create and update vs set is what we want, then we can do this, but I'd rather not.

Splitting create and update is useful when there are semantic differences between the two commands. For example, some options may only be available during create. Update is used to mutate an existing resource, preserving some core essence of its original self. This is not true for credentials. The same flags are available for both commands and they are applied the same way. The resulting secret data from the create command and the update command are identical, no matter the preexisting state.

Another way to think about this, imagine I took away `kubectl apply` and forced you to use `kubectl create` and `kubectl update` instead. There are times when `k create` and `k updated` carry meaningful semantics, but most of the time those difference don't matter and forcing the user to think about which to use is a cognitive burden, especially when the result is identical.

markfisher:
> There are times when `k create` and `k updated` carry meaningful semantics, but most of the time those difference don't matter and forcing the user to think about which to use is a cognitive burden, especially when the result is identical.

wouldn't that same logic apply to the other commands like "function" and "application"?

scothis:
Merging so we can use the command with riff-system. Happy to continue iterating in issues and PRs

PR #1283, opened 2019-05-19
Local application and function builds

scothis:
For application and function create calls with --local-path, we use pack
to build and publish the container from local source before creating the
application or function resource.

The created application/function will not have a Source field in the
spec. The image is resolved to a digest by the riff system controller.

This PR uses mockery to mock the pack client for tests. Use of the mock
is limited to the few tests that use pack. We can swap to a different
mock/stub library in the future.


scothis:
We may not ultimately want to support local builds vs builds from local source. In either case, this PR is interesting as to how something like pack would integration into the new cli model.

Makefile
```diff
@@ -44,3 +44,12 @@ verify-docs: docs
 .PHONY: clean-docs
 clean-docs:
 	rm -fR docs
+
+.PHONY: check-mockery
+check-mockery:
+    # Use go get in GOPATH mode to install/update counterfeiter. This avoids polluting go.mod/go.sum.
```
ericbottard:
```suggestion
    # Use go get in GOPATH mode to install/update mockery. This avoids polluting go.mod/go.sum.
```

scothis:
Merging to unblock further updates to riff-system

PR #1284, opened 2019-05-20
Better support for shell completion

ericbottard:


pkg/riff/commands/completion.go
```diff
@@ -17,20 +17,54 @@
...
 		Example: "<todo>",
 		Args:    cli.Args(),
-		Run: func(cmd *cobra.Command, args []string) {
-			cmd.Root().GenBashCompletion(cmd.OutOrStdout())
-		},
+		PreRunE: cli.ValidateOptions(opts),
+		RunE:    cli.ExecOptions(c, opts),
 	}
 
+	cmd.Flags().StringVar(&opts.Shell, cli.StripDash(cli.ShellFlagname), "bash", "shell to generate completion for: bash or zsh")
```
scothis: If you don't like a flag here, subcommands for bash and zsh would be preferable to an opaque argument. Positional arguments don't make sense to me when the value is from a fixed set, especially n=2.

ericbottard: I think I was drawn to arg as a matter of habit. Your argument does make sense, going with a flag

ericbottard:
Merged as 19c6df7c

PR #1285, opened 2019-05-20
Forbid setting --namespace and --all-namespaces

scothis:
This condition is tricky to validate traditionally because the namespace
default is derived from the kubeconfig which is parsed after the command
has been defined.


scothis:
CI burped, but ran successful in https://dev.azure.com/projectriff/projectriff/_build/results?buildId=1118

ericbottard:
Merged as e1ac3670

PR #1286, opened 2019-05-20
Drop update commands

scothis:
Update is a command that needs more thought. In most cases, we don't
want a generic update command, but more specific commands about the
specific part of the resource being updated.

Delete and create can substibute as a minimal update.


markfisher:
LGTM

PR #1292, opened 2019-05-24
Backport default image prefix support

scothis:
The --image flag is now optional for applications and functions. The
default value is `_`, which applies the configured default prefix
with the application/function's name. A custom repo or tag can be
specified like `_/repo:tag`.

For in-cluser builds this resolution occurs in the reconciler. For local
builds, it occurs within the CLI. In both cases, the resolution logic is
defined within projectriff/system.

Depends on https://github.com/projectriff/system/pull/21

### DIFF
--- a/pkg/riff/commands/credential_apply.go
+++ b/pkg/riff/commands/credential_apply.go
@@ -39,6 +39,7 @@ type CredentialApplyOptions struct {
 	Registry              string
 	RegistryUser          string
 	RegistryPassword      []byte
+	DefaultImagePrefix    string
 	SetDefaultImagePrefix bool
 }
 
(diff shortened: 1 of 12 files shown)


pkg/riff/commands/credential_apply.go
```diff
@@ -154,9 +162,8 @@ func NewCredentialApplyCommand(c *cli.Config) *cobra.Command {
 	cmd.Flags().StringVar(&opts.GcrTokenPath, cli.StripDash(cli.GcrFlagName), "", "path to Google Container Registry service account token `file`")
 	cmd.Flags().StringVar(&opts.Registry, cli.StripDash(cli.RegistryFlagName), "", "registry `url`")
 	cmd.Flags().StringVar(&opts.RegistryUser, cli.StripDash(cli.RegistryUserFlagName), "", "`username` for a registry, the password must be provided via stdin")
+	cmd.Flags().StringVar(&opts.DefaultImagePrefix, cli.StripDash(cli.DefaultImagePrefixFlagName), "", fmt.Sprintf("use this `registry` as the default for built images, implies %s", cli.SetDefaultImagePrefixFlagName))
```
trisberg: we should change `registry` to be either `prefix` or possibly `registry-prefix` since it is the registry + / + user or project like `--default-image-prefix harbor.springdeveloper.com/dev`

scothis: We need to dispel the notion that the repository has anything to do with the username to authenticate the request. In some cases it does, but more often than not, there is no connection.

I agree that `registry` is the wrong phrase here since the value is for a `repository` prefix.

PR #1293, opened 2019-05-28
Normalize READY/SUCCEEDED columns in listings

scothis:
The READY and SUCCEEDED columns are now a uniform STATUS column, the
value within the column is no longer "True", but "Ready" or "Succeeded"
depending on the condition type.

If the condition is false, the reason is printed as an error.

### DIFF
--- a/pkg/cli/format.go
+++ b/pkg/cli/format.go
@@ -45,17 +45,31 @@ func FormatConditionStatus(cond *duckv1alpha1.Condition) string {
 	status := string(cond.Status)
 	switch status {
 	case "True":
-		return Ssuccessf(status)
+		return Ssuccessf(string(cond.Type))
 	case "False":
-		return Serrorf(status)
+		if cond.Reason == "" {
+			// display something if there is no reason
+			return Serrorf("not-" + string(cond.Type))
+		}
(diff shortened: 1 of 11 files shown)


markfisher:
should we consider changing the SUCCEEDED cases to READY? The former implies a one-off action whereas the latter seems more appropriate for something that can change over time (e.g. rebuild)

If we were to do that we could maybe keep the READY header (with True/False in the column values) which would avoid overloading the use of "status" since in the underlying resource, this is just one part of what's provided via the status.

scothis:
> should we consider changing the SUCCEEDED cases to READY?

Yes, but not in this PR/repo. Switching from a one-off to a continuous model is where we want to be, but builds right now are one-off, which is why they use succeeded as their primary condition.

> If we were to do that we could maybe keep the READY header (with True/False in the column values) 

The primary driver for this change is to provide more meaningful errors when the condition is false. See https://github.com/projectriff/system/pull/21#pullrequestreview-241934167

> which would avoid overloading the use of "status" since in the underlying resource, this is just one part of what's provided via the status.

The value printed is the most succinct summary of the resources' status, which is why I feel comfortable reusing the term. We can go back to having a READY/SUCCEEDED column heading instead.

PR #1294, opened 2019-05-29
Tail log output

scothis:
Each command with a compute component now has a `--tail` flag for the
create command which runs until the resource is ready, and a `tail`
command which will return log output until the process is terminated.


pkg/k8s/wait.go
```diff
@@ -0,0 +1,59 @@
...
+	"k8s.io/apimachinery/pkg/watch"
+)
+
+func WaitUntilReady(watcher watch.Interface) {
+	for {
+		select {
+		case ev := <-watcher.ResultChan():
+			switch ev.Type {
+			case watch.Added, watch.Modified:
+				// TODO create an interface for this
```
glyn: Looks like a missed TODO. If not, please could you create an issue and link to it.

pkg/kail/logger.go
```diff
@@ -0,0 +1,121 @@
...
+func (c *logger) ProcessorLogs(ctx context.Context, processor *streamv1alpha1.Processor, since time.Duration, out io.Writer) error {
+	selector, err := labels.Parse(fmt.Sprintf("%s=%s", stream.ProcessorLabelKey, processor.Name))
+	if err != nil {
+		panic(err)
+	}
+	containers := []string{"function", "processor"}
+	return c.stream(ctx, processor.Namespace, selector, containers, since, out)
+}
+
+func (c *logger) stream(ctx context.Context, namespace string, selector labels.Selector, containers []string, since time.Duration, out io.Writer) error {
```
glyn: This is implemented using kail, but I thought kail was removed on the basis that it was faulty (duplicating or omitting logs). What's the rationale for risking kail again?

pkg/riff/commands/application_create.go
```diff
@@ -142,6 +146,22 @@ func (opts *ApplicationCreateOptions) Exec(ctx context.Context, c *cli.Config) e
...
+	if opts.Tail {
+		// cancel ctx when application becomes ready
+		ctx, cancel := context.WithCancel(ctx)
+		go func() {
+			defer cancel()
+			applicationWatch, err := c.Build().Applications(opts.Namespace).Watch(metav1.ListOptions{
+				ResourceVersion: application.ResourceVersion,
+			})
+			if err != nil {
+				return
```
glyn: Unless we know the error has already been logged, it would be good to log it before returning to provide some visibility into what went wrong.

pkg/k8s/wait.go
```diff
@@ -0,0 +1,59 @@
...
+					if stream.Status.IsReady() {
+						return
+					}
+				} else if processor, ok := ev.Object.(*streamv1alpha1.Processor); ok {
+					if processor.Status.IsReady() {
+						return
+					}
+				}
+			case watch.Deleted, watch.Error:
+				return
```
glyn: I think we should return errors rather than swallowing them, but we'll also need to add logging in the caller.

pkg/riff/commands/application_create.go
```diff
@@ -142,6 +146,22 @@ func (opts *ApplicationCreateOptions) Exec(ctx context.Context, c *cli.Config) e
...
+			}
+			defer applicationWatch.Stop()
+			k8s.WaitUntilReady(applicationWatch)
+		}()
+		return c.Kail.ApplicationLogs(ctx, application, time.Minute, c.Stdout)
```
glyn: The `time.Minute` constant is repeated in the four similar blocks and should probably be factored out to preserve consistency.

pkg/riff/commands/application_create.go
```diff
@@ -142,6 +146,22 @@ func (opts *ApplicationCreateOptions) Exec(ctx context.Context, c *cli.Config) e
 		return err
 	}
 	c.Successf("Created application %q\n", application.Name)
+	if opts.Tail {
```
glyn: This block is repeated four times. It would be worth factoring out the common logic to avoid quadruple maintenance. If that was done, it may even be worth writing a unit test of the common logic to avoid having to repeat certain test code. This may be something for a future PR though.

ericbottard:
I love the addition of the `riff xxx tail` variants that allow tailing after the fact.

> Also, it would be easy to re-introduce the --wait...
Somehow, I realized I did not use that flag that much, often relying on a separate watch for that (whereas an ever running kail is a bit more involved). Not against it, but definitely another PR

> Why reintroduce kail?
If the rationale is that we haven't found (and looked for?) anything better, I'm fine with that for now. As long as the feature is hidden behind an interface, I think the short term value add is worth it

glyn:
I like the `tail` variants too, but I'd like to understand what the behaviour is supposed to be.

For instance, if I kick off a "create" with "--tail" and let it run to completion, what will happen if I then issue "tail" on the same resource? Will the output include the creation logs or will it just consist of subsequent logs?

Also, does "tail" keep tailing even after the resource has reached "ready"? I hope so.

pkg/kail/logger.go
scothis: kail does still miss some logs and duplicate others. I'm reintroducing it to restore the functionality. All kail usage is encapsulated within this package, so we can swap it out with something better in the future.

pkg/k8s/wait.go
scothis: done

pkg/riff/commands/application_create.go
scothis: done

pkg/k8s/wait.go
scothis: done

pkg/riff/commands/application_create.go
scothis: exposing since as a flag, with a default of 1 second.

scothis:
> like the tail variants too, but I'd like to understand what the behaviour is supposed to be.
>
> For instance, if I kick off a "create" with "--tail" and let it run to completion, what will happen if I then issue "tail" on the same resource? Will the output include the creation logs or will it just consist of subsequent logs?

By default, the tail command will only pickup logs created since the command was executed, so you shouldn't see duplicate logs. However, if you ran the command while the create is still not ready then you would see duplicate logs during that overlap.

> Also, does "tail" keep tailing even after the resource has reached "ready"? I hope so.

The `riff * tail` command will keep running until it is interrupted, similar to the actual `tail` command. The `--tail` flag for a `riff * create` command will exit once the created resource is ready.

There's a gotcha in that kail seems to only detect logs for containers that have not exited. So once a build finishes, even though the pod still exists, it won't show logs. I think this is also a contributor to the missing logs as the build has a sequence of init containers, the window where a given container is live can be very short.

scothis:
> Also, it would be easy to re-introduce the `--wait` switch for users who want to ensure a resource is ready, but don't want the clutter of logs. Is this worth adding in?

I haven't found the `--wait` flag to be particularly useful. In general, the pattern in k8s is to define all of your resources at once and then let the system make everything available. Ordering the creation of resources is a sign that your system is fragile and won't recover from a hard failure. The exception being CI where you want to prove something worked as quickly as possible so you can throw it away, but I want the logs in that case. The other exception being configuration of the Kubernetes API service, which we are not doing.

We can `--wait` to see if any users complain before implementing.

glyn:
> > Also, it would be easy to re-introduce the `--wait` switch for users who want to ensure a resource is ready, but don't want the clutter of logs. Is this worth adding in?
> 
> I haven't found the `--wait` flag to be particularly useful. In general, the pattern in k8s is to define all of your resources at once and then let the system make everything available. Ordering the creation of resources is a sign that your system is fragile and won't recover from a hard failure. The exception being CI where you want to prove something worked as quickly as possible so you can throw it away, but I want the logs in that case. The other exception being configuration of the Kubernetes API service, which we are not doing.
> 
> We can `--wait` to see if any users complain before implementing.

Agreed. The one potentially useful case I've found in the past doesn't even work: create a function and wait for it to become available and then invoke it. IIRC invocation can fail because containers take a little while to get going after the status has reached "ready". But thanks for this discussion, as it's good to get rationale written down for posterity.

pkg/k8s/wait.go
```diff
@@ -0,0 +1,52 @@
...
+				return fmt.Errorf("error waiting for ready")
+			}
+			if obj, ok := ev.Object.(metav1.Object); !ok || obj.GetUID() != target.GetUID() {
+				// event is not for the target resource
+				continue
+			}
+			switch ev.Type {
+			case watch.Added, watch.Modified:
+				// use reflection since there is no common interface
+				if readyFunc := reflect.ValueOf(ev.Object).Elem().FieldByName("Status").Addr().MethodByName("IsReady"); readyFunc.Kind() == reflect.Func {
```
glyn: I can't find the concrete types which this is reflecting upon, so I have no way of knowing whether the reflection is valid. Any chance of a comment to make this navigable?

pkg/riff/commands/application_create.go
glyn: Er, the code is unchanged and still refers to `time.Minute`. What am I missing?

pkg/k8s/wait.go
scothis: actually done now

pkg/riff/commands/application_create.go
scothis: actually done now

pkg/riff/commands/application_create.go
scothis: > This block is repeated four times. It would be worth factoring out the common logic to avoid quadruple maintenance.

The behavior waiting for a resource to become ready is already separate from the behavior that grabs the logs. I'd rather not conflate those two concerns.

> it may even be worth writing a unit test of the common logic to avoid having to repeat certain test code.

Test coverage for commands is quite high, but coverage for utility code is quite low and needs to be improved

> This may be something for a future PR though.

agreed

PR #1303, opened 2019-06-03
Expose content-type flag to stream creation command

fbiville:
Depends on https://github.com/projectriff/system/pull/28

### DIFF
--- a/pkg/riff/commands/stream_create.go
+++ b/pkg/riff/commands/stream_create.go
@@ -93,7 +93,7 @@ func NewStreamCreateCommand(c *cli.Config) *cobra.Command {
 
 	cli.NamespaceFlag(cmd, c, &opts.Namespace)
 	cmd.Flags().StringVar(&opts.Provider, cli.StripDash(cli.ProviderFlagName), "", "`name` of stream provider")
-	cmd.Flags().StringVar(&opts.ContentType, cli.StripDash(cli.ContentTypeName), "", "`content-type` of stream")
+	cmd.Flags().StringVar(&opts.ContentType, cli.StripDash(cli.ContentTypeName), "", "`MIME type` for message payloads accepted by the stream")
 
 	return cmd
 }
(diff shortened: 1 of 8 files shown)


pkg/riff/commands/stream_create_test.go
```diff
@@ -46,6 +46,15 @@ func TestStreamCreateOptions(t *testing.T) {
 			},
 			ShouldValidate: true,
 		},
+		{
+			Name: "valid stream",
```
scothis:
```suggestion
			Name: "with content type",
```

move after the `no provider` test

pkg/riff/commands/stream_create.go
```diff
@@ -30,7 +30,8 @@ import (
 type StreamCreateOptions struct {
 	cli.ResourceOptions
 
-	Provider string
+	Provider    string
+	ContentType string
```
scothis: do we need to apply any validation?

fbiville: we can check the presence of `/` and check that the first slash is not also the last. I may be missing something but I don't think we can do much more.

fbiville:
> Looks like go.mod is missing

Yep, I need https://github.com/projectriff/system/pull/28 merged first.

fbiville:
Please do not merge as-is, the changes includes a reference to my riff system fork at the moment.

pkg/riff/commands/stream_create.go
```diff
@@ -42,6 +43,14 @@ func (opts *StreamCreateOptions) Validate(ctx context.Context) *cli.FieldError {
 		errs = errs.Also(cli.ErrMissingField(cli.ProviderFlagName))
 	}
 
+	contentType := opts.ContentType
+	if contentType != "" {
+		index := strings.Index(contentType, "/")
```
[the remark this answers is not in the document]
fbiville: I would personally wait for a second usage before extracting the validation to a separate function but I don't mind doing it now.

ericbottard:
Merged as d488573d

PR #1308, opened 2019-06-04
Run all functions for full FATS runs

scothis:
also use Docker Hub since it's so much faster and we have credentials


ci/fats.sh
```diff
@@ -60,7 +60,13 @@ travis_fold end system-install
 # run test functions
 source $fats_dir/functions/helpers.sh
 
-for test in command; do
+if [ "$mode" = "full" ]; then
+  functions=(command)
```
sbawaska: Should we run everything in full mode and a subset in lite mode? This seems inverted to me.

scothis: yep, 🤦‍♂️

PR #1310, opened 2019-06-05
Properly control flow waiting for a resource to become ready

scothis:
- no more race conditions
- consistent log output
- push timeout into the context

### DIFF
--- a/pkg/riff/commands/function_create.go
+++ b/pkg/riff/commands/function_create.go
@@ -172,29 +172,35 @@ func (opts *FunctionCreateOptions) Exec(ctx context.Context, c *cli.Config) erro
 	}
 	c.Successf("Created function %q\n", function.Name)
 	if opts.Tail {
-		// cancel ctx when function becomes ready
-		ctx, cancel := context.WithCancel(ctx)
-		defer cancel()
+		// err guarded by Validate()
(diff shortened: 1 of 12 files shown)


pkg/riff/commands/function_create.go
```diff
@@ -172,29 +172,35 @@ func (opts *FunctionCreateOptions) Exec(ctx context.Context, c *cli.Config) erro
...
+		if ctx.Err() == context.DeadlineExceeded {
 			c.Errorf("Timeout after %q waiting for %q to become ready\n", opts.WaitTimeout, opts.Name)
 			c.Infof("To view status run: %s function list %s %s\n", c.Name, cli.NamespaceFlagName, opts.Namespace)
 			c.Infof("To continue watching logs run: %s function tail %s %s %s\n", c.Name, opts.Name, cli.NamespaceFlagName, opts.Namespace)
-			cancel()
-		})
-		defer timer.Stop()
-
-		return c.Kail.FunctionLogs(ctx, function, cli.TailSinceCreateDefault, c.Stdout)
+			errChan <- cli.ErrSilent
```
glyn: I think ErrSilent should wrap the underlying error otherwise silent errors will become difficult to reason about.

scothis: done

PR #1311, opened 2019-06-06
Dump riff, build and serving controller logs on FATS failure

scothis:


scothis:
I intentionally split them into independent steps so that each log would be stored independently.

PR #1312, opened 2019-06-06
DNM: Vendor our go dependencies

scothis:
This means our repo is bigger, but builds will no longer be impacted by
network connectivity issues. It should also speed up CI quite a bit.


scothis:
I'm not sure if this is a good idea, but I wanted to see how CI behaves.

scothis:
By skipping module downloads, we can shave ~90s off each unit test, stage and FATS lite job.

glyn:
Should help local builds too.

ericbottard:
Let's try it!

scothis:
What could go wrong?

scothis:
> What could go wrong?

for starters a silent merge conflict :)

PR #1313, opened 2019-06-06
Stop generating HTML coverage reports

scothis:
...Azure Pipelines generates HTML directly from the coverage.xml, and
ignores what we generate.


scothis:
Superseded by #1314

PR #1317, opened 2019-06-07
Improve test coverage

scothis:
- generate more accurate coverage reports


fbiville:
Hmm, I am not sure I am following. What is the value of empty tests here?

sbawaska:
I'm not sure we need to include the testing package. rest looks good.

scothis:
> Hmm, I am not sure I am following. What is the value of empty tests here?

A package with no tests doesn't contribute towards coverage counts. So the resulting coverage numbers are inflated. Adding empty tests gives a more realistic coverage number. My intent is to push more tests to this PR to raise the overall coverage numbers.

fbiville:
That's quite a strange behavior: it may be worth opening an issue against codecov as well?

scothis:
> That's quite a strange behavior: it may be worth opening an issue against codecov as well?

That seems to be normal for go

PR #1322, opened 2019-06-11
Normalize tail control flow

scothis:
Create a race util that will return with the first task to complete.
Each task is run in an independent go routine. If no task completes the
whole process will timeout.

### DIFF
--- a/pkg/util/timeout.go
+++ b/pkg/util/timeout.go
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2019 The original author or authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


pkg/util/timeout.go
```diff
@@ -0,0 +1,49 @@
...
+package util
+
+import (
+	"context"
+	"time"
+)
+
+type RaceTask func(ctx context.Context) error
+
+// RaceUntil runs each task in go routine and returns the result of the first task to
```
fbiville: I need to have a closer look at the implementation, but the description makes me think of Golang's WaitGroup (with a count of 1). It surprisingly does not handle any timeout, though. 
I guess you already looked at it, or? https://golang.org/pkg/sync/#example_WaitGroup

scothis: thanks for the pointer. In this case, we could refactor the function to use WaitGroup, I'm not sure it would add much value either semantically or structurally. Since we only care about the first go routine to be done with a single go routine blocked on completion a single cancelable context is good enough (which also fits nicely with the timeout model. Additionally, we'd still need to setup a channel to track the error status to make sure we only return an error from the first go routine to complete.

If WaitGroup ends up providing more reliable execution, then this func can be refactored without affecting the public interface.

glyn: I looked into the use of WaitGroup and also the errgroup package while reviewing this, but both those utilities are really intended for groups of functions which should all run to completion. So unless a clear advantage of using WaitGroup  (with a count of 1?) emerges, I think we should probably stick with the current code.

PR #1327, opened 2019-06-12
Drop FATS lite

scothis:
For pull requests, FATS provides more friction than benefit. The
important thing is for the cli to be fully tested before release
artifacts are published. We still run FATS for pushed commits, it is
only skipped for pull requests.

...also adding a TODO to add Windows FATS tests.

### DIFF
--- a/ci/fats.sh
+++ b/ci/fats.sh
@@ -4,7 +4,6 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-mode=${1:-full}
 version=`cat VERSION`
 commit=$(git rev-parse HEAD)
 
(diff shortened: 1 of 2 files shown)


ci/fats.sh
```diff
@@ -60,13 +54,7 @@ travis_fold end system-install
...
 source $fats_dir/functions/helpers.sh
 
-if [ "$mode" = "full" ]; then
-  functions=(java java-boot node npm command)
-else
-  functions=(command)
-fi
-
-for test in "${functions[@]}"; do
+for test in command; do
```
sbawaska: Since we are not running in lite mode, should we run `java java-boot node npm` too?

scothis: The cli knows nothing about the content of the function. All running with multiple functions does is test the same flow multiple times.

scothis:
It bothers me that the coverage report is not deterministic. This PR made no changes to go source or test files, yet the numbers are different.

ci/fats.sh
sbawaska: ok, I suppose it is more valuable for cnab-riff to run all functions before putting out a build for our users.

scothis: the most important place to test all of the functions is before releasing the builder (we test all of the functions here already). The second most important place is when releasing the bundle.

PR #1329, opened 2019-06-14
Improve STATUS for function/app list when no reason given

trisberg:
- use first 60 characters of message when reason is blank


trisberg:
closing this - we can get this info wen we implement a status sub-command - https://github.com/projectriff/riff/issues/1334

PR #1331, opened 2019-06-16
Add riff doctor

zhitongLIU:
from hack.commit.push Paris

Added command `riff doctor` to check if the follow namespaces are installed for riff:
- istio-system
- knative-build
- knative-serving
- riff-system

like this
```
➜  riff git:(feature/riff-doctor) ./riff doctor
Installation is OK
```
or
```
➜  riff git:(feature/riff-doctor) ./riff doctor
Something is wrong!
missing istio-system
missing knative-build
missing knative-serving
missing riff-system
```

co-worked from https://github.com/davidthou & https://github.com/zhitongliu

### DIFF
--- a/pkg/riff/commands/doctor.go
+++ b/pkg/riff/commands/doctor.go
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package commands
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/projectriff/riff/pkg/cli"
+	"github.com/spf13/cobra"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type DoctorOptions struct {
+}
+
+var RequiredNamespaces = []string{
+	"istio-system",
+	"knative-build",
+	"knative-serving",
+	"riff-system",
+}
+
+func (opts *DoctorOptions) Validate(ctx context.Context) *cli.FieldError {
+	return cli.EmptyFieldError
+}
+
+func (opts *DoctorOptions) Exec(ctx context.Context, c *cli.Config) error {
+	missingNamespaces, err := checkMissingNamespaces(c)
+
+	if err != nil {
+		c.Errorf(err.Error())
+		return err
+	} else if len(missingNamespaces) > 0 {
+		msg := "Something is wrong!\n"
+		for _, namespace := range missingNamespaces {
+			msg += fmt.Sprintf("missing %s\n", namespace)
+		}
+		c.Errorf(msg)
+	} else {
+		c.Successf("Installation is OK\n")
+	}
+	return nil
+}
+
+func NewDoctorCommand(c *cli.Config) *cobra.Command {
+	opts := &DoctorOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "doctor",
+		Short: "check riff's requirements are installed",
+		Long: strings.TrimSpace(`
+    <todo>
+    `),
+		Example: "riff doctor",
+		Args:    cli.Args(),
+		PreRunE: cli.ValidateOptions(opts),
+		RunE:    cli.ExecOptions(c, opts),
+	}
+
+	return cmd
+}
+
+func checkMissingNamespaces(c *cli.Config) ([]string, error) {
+	missingNamespaces := []string{}
+	namespaces, err := c.Core().Namespaces().List(metav1.ListOptions{})
+
+	if namespaces == nil || len(namespaces.Items) == 0 {
+		missingNamespaces = RequiredNamespaces
+	} else {
+		names := []string{}
+		for _, namespace := range namespaces.Items {
+			names = append(names, namespace.Name)
+		}
+		for _, requiredNamespace := range RequiredNamespaces {
+			if !stringInSlice(requiredNamespace, names) {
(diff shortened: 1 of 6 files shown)


pivotal-issuemaster:
@zhitongLIU Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=projectriff/riff&pullRequestId=1331)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=projectriff/riff&pullRequestId=1331) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@zhitongLIU Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=projectriff/riff&pullRequestId=1331)!

pkg/riff/commands/doctor.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (opts *DoctorOptions) Exec(ctx context.Context, c *cli.Config) error {
+  missingNamespaces, err := checkMissingNamespaces(c)
+
+  if (err != nil) {
+    c.Errorf(err.Error())
+    return err
+  } else if (len(missingNamespaces) > 0) {
+    msg := "Something is worng!\n"
```
davidthou: Small typo: "wrong"

zhitongLIU: fixed

pkg/riff/commands/riff_test.go
```diff
@@ -33,3 +34,14 @@ func TestRiffCommand(t *testing.T) {
 
 	table.Run(t, commands.NewRiffCommand)
 }
+
+func TestRiffMainSubCommands(t *testing.T) {
+	riffCommand := commands.NewRiffCommand(cli.NewDefaultConfig())
+	commands := riffCommand.Commands()
+
+	expectedCount := 7
```
ericbottard: While I understand the intent, I don't think this test should stay as is. The risk is that it becomes a PITA more than anything. Maybe just testing that there exist _some_ subcommands is enough?

pkg/riff/commands/doctor.go
```diff
@@ -0,0 +1,107 @@
...
+	return nil
+}
+
+func NewDoctorCommand(c *cli.Config) *cobra.Command {
+	opts := &DoctorOptions{}
+
+	cmd := &cobra.Command{
+		Use:   "doctor",
+		Short: "check riff's requirements are installed",
+		Long: strings.TrimSpace(`
```
ericbottard: TODO :)

Maybe this is a good opportunity to clarify what this does and what it _doesn't_ do. The short description may imply a false sense of security if the namespaces are present (but are otherwise empty for example).

pkg/riff/commands/doctor.go
```diff
@@ -0,0 +1,107 @@
...
+
+func (opts *DoctorOptions) Validate(ctx context.Context) *cli.FieldError {
+	return cli.EmptyFieldError
+}
+
+func (opts *DoctorOptions) Exec(ctx context.Context, c *cli.Config) error {
+	missingNamespaces, err := checkMissingNamespaces(c)
+
+	if err != nil {
+		c.Errorf(err.Error())
```
scothis: No need to log an error and return it. Logging the error is handled by the caller. By logging and returning the error will be printed on the console twice.

pkg/riff/commands/doctor.go
```diff
@@ -0,0 +1,107 @@
...
+		return err
+	} else if len(missingNamespaces) > 0 {
```
scothis: Since the end of the previous if block returns, we can convert the `else if` to a `if`

pkg/riff/commands/doctor.go
scothis: Every other command is `<TODO>` right now. Adding a long description is certainly nice, but not required at this point.

pkg/riff/commands/doctor.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	return cmd
+}
+
+func checkMissingNamespaces(c *cli.Config) ([]string, error) {
+	missingNamespaces := []string{}
+	namespaces, err := c.Core().Namespaces().List(metav1.ListOptions{})
+
+	if namespaces == nil || len(namespaces.Items) == 0 {
```
scothis: go idioms favor checking if the error needs to be handled and returning early.

The slice should never be nil if there isn't an error.

pkg/riff/commands/doctor.go
```diff
@@ -0,0 +1,107 @@
...
+		missingNamespaces = RequiredNamespaces
+	} else {
+		names := []string{}
+		for _, namespace := range namespaces.Items {
+			names = append(names, namespace.Name)
```
scothis: why not convert into a map, so we can latter avoid searching for a value in a slice?

pkg/riff/commands/doctor_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+	table.Run(t)
+}
+
+func TestDoctorCommand(t *testing.T) {
+	table := rifftesting.CommandTable{
+		{
+			Name:         "installation is ok",
+			Args:         []string{},
+			GivenObjects: requiredNamespacesMocks(commands.RequiredNamespaces),
```
scothis: please be explicit with the GivenObjects. This is one place where being DRY isn't helpful.

With a helper method, changes needed for one test can produce side effects for other tests. When testing it's helpful to avoid changing both the code under test and the test case at the same time.

pkg/riff/commands/doctor_test.go
```diff
@@ -0,0 +1,107 @@
...
+			},
+			ShouldError:  true,
+			ExpectOutput: `inducing failure for list namespaces`,
+		},
+	}
+
+	table.Run(t, commands.NewDoctorCommand)
+}
+
+func requiredNamespacesMocks(namespaces []string) []runtime.Object {
```
scothis: nit: these are actual resources being used by a stub client, not mocks.

pkg/riff/commands/riff_test.go
scothis: Testing that the command is actually wired is typically a function of acceptance tests. In this case, we also know the command is wired as a side effect of the docs being generated. If the command wasn't wired, then riff_doctor.md wouldn't exist (or CI would fail)

scothis:
Thanks again

PR #1332, opened 2019-06-17
Add `--dry-run` flag for all create/apply commands

scothis:
Dry run will print resources to stdout rather than apply them to the k8s cluster. Messages normally on stdout will be sent to stderr instead. This allows using the riff cli to scaffold yaml and apply local mutations before sending the configuration to the API server.

Note: this is not side effect free. While the k8s API server will not be updated, local builds will still run and push images.

I'm open to names other than dry-run.

### DIFF
--- a/pkg/cli/options.go
+++ b/pkg/cli/options.go
@@ -49,6 +49,7 @@ func ValidateOptions(opts apis.Validatable) func(cmd *cobra.Command, args []stri
 
 type Executable interface {
 	Exec(ctx context.Context, c *Config) error
+	IsDryRun() bool
 }
 
 // ExecOptions bridges a cobra RunE function to the Executable interface.  All flags and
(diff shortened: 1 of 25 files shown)


pkg/cli/options.go
```diff
@@ -64,11 +65,25 @@ type Executable interface {
...
+		if opts.IsDryRun() {
+			// reserve Stdout for resources, redirect normal stdout to stderr
+			ctx = withStdout(ctx, c.Stdout)
+			c.Stdout = c.Stderr
+		}
 		return opts.Exec(ctx, c)
 	}
 }
 
+type CommonOptions struct {
```
ericbottard: I find this introduction a bit dangerous and not really necessary. Go tends to use very targeted interfaces, that can be composed on demand.

My fear here is that next thing you know, `CommonOptions` will become a kitchen sink.

To put it another way, why not introduce something dedicated to DryRun and DryRun only?

ericbottard: Also, would be better to dynamically check whether the `Executable opts` conform to "DryRunnable" IMO.

scothis: done

pkg/cli/options_test.go
```diff
@@ -88,6 +88,11 @@ type StubExecOptions struct {
 	execErr error
 }
 
+var (
+	_ cli.Executable = (*StubExecOptions)(nil)
```
ericbottard: 👍

PR #1338, opened 2019-06-19
Enhance/riff doctor tests

zhitongLIU:
It is for fixing comments for https://github.com/projectriff/riff/pull/1331, didn't thought it got merged so quickly
 
- Modified riff sub comments test
- Added description for `riff doctor`

please close it if you don't think this PR is necessaire, thanks.

### DIFF
--- a/pkg/riff/commands/riff.go
+++ b/pkg/riff/commands/riff.go
@@ -40,6 +40,7 @@ func NewRiffCommand(ctx context.Context, c *cli.Config) *cobra.Command {
 	cmd.AddCommand(NewHandlerCommand(ctx, c))
 	cmd.AddCommand(NewStreamCommand(ctx, c))
 	cmd.AddCommand(NewProcessorCommand(ctx, c))
+	cmd.AddCommand(NewDoctorCommand(ctx, c))
 
 	return cmd
 }
--- a/pkg/riff/commands/riff_test.go
+++ b/pkg/riff/commands/riff_test.go
@@ -17,10 +17,14 @@
 package commands_test
 
 import (
+	"context"
+	"github.com/projectriff/riff/pkg/cli"
+	"strings"
 	"testing"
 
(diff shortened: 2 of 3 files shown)


pkg/riff/commands/riff.go
```diff
@@ -40,6 +40,7 @@ func NewRiffCommand(ctx context.Context, c *cli.Config) *cobra.Command {
 	cmd.AddCommand(NewHandlerCommand(ctx, c))
 	cmd.AddCommand(NewStreamCommand(ctx, c))
 	cmd.AddCommand(NewProcessorCommand(ctx, c))
+	cmd.AddCommand(NewDoctorCommand(ctx, c))
```
scothis: I moved the doctor command to be attached to the [root command](https://github.com/projectriff/cli/blob/e5a90713536175ecedaece277e0a98df49ca636e/pkg/riff/commands/root.go#L49). The root command is the riff command, however, the difference is that the riff command is designed to be embedded into other command hierarchies whereas the root command should only be used by the riff cli.

pkg/riff/commands/riff_test.go
```diff
@@ -33,3 +37,43 @@ func TestRiffCommand(t *testing.T) {
 
 	table.Run(t, commands.NewRiffCommand)
 }
+
+func TestRiffSubCommands(t *testing.T) {
```
scothis: This test is fragile and not really necessary. We know which commands are registered because we have CI generate markdown docs from each command. These docs are committed to the repository and any drift between the generated and committed content will fail CI.

https://github.com/projectriff/cli/blob/e5a90713536175ecedaece277e0a98df49ca636e/Makefile#L44-L46

https://github.com/projectriff/cli/blob/e5a90713536175ecedaece277e0a98df49ca636e/azure-pipelines.yml#L21

zhitongLIU:
thank you for reviewing!

PR #1344, opened 2019-08-01
Setup docusaurus on v0.3.x branch

jldec:
v0.3.x branch PR for docusaurus
Will generate new website and docs for v0.3.x (stable) release
Details in [website/README.md](https://github.com/projectriff/riff/blob/b7c2529c154b5682dfabd09596a4633cf53a88d6/website/README.md)


jldec:
thank you for this feedback @scothis - a few responses below

> static and generated docs are conflated in the same directory, running make docs will drop the static docs

New docs from the projectriff/cli repo have to be copied over, and it's not hard to preserve the non-generated docs which are in differently named directories. I was reluctant to try to circumventing the assumptions that Docusaurus makes for the single docs [source path](https://docusaurus.io/docs/en/site-config#customdocspath-string).

> the full website content including nav, scripts, images, page and blog post are all included, but will create a high maintenance burden

I don't see how the maintenance burden would  change if other pages and images etc. lived in another repo - but if you prefer not to keep any docs in this riff repo, we could put everything in a dedicated website or docs repo. 

> this branch has knowledge of the other other branches. Adding a new branch will require updating all previous branches.

The list of versions in [siteconfig.js](https://github.com/projectriff/riff/pull/1344/files#diff-c272134cac3612b41f93893bbaa34d24R22-R25) is used to generate the `/versions` page which  is published with (only) the latest stable branch docs build. Other branches contribute only the `/docs/x-x/...` pages for that branch.

> generated content here still needs to get over to the projectriff.io repo

I think you're suggesting that committing content changes on release or master branches should trigger an automated docs build and automated re-publishing of the docs for that branch. I agree that thiis would be nice, especially if we could automate preview builds on PRs (with say Netlify). 

> While I do like the idea of versioning docs closer to their source, and that's true for the 0.3.x branch, it's not true for master. I only see this current pattern as creating more work with no discernible gain. If we think there is an advantage to having the non-cli docs in this repo, only the source markdown should be included here, everything else belongs in the website repo.

The benefit (imo) of including docs in the main riff repo is that it gives github users another way to view those docs in the context of the "main" repo, without first searching for another docs or website repo, and (afaik), github search will index the docs from the main branch. 

Splitting the docs markdown repo from the website repo would create additional maintenance burden because of cross dependencies e.g. images in the markdown and nav links to docs pages in the sidebar.

jldec:
closing in favor of https://github.com/projectriff/projectriff.io/pull/91

PR #1359, opened 2019-11-13
RFC-0001: CLI support for publishing and subscribing to streams

sbawaska:
A proposal to add commands to riff cli that enable us to publish and subscribe events to/from a stream.


RFCs/RFC-001: Introspect streams.md
```diff
@@ -0,0 +1,62 @@
...
+## Problem
+When the streaming runtime is installed, riff users can create streams (backed by a stream provider e.g. kafka). Users can write functions that read from and write to these streams,
+but there is no easy mechanism to get events into the stream and look into the contents of the stream which are helpful while developing/demoing.
+
+### Anti-Goals
+We will only address this problem for development/demos, not production, so topics like auth/authz are out of scope for this document.
+
+## Solution
+### Ideal solution:
+The  is to provide users with only one cli (riff cli) that will enable users to send/receive messages to/from a stream (commands below). We will have to ask the users to port-forward the gateway service if they are running outside the cluster.
```
jldec: The `<idea>` ?

RFCs/RFC-001: Introspect streams.md
```diff
@@ -0,0 +1,62 @@
+# RFC-001: Introspect Streams
```
jldec: How about naming this `minimal CLI support for writing and monitoring streams`

RFCs/RFC-001: Introspect streams.md
```diff
@@ -0,0 +1,62 @@
...
+### Anti-Goals
+We will only address this problem for development/demos, not production, so topics like auth/authz are out of scope for this document.
+
+## Solution
+### Ideal solution:
+The  is to provide users with only one cli (riff cli) that will enable users to send/receive messages to/from a stream (commands below). We will have to ask the users to port-forward the gateway service if they are running outside the cluster.
+
+#### User Impact
+We should introduce the following commands to the riff cli:
+- send a massage to the stream:
```
fbiville:
```suggestion
- send a message to the stream:
```

RFCs/RFC-001: Introspect streams.md
ericbottard: In the "ideal solution", why would the user have to manually do the complicated step of portforwarding? The ideal solution is seamless

RFCs/RFC-001: Introspect streams.md
```diff
@@ -0,0 +1,62 @@
...
+    The messages will be displayed as json with the following fields:
+    ``` 
+    {
+        "payload": "the users payload",
+        "content-type": "the content type of the message",
+        "headers": {"user provided header": "while publishing"}
+    }
+    ```
+
+### Interim solution:
```
ericbottard: I'm really not at ease with the concept of interim solution vs ideal solution.
Either there is value to "solution B", which actually solves "problem B" and it should have its own RFC, including explanations about how problem B is different from problem A, or there is only one problem, and maybe several solutions, for which the RFC author can have (sometimes subjective) feelings about which is better than the other.

I think we fall in the former case, whereby @markfisher in particular sees a need for the low level binary, to be used in-cluster. Let's hear the usecases for that first, and then we can discuss the higher level solution (which I feel more familiar with). The "riff cli" solution may or may not build on this one, we'll see when we state the problem space then.

RFCs/RFC-001: Introspect streams.md
```diff
@@ -0,0 +1,62 @@
...
+    content type has already been specified while creating the stream.
+
+- subscribe for messages from a stream:
+
+    `riff streaming stream subscribe <stream-name> --offset <long-offset>`  
+    this will display all the message from the given offset and then block to display further messages, essentially preventing user from entering subsequent commands.
+    The messages will be displayed as json with the following fields:
+    ``` 
+    {
+        "payload": "the users payload",
```
ericbottard: Why not use JSON (it allows piping in jq for example), but should then clarify how the payload is displayed. The raw payload is bytes. How should that be displayed in JSON? Should we always output the whole payload?

RFCs/RFC-001: Introspect streams.md
```diff
@@ -0,0 +1,62 @@
...
+## Solution
+### Ideal solution:
+The  is to provide users with only one cli (riff cli) that will enable users to send/receive messages to/from a stream (commands below). We will have to ask the users to port-forward the gateway service if they are running outside the cluster.
+
+#### User Impact
+We should introduce the following commands to the riff cli:
+- send a massage to the stream:
+
+    `riff streaming stream publish <stream-name> --payload <payload-as-string> --headers <headers-as-string>`  
+    content type has already been specified while creating the stream.
```
ericbottard: The concrete concrete content-type of a particular piece of payload may be different (a compatible subtype of) from the stream content-type.

RFCs/RFC-001: Introspect streams.md
```diff
@@ -0,0 +1,62 @@
...
+
+Steps:
+1. Move the existing client.go and client_test.go to `pkg` folder. 
+1. create `cmd/publish/main.go` for the publish binary
+1. create `cmd/subscribe/main.go` for the subscribe binary.
+1. setup CI to publish the binaries.
+
+#### User Impact
+The users will download the publish and subscribe binaries and run them by passing positional parameters:  
+- `$ streampublish <gateway-address> <topic-name> <payload-as-string> <key> <headers> <content-type>`
```
ericbottard: Maybe allow the payload from stdin? Would allow piping.

RFCs/RFC-001: Introspect streams.md
scothis: adding commands to the `riff` cli seems out of alignment with the stated anti-goal of not being appropriate for production.

RFCs/RFC-001: Introspect streams.md
scothis: This is a good example of why content-type is a misleading field name for Streams.

RFCs/RFC-001: Introspect streams.md
scothis: the payload would need to be base64 encoded in order to be encoded as a string in JSON. Or an array of numbers where each item in the array is a byte (big or little endian?).

RFCs/RFC-001: Introspect streams.md
sbawaska: fair point. I removed the "Interim Solution"

RFCs/RFC-001: Introspect streams.md
sbawaska: Yes! incorporated.

RFCs/RFC-001: Introspect streams.md
sbawaska: changed to: "CLI support for publishing and subscribing to streams"

RFCs/RFC-001: Introspect streams.md
sbawaska: This section is removed.

RFCs/RFC-001: Introspect streams.md
sbawaska: I believe that we should not ask the user to download two CLIs. We can mark the individual commands as "experimental".

RFCs/RFC-001: Introspect streams.md
sbawaska: Should the consumer care about the subtype? If not, may be we should just publish using the content-type on the stream. 
Let's start off without the sub-type being specified on this command for now. We can add it later if desired.

RFCs/RFC-001: Introspect streams.md
sbawaska: added another option `--payload-file <relative-filepath>` to the `publish` command. These will be base64 encoded so that subscribe will always be able to display strings.

RFCs/RFC-001: Introspect streams.md
sbawaska: Sure, we could setup port forwarding automatically, but I think that should be its own rfc which should also tackle the deployers from core and knative runtime.

ericbottard:
IMO "CLI support ..." is too ambiguous and could be understood to refer to "the" riff CLI. Maybe "command line tool ..." or something?

rfc/rfc-0001-cli-support-for-publishing-and-subscribing-to-streams.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Solution
+The idea is to provide users with only one cli (riff cli) that will enable users to send/receive messages to/from a stream (commands below). We will have to ask the users to port-forward the gateway service if they are running outside the cluster.
+
+#### User Impact
+We should introduce the following commands to the riff cli:
+- send a message to the stream:
+
+    `riff streaming stream publish <stream-name> --payload <payload-as-string> --headers <headers-as-string> --payload-file <relative-filepath>`  
+    content type has already been specified while creating the stream.
```
ericbottard: As stated earlier, the content-type of the payload is not necessarily exactly the same as the "acceptable" CT configured on the stream. Should be an option here

rfc/rfc-0001-cli-support-for-publishing-and-subscribing-to-streams.md
```diff
@@ -0,0 +1,51 @@
...
+    Options `payload` and `payload-file` are mutually exclusive. The former is to be used for strings and the later for binary payloads. The binary payload will be base64 encoded.
```
ericbottard: What's a "binary payload"?
I don't understand (and I don't think I ultimately agree) what it means that it "will be base64 encoded"

ericbottard: The former is for "inline payload", the latter is for payload coming from a file.
They're all bytes, ultimately. The former will be interpreted as a string (with the terminal's encoding, I assume). The latter is taken as-is

rfc/rfc-0001-cli-support-for-publishing-and-subscribing-to-streams.md
ericbottard: Not too keen on several headers set as one flag. Maybe use a repeated flag, as we do eg for `processor --input`

rfc/rfc-0001-cli-support-for-publishing-and-subscribing-to-streams.md
```diff
@@ -0,0 +1,51 @@
...
+
+- subscribe for messages from a stream:
+
+    `riff streaming stream subscribe <stream-name> --offset <long-offset>`
```
ericbottard: Can't support custom offset

rfc/rfc-0001-cli-support-for-publishing-and-subscribing-to-streams.md
```diff
@@ -0,0 +1,51 @@
...
+    Options `payload` and `payload-file` are mutually exclusive. The former is to be used for strings and the later for binary payloads. The binary payload will be base64 encoded.
+
+- subscribe for messages from a stream:
+
+    `riff streaming stream subscribe <stream-name> --offset <long-offset>`  
+    this will display all the message from the given offset and then block to display further messages, essentially preventing user from entering subsequent commands.
+    The messages will be displayed as json with the following fields:
+    ``` 
+    {
+        "payload": "the users payload",
```
ericbottard: Need to clarify how (and why) we would output "complicated payloads" (eg base64 encoding of them)

rfc/rfc-0001-cli-support-for-publishing-and-subscribing-to-streams.md
```diff
@@ -0,0 +1,51 @@
...
+
+    `riff streaming stream publish <stream-name> --payload <payload-as-string> --headers <headers-as-string> --payload-file <relative-filepath>`  
+    content type has already been specified while creating the stream.
+    Options `payload` and `payload-file` are mutually exclusive. The former is to be used for strings and the later for binary payloads. The binary payload will be base64 encoded.
+
+- subscribe for messages from a stream:
+
+    `riff streaming stream subscribe <stream-name> --offset <long-offset>`  
+    this will display all the message from the given offset and then block to display further messages, essentially preventing user from entering subsequent commands.
+    The messages will be displayed as json with the following fields:
```
ericbottard: How about supporting a custom template, with a sensible default?

ericbottard:
While this focuses on the UX (and that's ok), I think it would be good to clarify what the CLI is actually doing (ie port forwarding, connecting to liiklus, etc)

RFCs/RFC-001: Introspect streams.md
scothis: Why would you base64 encode the content of the file? At that point bytes is fine. The issue Eric raised above is that you can't have arbitrary bytes in JSON. Remove JSON, remove the binary encoding issue.

RFCs/RFC-001: Introspect streams.md
scothis: each message has its own content-type.

> may be we should just publish using the content-type on the stream

This furthers my point that the name content-type on the stream is confusing. It's not a single MIME type. It's more analogous to the HTTP Accept header.

ericbottard: The stream doesn't necessarily have a *concrete* content-type (although for the time being this is all we support because we implement "compatibility" as simple "equality").
A stream may have one or several mime-types, including wildcard ones, so you can't always use that as the content-type of the actual, concrete piece of data you publish.

RFCs/RFC-001: Introspect streams.md
scothis: There's a difference between something that is currently an experiment that expect to solidify, and something that is not intended for production uses.

Based on the problem statement, I continue to assert that the riff cli is not an appropriate home for this functionaltiy. It will be much easier down the road to add the functionality into the riff cli, then remove it from the cli.

sbawaska: Would it then suffice to remove the "not for production" from the anti-goal?

RFCs/RFC-001: Introspect streams.md
sbawaska: I think we should keep JSON, so that users that users can then process the output if desired. 

In order to display the payload as string in json, I was hoping to ensure that the payload is always published as a string. Hence, if you had binary content in a file, base64 encode it. I now realize that users functions will be publishing to the stream as well, so encoding while publishing will not work.

rfc/rfc-0001-cli-support-for-publishing-and-subscribing-to-streams.md
sbawaska: I was thinking that we will take the later also as a string (hence the encoding) so that it can be displayed on subscribe, but that does not consider other publishers, so I am scrapping this.

RFCs/RFC-001: Introspect streams.md
markfisher: I also think we should keep the riff CLI focused only on production usage (same reason I argued for removal of the "invoke" command).

I don't see any problem with separate (go-binary or shell) *commands* - not necessarily full "CLIs" themselves - for things like invoke, publish, and subscribe. We could also create an image that includes those commands in a single directory, so that it's easy to run within the cluster, e.g. `kubectl exec ... -- publish --payload ...`

sbawaska:
I am inclined to close this as "rejected" in favor of: https://github.com/projectriff/riff/pull/1362

sbawaska:
merging as Rejected in favor of #1362

PR #1360, opened 2019-11-15
RFC 0002 - Bindings

scothis:
Please submit comments by: 2019-11-20


rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
...
+
+As a binding is backed by a ConfigMap and Secret a mechanism is needed to specify which ConfigMap and Secret. [ObjectReferences](https://godoc.org/k8s.io/api/core/v1#ObjectReference) are commonly used to define the coordinates to an arbitrary resources. A binding reference consists of metadata ObjectReference and a secret ObjectReference.
+
+Any resource that produces a binding should include the binding reference on it's status, for example:
```
ericbottard:
```suggestion
Any resource that produces a binding should include the binding reference on its status, for example:
```

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
...
+      namespace: default
+      name: provider-binding-metadata
+    secretRef:
+      apiVersion: v1
+      kind: Secret
+      namespace: default
+      name: provider-binding-secret
+```
+
+It is common to consume multiple bindings within a single workload. Each binding must have a unique name for the workload. The binding name is the directory the metadata and secret volumes are mounted within. While the name has no impact on the binding per se, workloads may use the name to convey semantic meaning about the intent of the binding (e.g. a primary database vs a replica).
```
ericbottard: > While the name has no impact on the binding per se

Does it not? It seems indeed that the names space is flat *for all bindings* (even of different "types", ie databases and streams would use the same set of filenames under the well known directory base).

If the name does not bear any particular meaning, it seems that there is yet another level of indirection needed to convey the "meaning" of the binding. For example, I would be interested to know what a riff processor defined as
```
riff streaming processor create foo --function-ref f --input input1:stream1 --input input2:stream1 --output output1
```
would see as mounted info.

rfc/rfc-0002-bindings.md
scothis: Please wait until comments are solicited

rfc/rfc-0002-bindings.md
scothis: Please wait until comments are solicited

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
...
+
+**Status:** **Active** | Accepted | Dropped | Superseded
+
+**Pull Request URL:** #1360
+
+**Superseded by:** N/A
+
+**Supersedes:** N/A
+
+**Related:** N/A
```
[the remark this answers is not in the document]
scothis: Please wait until comments are solicited

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
...
+
+While a user may continue to use lower-level Kubernetes idioms, a higher-level experience can provide users a simpler and more maintainable experience.
+
+### Backwards Compatibility and Upgrade Path
+There are no direct backwards compatibility concerns as riff does not current provide any support for bindings. Cloud Foundry services are exposed to application via the `VCAP_SERVICES` environment variable which is a different structure than the CNB Bindings spec.
+
+It is expected that the bindings space will actively evolve within the Kubernetes ecosystem. As the space matures this RFC will almost certainly be superseded to align with the broader community.
+
+## FAQ
+*Answers to common questions that may arise and those that you’ve commonly been asked after requesting comments for this proposal.*
```
trisberg: We should mention that, as of now, there are no libraries that users can depend on to map the bindings to their applications. They would have to write custom code to utilize the provided binding information.

rfc/rfc-0002-bindings.md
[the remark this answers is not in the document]
scothis: It's intentionally vague as we won't know the universe of resources that may be bindable.

One concrete example would be the Stream resource, as it has coordinates like the gateway address and topic (and someday credentials as well). As the Stream controller reconciles each Stream, it can create the binding and hold a reference to that binding on its status for consumers to discover. See https://github.com/projectriff/system/pull/181 as an early example.

I'm intentionally not proposing a Binding CRD, although I'm also not precluding such a resource from existing in the future that implements this duck type interface.

rfc/rfc-0002-bindings.md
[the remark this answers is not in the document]
scothis: That's up to the workload. The workload may:
- require a binding with a specific name
- require a binding of a specific kind
- react opportunistically to a binding
- ignore all bindings

The name is the only part of the binding that can be directly defined by the consumer. Every other aspect of the binding comes directly from the ConfigMap/Secret volumes.

rfc/rfc-0002-bindings.md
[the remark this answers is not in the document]
trisberg: It says "should", I assume that if there are resources that are used for binding and are not under our control, then they won't necessarily have this status.

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
...
+      kind: Secret
+      namespace: default
+      name: provider-binding-secret
+```
+
+It is common to consume multiple bindings within a single workload. Each binding must have a unique name for the workload. The binding name is the directory the metadata and secret volumes are mounted within. While the name has no impact on the binding per se, workloads may use the name to convey semantic meaning about the intent of the binding (e.g. a primary database vs a replica).
+
+Note: The ConfigMap and Secret for the binding must be in the same namespace as the workload Pod, as Kubernetes does not support mounting resources defined in other namespaces.
+
+The reconcile for the consumer resource is responsible for mapping the bindings to volume mounts within the pod.
```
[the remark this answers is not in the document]
scothis: This is also intentionally vague as we won't know the universe of potential resources that consume bindings.

In the above example of a Stream that produces a Binding, the consumer could be the Processor, or it could also be source that needs to write to the Stream.

A Processor may also have other kinds of bindings, for example a reference dataset.

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
...
+
+kpack, as of v0.0.5, does not provide any mechanism for injecting volumes into builds, or any other means to satisfy the CNB Bindings spec. Further work will be required for proper integration of bindings at buildtime.
+
+### User Impact
+Users are not required to use binding. They remain free to use any other mechanism provided by Kubernetes to discover and/or inject credentials into their workloads. riff resources like Deployers and Processors currently contain a PodSpec where custom volumes, environment variables and arguments may be defined.
```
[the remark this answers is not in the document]
scothis: That may be a better question for RFC-0000.

I took it to mean a "Cody" who is interacting with the riff CRDs.

rfc/rfc-0002-bindings.md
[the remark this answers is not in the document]
scothis: > I assume that if there are resources that are used for binding and are not under our control, then they won't necessarily have this status

I'm not trying to solve for the world at this point. We need to start somewhere and once we have a working foundation build out way out.

rfc/rfc-0002-bindings.md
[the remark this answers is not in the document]
scothis: How is rfc-0001 related? I don't see any coupling between the two RFCs

rfc/rfc-0002-bindings.md
scothis: Added an FAQ question on this topic

> This RFC does not prescribe a mechanism for discovering bindings, it is assumed that the consumer has foreknowledge. Another RFC may define a higher level experience.

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,107 @@
...
+
+riff should adopt CNB Bindings spec for bindings at both build and runtime.
+
+Kubernetes uses [Volumes](https://kubernetes.io/docs/concepts/storage/volumes/) to inject files from an external source into a running container. While there is a diverse spectrum of volume implementations available, in order to maximize both compatibility with bog standard Kubernetes and simplify the implementation, we should focus on a [configMap volume](https://kubernetes.io/docs/concepts/storage/volumes/#configmap) for binding metadata and a [secret volume](https://kubernetes.io/docs/concepts/storage/volumes/#secret) for binding secrets, which map directly to Kubernetes [ConfigMaps](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/) and [Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) respectively.
+
+As a binding is backed by a ConfigMap and Secret a mechanism is needed to specify which ConfigMap and Secret. [ObjectReferences](https://godoc.org/k8s.io/api/core/v1#ObjectReference) are commonly used to define the coordinates to an arbitrary resources. A binding reference consists of metadata ObjectReference and a secret ObjectReference.
```
matthewmcnew: Why a LocalObjectReference vs a ObjectReference?

matthewmcnew: It seems like it would parallel a [EnvFromSource](https://godoc.org/k8s.io/api/core/v1#EnvFromSource) with the secret/configmaps scoped in the same namespace as the provider/consumer?

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,107 @@
...
+      name: provider-binding-secret
+```
+
+It is common to consume multiple bindings within a single workload. Each binding must have a unique name for the workload. The binding name is the directory the metadata and secret volumes are mounted within. While the name has no impact on the binding per se, workloads may use the name to convey semantic meaning about the intent of the binding (e.g. a primary database vs a replica).
+
+Note: The ConfigMap and Secret for the binding must be in the same namespace as the workload Pod, as Kubernetes does not support mounting resources defined in other namespaces.
+
+The reconciler for the resource consuming a binding is responsible for mapping the ConfigMap and Secret to volume mounts within the pod.
+
+kpack, as of v0.0.5, does not provide any mechanism for injecting volumes into builds, or any other means to satisfy the CNB Bindings spec. Further work will be required for proper integration of bindings at buildtime.
```
matthewmcnew: kpack should provide a comparable consumer binding interface.

rfc/rfc-0002-bindings.md
scothis: I don't want the text of this RFC to get into the specifics of any particular type as that would belong to follow on RFC, but you are correct that the service binding will likely not contain enough metadata for a Process to consume Streams as bindings, and that's ok.

I wouldn't expect the Processor CRD to change initially. Since the Processor reconciler is already aware of Streams, it can continue to be aware of Streams to provide an experience higher than the raw bindings. Under the hood, the processor sidecar can switch from its current scheme to use the CNB binding format with binding published for each Stream by the Stream reconciler.

It would be possible to add some encoding scheme into the binding name, but that probably doesn't provide a good experience. Specifically for Processors, what I would initially propose is that the binding name represent the stream alias, the INPUTS, OUTPUTS and OUTPUT_CONTENT_TYPES move into the binding. INPUT_NAMES and OUTPUT_NAMES continue as envvar that conveys the order and input/output categorization.

Another step could open up the function container to accept binding. This would augment the current Processor CRD.
```

File: rfc/rfc-0002-bindings.md (hunk 4 above)
Comment (scothis) [reply rfc/rfc-0002-bindings.md]: Added an FAQ

> It's anticipated that frameworks and libraries will manage the binding within the container without the developer needing to directly read from the filesystem. This RFC has no opinion how a container should consume the binding metadata or secret. At this time, there are no known such libraries. It would be logical for riff function invokers to adopt an appropriate library to provide a higher level experience to function authors.

File: rfc/rfc-0002-bindings.md (hunk 7 above)
Comment (scothis) [reply rfc/rfc-0002-bindings.md]: A LocalObjectReference would work, and would make the binding reference a good bit less verbose. We could broaden back out to an ObjectReference if we ever need the extra flexibility.

File: rfc/rfc-0002-bindings.md (hunk 8 above)
Comment (scothis) [reply rfc/rfc-0002-bindings.md]: We'd like to collaborate with the kpack team on how the Image resource consumes bindings.

**Comment by ojhughes** [comment]
It might be useful to reference [Service Catalog ](https://svc-cat.io/docs/resources/#servicebinding) bindings and any similarities or differences here

File: rfc/rfc-0002-bindings.md [hunk 9]
```diff
@@ -0,0 +1,95 @@
...
+
+riff should adopt CNB Bindings spec for bindings at both build and runtime.
+
+Kubernetes uses [Volumes](https://kubernetes.io/docs/concepts/storage/volumes/) to inject files from an external source into a running container. While there is a diverse spectrum of volume implementations available, in order to maximize both compatibility with bog standard Kubernetes and simplify the implementation, we should focus on a [configMap volume](https://kubernetes.io/docs/concepts/storage/volumes/#configmap) for binding metadata and a [secret volume](https://kubernetes.io/docs/concepts/storage/volumes/#secret) for binding secrets, which map directly to Kubernetes [ConfigMaps](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/) and [Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) respectively.
```
sbawaska:
```suggestion
Kubernetes uses [Volumes](https://kubernetes.io/docs/concepts/storage/volumes/) to inject files from an external source into a running container. While there is a diverse spectrum of volume implementations available, in order to maximize both compatibility with standard Kubernetes and simplify the implementation, we should focus on a [configMap volume](https://kubernetes.io/docs/concepts/storage/volumes/#configmap) for binding metadata and a [secret volume](https://kubernetes.io/docs/concepts/storage/volumes/#secret) for binding secrets, which map directly to Kubernetes [ConfigMaps](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/) and [Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) respectively.
```
remove the word `bog`

scothis:
@ojhughes added an FAQ item:

> *How do CNB Bindings relate to a [Service Catalog ServiceBinding](https://svc-cat.io/docs/resources/#servicebinding)?*
>
>Service Catalog is a client to an Open Service Broker. A Service Catalog ServiceBinding is a mechanism to bind a provisioned "service" to the cluster by exposing the credentials as a Secret. CNB Bindings have no mechanism to create "service" instances, instead, its focus is binding metadata and secrets to a workload. The output of the ServiceBinding is not directly consumable by CNB Bindings. A mechanism to convert a ServiceBinding to a format consumable by CNB Bindings is out of scope for this RFC.

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,95 @@
...
+
+As a binding is backed by a ConfigMap and Secret a mechanism is needed to specify which ConfigMap and Secret. [LocalObjectReferences](https://godoc.org/k8s.io/api/core/v1#LocalObjectReference) are commonly used to define the coordinates to an arbitrary resources. A binding reference consists of `metadata` LocalObjectReference to a ConfigMap and a `secret` LocalObjectReference to a Secret.
+
+Any resource that provides a binding should include the binding reference on its status, for example:
```
sbawaska: Who has the responsibility for making the resource adhere to this requirement? I do not see us doing it, which means it will be an additional step that our users will have to undertake.

It would be useful to list the steps that a user has to when when they have provisioned say a mysql using a helm chart, and wants to bind to it their function.

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
...
+
+Kubernetes uses [Volumes](https://kubernetes.io/docs/concepts/storage/volumes/) to inject files from an external source into a running container. While there is a diverse spectrum of volume implementations available, in order to maximize both compatibility with bog standard Kubernetes and simplify the implementation, we should focus on a [configMap volume](https://kubernetes.io/docs/concepts/storage/volumes/#configmap) for binding metadata and a [secret volume](https://kubernetes.io/docs/concepts/storage/volumes/#secret) for binding secrets, which map directly to Kubernetes [ConfigMaps](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/) and [Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) respectively.
+
+As a binding is backed by a ConfigMap and Secret a mechanism is needed to specify which ConfigMap and Secret. [LocalObjectReferences](https://godoc.org/k8s.io/api/core/v1#LocalObjectReference) are commonly used to define the coordinates to an arbitrary resources. A binding reference consists of `metadata` LocalObjectReference to a ConfigMap and a `secret` LocalObjectReference to a Secret.
```
fbiville:
```suggestion
As a binding is backed by a ConfigMap and Secret a mechanism is needed to specify which ConfigMap and Secret. [LocalObjectReferences](https://godoc.org/k8s.io/api/core/v1#LocalObjectReference) are commonly used to define the coordinates to an arbitrary resource. A binding reference consists of `metadata` LocalObjectReference to a ConfigMap and a `secret` LocalObjectReference to a Secret.
```

fbiville:
If that spec were to be merged in the current state, could the implementation of service binding start?

It seems to me that the spec currently describes what bindings are but now how they are to be used. It is likely that thinking about the usage early (in that spec) would help refine binding definitions as well. What do you think @scothis?

rfc/rfc-0002-bindings.md
scothis: It's encouraged for resources to publicize the binding reference, but is not required. It's an anti-goal of this RFC to define such a mechanism.

As there is no requirement, there is no enforcement mechanism. If desired, such a requirement/mechanism should be the focus of another RFC.

scothis:
> If that spec were to be merged in the current state, could the implementation of service binding start?

@fbiville see https://github.com/projectriff/system/pull/181, which predates the RFC

rfc/rfc-0002-bindings.md
```diff
@@ -91,3 +91,7 @@ Maybe. The name of the binding appears in the filesystem as the directory contai
 *How do consumers discover bindings?*
 
 This RFC does not prescribe a mechanism for discovering bindings, it is assumed that the consumer has foreknowledge. Another RFC may define a higher level experience.
+
+*How do CNB Bindings relate to a [Service Catalog ServiceBinding](https://svc-cat.io/docs/resources/#servicebinding)?*
+
+Service Catalog is a client to an Open Service Broker. A Service Catalog ServiceBinding is a mechanism to bind a provisioned "service" to the cluster by exposing the credentials as a Secret. CNB Bindings have no mechanism to create "service" instances, instead, its focus is binding metadata and secrets to a workload. The output of the ServiceBinding is not directly consumable by CNB Bindings. A mechanism to convert a ServiceBinding to a format consumable by CNB Bindings is out of scope for this RFC.
```
trisberg: Maybe add:
"The secrets created by the Service Catalog don't have a standardized format. It is left to the Open Service Broker to generate the content, so any mapping to CNB Bindings would have to be on a case-by-case basis."

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
...
+
+kpack, as of v0.0.5, does not provide any mechanism for injecting volumes into builds, or any other means to satisfy the CNB Bindings spec. Further work will be required for proper integration of bindings at buildtime.
+
+### User Impact
+This RFC does not require users to take any action. They remain free to use any other mechanism provided by Kubernetes to discover and/or inject credentials into their workloads. riff resources like Deployers and Processors currently contain a PodSpec where custom volumes, environment variables and arguments may be defined. Individual riff resources may start to produce and/or consume bindings, which users may then consume.
+
+While a user may continue to use lower-level Kubernetes idioms, a higher-level experience can provide users a simpler and more maintainable experience.
+
+### Backwards Compatibility and Upgrade Path
+There are no direct backwards compatibility concerns as riff does not current provide any support for bindings. Cloud Foundry services are exposed to application via the `VCAP_SERVICES` environment variable which is a different structure than the CNB Bindings spec.
```
ericbottard:
```suggestion
There are no direct backwards compatibility concerns as riff does not currently provide any support for bindings. Cloud Foundry services are exposed to application via the `VCAP_SERVICES` environment variable which is a different structure than the CNB Bindings spec.
```

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
...
+
+## FAQ
+*How do functions and applications consume the binding?*
+
+It's anticipated that frameworks and libraries will manage the binding within the container without the developer needing to directly read from the filesystem. This RFC has no opinion how a container should consume the binding metadata or secret. At this time, there are no known such libraries. It would be logical for riff function invokers to adopt an appropriate library to provide a higher level experience to function authors.
+
+*Does the binding name have any meaning?*
+
+Maybe. The name of the binding appears in the filesystem as the directory containing the `metadata` and `secret` directories that comprise the binding. A workload may take the name of a binding into considering when consuming the binding, or it may ignore it. The binding name is the only aspect of the binding that is under control of the consumer. All other elements of the binding are defined by the binding producer. Additional metadata may be injected into the workload about the binding out of band, but is not part of this RFC.
```
ericbottard:
```suggestion
Maybe. The name of the binding appears in the filesystem as the directory containing the `metadata` and `secret` directories that comprise the binding. A workload may take the name of a binding into consideration when consuming the binding, or it may ignore it. The binding name is the only aspect of the binding that is under control of the consumer. All other elements of the binding are defined by the binding producer. Additional metadata may be injected into the workload about the binding out of band, but is not part of this RFC.
```

rfc/rfc-0002-bindings.md
scothis: CNB Bindings also don't define any structure for the secret. So no problem there.

rfc/rfc-0002-bindings.md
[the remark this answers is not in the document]
ericbottard: nvm, from a quick glance of the title of 0001, I thought it was the RFC we initially discussed about https://github.com/projectriff/system/pull/181

fbiville:
> > If that spec were to be merged in the current state, could the implementation of service binding start?
> 
> @fbiville see [projectriff/system#181](https://github.com/projectriff/system/pull/181), which predates the RFC

Why start with streams and not service bindings for riff applications?

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
...
+
+The binding contains both metadata and credentials to provide a workload the details it needs to connect and consume the resource.
+
+### Anti-Goals
+A generic mechanism to provision bindings, and the specific structure of the metadata and credentials are out of scope for this RFC.
+
+## Solution
+The Cloud Native Buildpacks project has a [spec for Bindings](https://github.com/buildpack/spec/blob/master/extensions/bindings.md) which defines the form bindings should take at both build and runtime. In short, each binding is represented as a directory inside a well known directory with two sub-directories: `metadata` for general non-sensitive information about the binding, and `secret` for credentials and other sensitive information. Each attribute for the binding is a file in one of those two directories. The metadata is available at buildtime and both the metadata and secret is available at runtime.
```
markfisher:
```suggestion
The Cloud Native Buildpacks project has a [spec for Bindings](https://github.com/buildpack/spec/blob/master/extensions/bindings.md) which defines the form bindings should take at both build and runtime. In short, each binding is represented as a directory inside a well known base directory, and each binding directory contains two sub-directories: `metadata` for general non-sensitive information about the binding, and `secret` for credentials and other sensitive information. Each attribute for the binding is a file in one of those two directories. The metadata is available at buildtime and both the metadata and secret are available at runtime.
```

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
...
+
+riff should adopt CNB Bindings spec for bindings at both build and runtime.
```
markfisher:
```suggestion
riff should adopt the CNB Bindings spec for bindings at both buildtime and runtime.
```

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
...
+
+Kubernetes uses [Volumes](https://kubernetes.io/docs/concepts/storage/volumes/) to inject files from an external source into a running container. While there is a diverse spectrum of volume implementations available, in order to maximize both compatibility with standard Kubernetes and simplify the implementation, we should focus on a [configMap volume](https://kubernetes.io/docs/concepts/storage/volumes/#configmap) for binding metadata and a [secret volume](https://kubernetes.io/docs/concepts/storage/volumes/#secret) for binding secrets, which map directly to Kubernetes [ConfigMaps](https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/) and [Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) respectively.
+
+As a binding is backed by a ConfigMap and Secret a mechanism is needed to specify which ConfigMap and Secret. [LocalObjectReferences](https://godoc.org/k8s.io/api/core/v1#LocalObjectReference) are commonly used to define the coordinates to an arbitrary resource. A binding reference consists of `metadata` LocalObjectReference to a ConfigMap and a `secret` LocalObjectReference to a Secret.
```
markfisher:
```suggestion
As a binding is backed by a ConfigMap and Secret a mechanism is needed to specify which ConfigMap and Secret. [LocalObjectReferences](https://godoc.org/k8s.io/api/core/v1#LocalObjectReference) are commonly used to define the coordinates to an arbitrary resource. Therefore, a binding reference consists of a `metadata` LocalObjectReference to a ConfigMap and a `secret` LocalObjectReference to a Secret.
```

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
...
+
+Note: The ConfigMap and Secret for the binding must be in the same namespace as the workload Pod.
+
+The reconciler for the resource consuming a binding is responsible for mapping the ConfigMap and Secret to volume mounts within the pod.
+
+kpack, as of v0.0.5, does not provide any mechanism for injecting volumes into builds, or any other means to satisfy the CNB Bindings spec. Further work will be required for proper integration of bindings at buildtime.
+
+### User Impact
+This RFC does not require users to take any action. They remain free to use any other mechanism provided by Kubernetes to discover and/or inject credentials into their workloads. riff resources like Deployers and Processors currently contain a PodSpec where custom volumes, environment variables and arguments may be defined. Individual riff resources may start to produce and/or consume bindings, which users may then consume.
```
markfisher: what is meant by "consume" from the user's perspective here? might be confusing since "produce" and "consume" are used earlier in this sentence with more direct relevance to the preceding description of the feature

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
...
+
+While a user may continue to use lower-level Kubernetes idioms, a higher-level experience can provide users a simpler and more maintainable experience.
```
markfisher: this wording seems tautological:

> higher-level experience can provide... experience

scothis:
> Why start with streams and not service bindings for riff applications?

@fbiville this is out of scope for the RFC, but to answer your question. The idea to start with Streams is because the problem space contains both the production of a binding (via the Stream resource) and the consumption of the binding (via the Processor resource and sources). One of the problems inhibiting bindings is that there is a network effect as more components use bindings. Creating bindings without consumers is pointless. Without producers, consuming bindings is impossible.

There's nothing to stop applications from also consuming bindings today, other than the lack of producers. The UX is meh, but it's possible. I expect to see a further RFC to refine the UX.

Again, anything related to a specific producer or consumer is out of scope for this RFC.

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Backwards Compatibility and Upgrade Path
+There are no direct backwards compatibility concerns as riff does not currently provide any support for bindings. Cloud Foundry services are exposed to application via the `VCAP_SERVICES` environment variable which is a different structure than the CNB Bindings spec.
```
markfisher:
```suggestion
There are no direct backwards compatibility concerns as riff does not currently provide any support for bindings. Cloud Foundry services are exposed to applications via the `VCAP_SERVICES` environment variable which is a different structure than the CNB Bindings spec.
```

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
...
+
+*Does the binding name have any meaning?*
+
+Maybe. The name of the binding appears in the filesystem as the directory containing the `metadata` and `secret` directories that comprise the binding. A workload may take the name of a binding into consideration when consuming the binding, or it may ignore it. The binding name is the only aspect of the binding that is under control of the consumer. All other elements of the binding are defined by the binding producer. Additional metadata may be injected into the workload about the binding out of band, but is not part of this RFC.
+
+*How do consumers discover bindings?*
+
+This RFC does not prescribe a mechanism for discovering bindings, it is assumed that the consumer has foreknowledge. Another RFC may define a higher level experience.
```
markfisher:
```suggestion
This RFC does not prescribe a mechanism for discovering bindings. It is assumed that the consumer has foreknowledge. Another RFC may define a higher level experience.
```

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
...
+
+## FAQ
+*How do functions and applications consume the binding?*
+
+It's anticipated that frameworks and libraries will manage the binding within the container without the developer needing to directly read from the filesystem. This RFC has no opinion how a container should consume the binding metadata or secret. At this time, there are no known such libraries. It would be logical for riff function invokers to adopt an appropriate library to provide a higher level experience to function authors.
+
+*Does the binding name have any meaning?*
+
+Maybe. The name of the binding appears in the filesystem as the directory containing the `metadata` and `secret` directories that comprise the binding. A workload may take the name of a binding into consideration when consuming the binding, or it may ignore it. The binding name is the only aspect of the binding that is under control of the consumer. All other elements of the binding are defined by the binding producer. Additional metadata may be injected into the workload about the binding out of band, but is not part of this RFC.
```
markfisher:
```suggestion
Maybe. The name of the binding appears in the filesystem as the directory containing the `metadata` and `secret` directories that comprise the binding. A workload may take the name of a binding into consideration when consuming the binding, or it may ignore it. The binding name is the only aspect of the binding that is under the control of the consumer. All other elements of the binding are defined by the binding producer. Additional metadata may be injected into the workload about the binding out of band, but is not part of this RFC.
```

rfc/rfc-0002-bindings.md
scothis:
```suggestion
This RFC does not require users to take any action. They remain free to use any other mechanism provided by Kubernetes to discover and/or inject credentials into their workloads. riff resources like Deployers and Processors currently contain a PodSpec where custom volumes, environment variables and arguments may be defined. Individual riff resources may start to produce and/or consume bindings.
```

rfc/rfc-0002-bindings.md
scothis:
```suggestion
While a user may use lower-level binding ConfigMap and Secret resources directly, higher-level user experience are out of scope for this RFC and should be the focus of future RFCs.
```

markfisher: should be "higher-level user experience**s**" (or "**a** higher-level..." and change "are" to "is")

scothis:
> How could we align Crossplane's Secret propagation with this model?

With the standard caveat that the discussion is out of scope for this RFC:

The gotcha here (unless I'm mistaken) is that the secret created by Crossplane only covers the CNB Binding secret fields, and does not contain the CNB Binding metadata (kind, provider, tags, etc). The ResourceClaim has that metadata. So the trick is to coerce Crossplane into creating a ConfigMap from the ResourceClaim along side the Secret.

rfc/rfc-0002-bindings.md
```diff
@@ -0,0 +1,97 @@
+# RFC-0002: Bindings
+
+**Authors:** @scothis
+
+**Status:**
```
scothis:
```suggestion
**Status:** Approved
```

PR #1362, opened 2019-11-21
rfc-0003: A pod with utilities for better developer experience

sbawaska:
supersedes rfc-0001.

### DIFF
--- a/rfc/rfc-0003-developer-utils.md
+++ b/rfc/rfc-0003-developer-utils.md
@@ -46,9 +46,9 @@ The command takes the form:
 1. **subscribe:** To subscribe for events from the given stream.
 The command takes the form:
     ```
-    subscribe <stream-name> --payload-as-string --from-latest
+    subscribe <stream-name> --payload-as-string --from-beginning
     ```
-    If the `--from-latest` option is present, displays new events, otherwise will display all the events in the stream in the following json format:
+    If the `--from-beginning` option is present, display all the events in the stream, otherwise only new events are displayed in the following json format:
     ```
     {"payload": "base64 encoded user payload","content-type": "the content type of the message","headers": {"user provided header": "while publishing"}}
     ```
--- a/rfc/rfc-0003-developer-utils.md
+++ b/rfc/rfc-0003-developer-utils.md
@@ -1,16 +1,16 @@
 # RFC-#: A pod with utilities for better developer experience
 
-**Authors:**
+**Authors:** Swapnil Bawaskar
 
 **Status:**
 
-**Pull Request URL:** link
+**Pull Request URL:** https://github.com/projectriff/riff/pull/1362
 
 **Superseded by:** N/A
 
-**Supersedes:** N/A
+**Supersedes:** https://github.com/projectriff/riff/pull/1359
 
-**Related:** https://github.com/projectriff/riff/pull/1359 and 
+**Related:**
 
 
 ## Problem
--- a/rfc/rfc-0003-developer-utils.md
+++ b/rfc/rfc-0003-developer-utils.md
@@ -27,15 +27,15 @@ We will have the following commands to start with:
 1. **invoke-core:** To invoke the given core deployer.  
 The command takes the form:  
     ```
-    invoke-core <deployer-name> -n <namespace> -c "<curl-params>"
+    invoke-core <deployer-name> -n <namespace> -- <curl-params>
     ```
-    where `<curl-params>` are passed along to the post-request.
+    where everything after `--` is passed as parameter to curl
 1. **invoke-knative:** To invoke the given knative deployer.
 The command takes the form:  
     ```
-    invoke-knative <deployer-name> -n <namespace> -c "<curl-params>"
+    invoke-knative <deployer-name> -n <namespace> -- <curl-params>
     ```
-    where `<curl-params>` are passed along to the post-request.
+    where everything after `--` is passed as parameter to curl
 
 1. **publish:** To publish an event to the given stream.
 The command takes the form:
--- a/rfc/rfc-0003-developer-utils.md
+++ b/rfc/rfc-0003-developer-utils.md
@@ -37,16 +37,16 @@ The command takes the form:
     ```
     where `<curl-params>` are passed along to the post-request.
 
-1. **publish-stream:** To publish an event to the given stream.
+1. **publish:** To publish an event to the given stream.
 The command takes the form:
     ```
-    publish-stream <stream-name> -n <namespace> --payload <payload-as-string> --content-type <content-type> --header "<header-name>: <header-value>"
+    publish <stream-name> -n <namespace> --payload <payload-as-string> --content-type <content-type> --header "<header-name>: <header-value>"
     ```
     where `stream-name`, `payload` and `content-type` are mandatory and `header` can be used multiple times.
-1. **subscribe-stream:** To subscribe for events from the given stream.
+1. **subscribe:** To subscribe for events from the given stream.
 The command takes the form:
     ```
-    subscribe-stream <stream-name> --offset <long-offset> --payload-as-string
+    subscribe <stream-name> --offset <long-offset> --payload-as-string
     ```
     This will display all the events in the stream from the given offset in the following json format:
     ```
--- a/rfc/rfc-0003-developer-utils.md
+++ b/rfc/rfc-0003-developer-utils.md
@@ -1,4 +1,4 @@
-# RFC-#: A pod with utilities for better developer experience
+# RFC-0003: A pod with utilities for better developer experience
 
 **Authors:** Swapnil Bawaskar
 
--- a/rfc/rfc-0003-developer-utils.md
+++ b/rfc/rfc-0003-developer-utils.md
@@ -4,19 +4,19 @@
 
 **Status:**
 
-**Pull Request URL:** https://github.com/projectriff/riff/pull/1362
+**Pull Request URL:** [#1362](https://github.com/projectriff/riff/pull/1362)
 
 **Superseded by:** N/A
 
-**Supersedes:** https://github.com/projectriff/riff/pull/1359
+**Supersedes:** [RFC 0001](https://github.com/projectriff/riff/pull/1359)
 
 **Related:**
 
 
 ## Problem
-The function developer would like to invoke the function for which they just created the `core deployer` or the `knative deployer`.
-When the streaming runtime is installed, they would like to test the function by getting events into the stream and look into the contents of the stream. There is no one step mechanism for accomplishing any of these currently.
-An [earlier proposal](https://github.com/projectriff/riff/pull/1359) was to provide some of this functionality in the riff cli, this proposal expands on Mark's comment https://github.com/projectriff/riff/pull/1359#discussion_r348617981
+A developer would like to invoke the function for which they just created the `core deployer` or the `knative deployer`.
+When the streaming runtime is installed, they would like to test the processor by getting events into the stream and look into the contents of the stream. There is no one step mechanism for accomplishing any of these currently.
+An [earlier proposal](https://github.com/projectriff/riff/pull/1359) was to provide some of this functionality in the riff cli, this proposal expands on [Mark's comment](https://github.com/projectriff/riff/pull/1359#discussion_r348617981).
 
 ### Anti-Goals
 We will only address this problem for development/demos, not production, so topics like auth/authz are out of scope for this document.
@@ -40,9 +40,9 @@ The command takes the form:
 1. **publish-stream:** To publish an event to the given stream.
 The command takes the form:
     ```
-    publish-stream <stream-name> -n <namespace> --payload <payload-as-string> --content-type <content-type> --header <headers-as-string>
+    publish-stream <stream-name> -n <namespace> --payload <payload-as-string> --content-type <content-type> --header "<header-name>: <header-value>"
     ```
-    where `stream-name`, `payload` and `content-type` are mandatory and `headers` can be used multiple times.
+    where `stream-name`, `payload` and `content-type` are mandatory and `header` can be used multiple times.
 1. **subscribe-stream:** To subscribe for events from the given stream.
 The command takes the form:
     ```
--- a/rfc/rfc-0003-developer-utils.md
+++ b/rfc/rfc-0003-developer-utils.md
@@ -46,9 +46,9 @@ The command takes the form:
 1. **subscribe:** To subscribe for events from the given stream.
 The command takes the form:
     ```
-    subscribe <stream-name> --payload-as-string
+    subscribe <stream-name> --payload-as-string --from-latest
     ```
-    This will display all the events in the stream in the following json format:
+    If the `--from-latest` option is present, displays new events, otherwise will display all the events in the stream in the following json format:
     ```
     {"payload": "base64 encoded user payload","content-type": "the content type of the message","headers": {"user provided header": "while publishing"}}
     ```
--- a/rfc/rfc-0003-developer-utils.md
+++ b/rfc/rfc-0003-developer-utils.md
@@ -27,13 +27,13 @@ We will have the following commands to start with:
 1. **invoke-core:** To invoke the given core deployer.  
 The command takes the form:  
     ```
-    invoke-core <deployer-name> -n <namespace> <curl-params>
+    invoke-core <deployer-name> -n <namespace> -c "<curl-params>"
     ```
     where `<curl-params>` are passed along to the post-request.
 1. **invoke-knative:** To invoke the given knative deployer.
 The command takes the form:  
     ```
-    invoke-knative <deployer-name> -n <namespace> <curl-params>
+    invoke-knative <deployer-name> -n <namespace> -c "<curl-params>"
     ```
     where `<curl-params>` are passed along to the post-request.
 
--- a/rfc/rfc-0003-developer-utils.md
+++ b/rfc/rfc-0003-developer-utils.md
@@ -14,59 +14,71 @@
 
 
 ## Problem
-A developer would like to invoke the function for which they just created the `core deployer` or the `knative deployer`.
-When the streaming runtime is installed, they would like to test the processor by getting events into the stream and look into the contents of the stream. There is no one step mechanism for accomplishing any of these currently.
+A developer would like to interact with riff resources inside the cluster in ways that may not be desirable to expose outside of the cluster. This may be reading or writing message from a stream or invoking an http workload whose ingress policy is cluster-local. There is no one step mechanism for accomplishing any of these currently.
+
 An [earlier proposal](https://github.com/projectriff/riff/pull/1359) was to provide some of this functionality in the riff cli. The current proposal is a response to [feedback](https://github.com/projectriff/riff/pull/1359#discussion_r348617981) on that earlier proposal.
 
 ### Anti-Goals
 We will only address this problem for development/demos, not production, so topics like auth/authz are out of scope for this document.
 
 ## Solution
-We will ask the developers to run a `riff-developer-utils` pod in their development cluster. This pod will bundle a few binaries/scripts that users can invoke using `kubectl exec`. Since the commands will be run in-cluster, users won't need to run `kubectl port-forward`.
+We will ask the developers to run a `riff-dev` pod in their development cluster. This pod will bundle a few commands that users can invoke using `kubectl exec`. Since the commands will be run in-cluster, users won't need to run `kubectl port-forward`, however, the pod will need to be run with a service account that has appropriate RBAC permissions.
+
 We will have the following commands to start with:
-1. **invoke-core:** To invoke the given core deployer.  
-The command takes the form:  
-    ```
-    invoke-core <deployer-name> -n <namespace> -- <curl-params>
-    ```
-    where everything after `--` is passed as parameter to curl
-1. **invoke-knative:** To invoke the given knative deployer.
-The command takes the form:  
-    ```
-    invoke-knative <deployer-name> -n <namespace> -- <curl-params>
-    ```
-    where everything after `--` is passed as parameter to curl
 
 1. **publish:** To publish an event to the given stream.
-The command takes the form:
+    
+    The command takes the form:
+    
     ```
-    publish <stream-name> -n <namespace> --payload <payload-as-string> --content-type <content-type> --header "<header-name>: <header-value>"
+    publish <stream-name> --content-type <content-type> [--payload <payload-as-plain-text>] [--payload-base64 <payload-as-base64-text>] [--header "<header-name>: <header-value>"]
     ```
-    where `stream-name`, `payload` and `content-type` are mandatory and `header` can be used multiple times.
-1. **subscribe:** To subscribe for events from the given stream.
-The command takes the form:
-    ```
-    subscribe <stream-name> --payload-as-string --from-beginning
-    ```
-    If the `--from-beginning` option is present, display all the events in the stream, otherwise only new events are displayed in the following json format:
-    ```
-    {"payload": "base64 encoded user payload","content-type": "the content type of the message","headers": {"user provided header": "while publishing"}}
-    ```
-    The payload will be base64 encoded unless the `--payload-as-string` flag is present, in which case it will be displayed as a string.
+    
+    where `stream-name`, and `--content-type` are mandatory and `--header` can be used multiple times. `--payload` and `--payload-base64` are mutually exclusive. Binary data should only be passed via `--payload-base64` after it has been `base64` encoded.
 
-The namespace parameter is optional for all the commands. If not specified, the namespace of the `riff-developer-utils` pod will be assumed.
+1. **subscribe:** To subscribe for events from the given stream.
+    The command takes the form:
+    
+    ```
+    subscribe <stream-name> [--from-beginning]
+    ```
+    
+    If the `--from-beginning` option is present, display all the events in the stream, otherwise only new events are displayed as [JSON Lines](http://jsonlines.org) in the following form:
+    
+    ```
+    {"payload":"base64 encoded message payload","content-type":"the content type of the message","headers":{"header name": "header value"}}
+    {"payload":"base64 encoded message payload","content-type":"the content type of the message","headers":{"header name": "header value"}}
+    ```
+    
+    Since the payload may contain binary data, it will be displayed as a base64 encoded string. Tools like `jq` and `base64 --decode` are useful to get a raw payload.
+    
+    The subscribe command will run until terminated (SIGTERM), or an error occurs on the stream.
+
+1. [**jq**](https://stedolan.github.io/jq/): Process JSON.
+
+1. [**base64:**](http://manpages.ubuntu.com/manpages/bionic/man1/base64.1.html) Encode and decode base64 strings.
+
+1. [**curl:**](https://curl.haxx.se) Make HTTP requests.
+
+Each command targets resources in the same namespace as the pod is running. Additional commands and behaviors may be defined by future RFCs.
 
 ### User Impact
-The user will be able to able to invoke functions/deployers and publish and subscribe to streams using the following workflow:
-1. Run the `riff-developer-utils` pod in their cluster with something like:
+The user will be able to run these using the following workflow:
+
+1. Run the `riff-dev` pod in their cluster with something like:
+    
     ```
-    kubectl run --image=projectriff/developer-utils
-    ```
-1. Invoke the commands specified above using kubectl exec. An example follows:
-    ```
-    kubectl exec -it riff-developer-utils -- invoke-core upper -d test
+    # the pod should be run with a service account that has appropriate access to the api server
+    kubectl run riff-dev --image=<image>
     ```
 
+1. Invoke the commands specified above using kubectl exec. An example follows:
+    
+    ```
+    kubectl exec -it riff-dev -- <command to run inside pod with args>
+    ```
+
+The implementation should recommend specific RBAC policies needed to use the commands in the pod as well as specific commands and arguments.
 
 ### Backwards Compatibility and Upgrade Path
 Net new functionality, no impact on backward compatibility.
--- a/rfc/rfc-0003-developer-utils.md
+++ b/rfc/rfc-0003-developer-utils.md
@@ -46,15 +46,11 @@ The command takes the form:
 1. **subscribe-stream:** To subscribe for events from the given stream.
 The command takes the form:
     ```
-    riff streaming stream subscribe <stream-name> --offset <long-offset>
+    subscribe-stream <stream-name> --offset <long-offset>
     ```
     This will display all the events in the stream from the given offset in the following json format:
     ```
-    {
-        "payload": "base64 encoded user payload",
-        "content-type": "the content type of the message",
-        "headers": {"user provided header": "while publishing"}
-    }
+    {"payload": "base64 encoded user payload","content-type": "the content type of the message","headers": {"user provided header": "while publishing"}}
     ```
     We will use the content-type of the event to determine how the payload should be displayed. i.e. for content-type "text/plain" and "application/json" convert the binary payload to string, for all others display base64 encoded string.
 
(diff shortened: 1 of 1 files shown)


rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,81 @@
...
+
+**Authors:** Swapnil Bawaskar
+
+**Status:**
+
+**Pull Request URL:** https://github.com/projectriff/riff/pull/1362
+
+**Superseded by:** N/A
+
+**Supersedes:** https://github.com/projectriff/riff/pull/1359
```
scothis:
```suggestion
**Supersedes:** [RFC 0001](https://github.com/projectriff/riff/pull/1359)
```

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,81 @@
+# RFC-#: A pod with utilities for better developer experience
+
+**Authors:** Swapnil Bawaskar
+
+**Status:**
+
+**Pull Request URL:** https://github.com/projectriff/riff/pull/1362
```
scothis:
```suggestion
**Pull Request URL:** [#1362](https://github.com/projectriff/riff/pull/1362)
```

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,81 @@
...
+
+**Superseded by:** N/A
+
+**Supersedes:** https://github.com/projectriff/riff/pull/1359
+
+**Related:**
+
+
+## Problem
+The function developer would like to invoke the function for which they just created the `core deployer` or the `knative deployer`.
```
scothis: The deployers can also deploy applications and container images

```suggestion
A developer would like to invoke the function for which they just created the `core deployer` or the `knative deployer`.
```

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,81 @@
...
+When the streaming runtime is installed, they would like to test the function by getting events into the stream and look into the contents of the stream. There is no one step mechanism for accomplishing any of these currently.
```
scothis:
```suggestion
When the streaming runtime is installed, they would like to test the processor by getting events into the stream and look into the contents of the stream. There is no one step mechanism for accomplishing any of these currently.
```

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,81 @@
...
+An [earlier proposal](https://github.com/projectriff/riff/pull/1359) was to provide some of this functionality in the riff cli, this proposal expands on Mark's comment https://github.com/projectriff/riff/pull/1359#discussion_r348617981
```
scothis:
```suggestion
An [earlier proposal](https://github.com/projectriff/riff/pull/1359) was to provide some of this functionality in the riff cli, this proposal expands on [Mark's comment](https://github.com/projectriff/riff/pull/1359#discussion_r348617981).
```

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,81 @@
...
+### Anti-Goals
+We will only address this problem for development/demos, not production, so topics like auth/authz are out of scope for this document.
+
+## Solution
+We will ask the developers to run a `riff-developer-utils` pod in their development cluster. This pod will bundle a few binaries/scripts that users can invoke using `kubectl exec`. Since the commands will be run in-cluster, users won't need to run `kubectl port-forward`.
+We will have the following commands to start with:
+1. **invoke-core:** To invoke the given core deployer.  
+The command takes the form:  
+    ```
+    invoke-core <deployer-name> -n <namespace> <curl-params>
```
scothis: how is the deployer name resolved to a host name? A vanilla pod won't have permission

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,81 @@
...
+    ```
+    where `<curl-params>` are passed along to the post-request.
+1. **invoke-knative:** To invoke the given knative deployer.
+The command takes the form:  
+    ```
+    invoke-knative <deployer-name> -n <namespace> <curl-params>
```
scothis: Do we care about the knative adapter resource?

rfc/rfc-0003-developer-utils.md
scothis:
```suggestion
    invoke-core <deployer-name> -n <namespace> -- <curl-params>
```

rfc/rfc-0003-developer-utils.md
scothis:
```suggestion
    invoke-knative <deployer-name> -n <namespace> -- <curl-params>
```

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,81 @@
...
+    ```
+    where `<curl-params>` are passed along to the post-request.
+
+1. **publish-stream:** To publish an event to the given stream.
+The command takes the form:
+    ```
+    publish-stream <stream-name> -n <namespace> --payload <payload-as-string> --content-type <content-type> --header <headers-as-string>
```
scothis: How are header names and values distinguished? Following curl we could do something like:
```suggestion
    publish-stream <stream-name> -n <namespace> --payload <payload-as-string> --content-type <content-type> --header "<header-name>: <header-value>"
```

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,81 @@
...
+    ```
+    where `stream-name`, `payload` and `content-type` are mandatory and `headers` can be used multiple times.
```
scothis:
```suggestion
    where `stream-name`, `payload` and `content-type` are mandatory and `header` can be used multiple times.
```

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,81 @@
...
+1. **subscribe-stream:** To subscribe for events from the given stream.
+The command takes the form:
+    ```
+    riff streaming stream subscribe <stream-name> --offset <long-offset>
```
scothis: riff cli holdover?

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,81 @@
...
+    ```
+    This will display all the events in the stream from the given offset in the following json format:
+    ```
+    {
```
scothis: since there will be multiple json objects emitted, using newlines this way is a bit awkward to parse by tools. Since the output is being streamed, I'd recommend [line delimited json](https://en.wikipedia.org/wiki/JSON_streaming#Line-delimited_JSON).

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,81 @@
...
+        "payload": "base64 encoded user payload",
+        "content-type": "the content type of the message",
+        "headers": {"user provided header": "while publishing"}
+    }
+    ```
+    We will use the content-type of the event to determine how the payload should be displayed. i.e. for content-type "text/plain" and "application/json" convert the binary payload to string, for all others display base64 encoded string.
```
scothis: I urge treating all messages the same unless the user opts-in to some other behavior. There is no universal way to detect if the content is text, so any attempt will be imperfect.

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,81 @@
...
+kubectl run --image=projectriff/developer-utils:latest
+```
+1. Invoke the commands specified above using kubectl exec. An example follows:
+```
+kubectl exec -it riff-developer-utils -- invoke-core upper -d test
+```
+
+
+### Backwards Compatibility and Upgrade Path
+*How do the proposed changes impact backwards-compatibility? Are APIs and CLI commands changing?*
```
scothis: do these questions have answers?

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,81 @@
...
+    }
+    ```
+    We will use the content-type of the event to determine how the payload should be displayed. i.e. for content-type "text/plain" and "application/json" convert the binary payload to string, for all others display base64 encoded string.
+
+The namespace parameter is optional for all the commands. If not specified, namespace of the `riff-developer-utils` pod will be assumed.
+
+### User Impact
+The user will be able to able to invoke functions/deployers and publish and subscribe to streams using the following workflow:
+1. Run the `riff-developer-utils` pod in their cluster with something like:
+```
```
scothis: indent so that the order list renders correctly

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,81 @@
...
+kubectl run --image=projectriff/developer-utils:latest
+```
+1. Invoke the commands specified above using kubectl exec. An example follows:
+```
```
scothis: indent so that the order list renders correctly

rfc/rfc-0003-developer-utils.md
sbawaska: When running these commands using `kubectl exec` the separator `--` is already used. How about `-c` for curl options?

rfc/rfc-0003-developer-utils.md
sbawaska: incorporated!

rfc/rfc-0003-developer-utils.md
sbawaska: fixed!

rfc/rfc-0003-developer-utils.md
sbawaska: fixed.

rfc/rfc-0003-developer-utils.md
sbawaska: fixed.

rfc/rfc-0003-developer-utils.md
sbawaska: fixed.

rfc/rfc-0003-developer-utils.md
sbawaska: Incorporated!

rfc/rfc-0003-developer-utils.md
sbawaska: updated proposal to use `-c`

rfc/rfc-0003-developer-utils.md
sbawaska: updated proposal to use `-c`

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,74 @@
+# RFC-#: A pod with utilities for better developer experience
```
glyn:
```suggestion
# RFC-0003: A pod with utilities for better developer experience
```

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,74 @@
...
+
+**Authors:** Swapnil Bawaskar
+
+**Status:**
+
+**Pull Request URL:** [#1362](https://github.com/projectriff/riff/pull/1362)
+
+**Superseded by:** N/A
+
+**Supersedes:** [RFC 0001](https://github.com/projectriff/riff/pull/1359)
```
glyn: I guess this PR should also update the "Superseded by" section of RFC 0001?

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,74 @@
...
+    ```
+    subscribe-stream <stream-name> --offset <long-offset> --payload-as-string
+    ```
+    This will display all the events in the stream from the given offset in the following json format:
+    ```
+    {"payload": "base64 encoded user payload","content-type": "the content type of the message","headers": {"user provided header": "while publishing"}}
+    ```
+    The payload will be base64 encoded unless the `--payload-as-string` flag is present, in which case it will be displayed as a string.
+
+The namespace parameter is optional for all the commands. If not specified, namespace of the `riff-developer-utils` pod will be assumed.
```
glyn: How will the CLI determine the namespace of the `riff-developer-utils` pod?

rfc/rfc-0003-developer-utils.md
sbawaska: Superseded by for rfc-0001 was updated when I updated the status to "rejected" and merged it: https://github.com/projectriff/riff/pull/1359.

rfc/rfc-0003-developer-utils.md
sbawaska: It is available on the pod at: `/var/run/secrets/kubernetes.io/serviceaccount/namespace`

glyn: Great, thanks.

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,74 @@
...
+
+### User Impact
+The user will be able to able to invoke functions/deployers and publish and subscribe to streams using the following workflow:
+1. Run the `riff-developer-utils` pod in their cluster with something like:
+    ```
+    kubectl run --image=projectriff/developer-utils:latest
+    ```
+1. Invoke the commands specified above using kubectl exec. An example follows:
+    ```
+    kubectl exec -it riff-developer-utils -- invoke-core upper -d test
```
glyn: If the pod is run in another namespace, is it merely necessary to provide the same namespace on `kubectl exec`? If so, it may be worth adding a namespace parameter to both examples for clarity.

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,74 @@
...
+### Anti-Goals
+We will only address this problem for development/demos, not production, so topics like auth/authz are out of scope for this document.
+
+## Solution
+We will ask the developers to run a `riff-developer-utils` pod in their development cluster. This pod will bundle a few binaries/scripts that users can invoke using `kubectl exec`. Since the commands will be run in-cluster, users won't need to run `kubectl port-forward`.
+We will have the following commands to start with:
+1. **invoke-core:** To invoke the given core deployer.  
+The command takes the form:  
+    ```
+    invoke-core <deployer-name> -n <namespace> -c "<curl-params>"
```
scothis: it's very common to need to quote curl params, forcing nested quoting on users is 😿

rfc/rfc-0003-developer-utils.md
scothis: > how is the deployer name resolved to a host name? A vanilla pod won't have permission

this question wasn't addressed

rfc/rfc-0003-developer-utils.md
scothis: > Do we care about the knative adapter resource?

this question wasn't addressed

rfc/rfc-0003-developer-utils.md
sbawaska: In my prototype, I ended up creating a clusterrole (with read permissions for riff resources) and clusterrolebinding for the service account that our pod ran under. Do you know if there is anything else we can do?

rfc/rfc-0003-developer-utils.md
sbawaska: may be not until we see that it would be used.

rfc/rfc-0003-developer-utils.md
ericbottard: Wouldn't it be better to re-do what we did before, having a `--` separator and passing everything after that to curl?

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,74 @@
...
+    ```
+    The payload will be base64 encoded unless the `--payload-as-string` flag is present, in which case it will be displayed as a string.
+
+The namespace parameter is optional for all the commands. If not specified, namespace of the `riff-developer-utils` pod will be assumed.
+
+### User Impact
+The user will be able to able to invoke functions/deployers and publish and subscribe to streams using the following workflow:
+1. Run the `riff-developer-utils` pod in their cluster with something like:
+    ```
+    kubectl run --image=projectriff/developer-utils:latest
```
markfisher:
```suggestion
    kubectl run --image=projectriff/developer-utils
```

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,74 @@
...
+1. **publish-stream:** To publish an event to the given stream.
+The command takes the form:
+    ```
+    publish-stream <stream-name> -n <namespace> --payload <payload-as-string> --content-type <content-type> --header "<header-name>: <header-value>"
+    ```
+    where `stream-name`, `payload` and `content-type` are mandatory and `header` can be used multiple times.
+1. **subscribe-stream:** To subscribe for events from the given stream.
+The command takes the form:
+    ```
+    subscribe-stream <stream-name> --offset <long-offset> --payload-as-string
```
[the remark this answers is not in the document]
ericbottard: liiklus doesn't support subscribing from a given offset anyway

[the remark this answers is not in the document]
sbawaska: hmm, I saw `LastKnownOffset` was part of the receiveRequest, but had not tested it: https://github.com/projectriff/stream-client-go/blob/master/pkg/liiklus/LiiklusService.pb.go#L418

will remove long-offset for now.

rfc/rfc-0003-developer-utils.md
sbawaska: Will investigate if kubectl exec will allow an additional `--` since it will already have one:
```
kubectl exec -it riff-developer-utils -- invoke-core upper -d test
```

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,74 @@
...
+    ```
+    where `<curl-params>` are passed along to the post-request.
+1. **invoke-knative:** To invoke the given knative deployer.
+The command takes the form:  
+    ```
+    invoke-knative <deployer-name> -n <namespace> -c "<curl-params>"
+    ```
+    where `<curl-params>` are passed along to the post-request.
+
+1. **publish-stream:** To publish an event to the given stream.
```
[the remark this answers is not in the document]
sbawaska: done.

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,74 @@
...
+The command takes the form:
+    ```
+    publish-stream <stream-name> -n <namespace> --payload <payload-as-string> --content-type <content-type> --header "<header-name>: <header-value>"
+    ```
+    where `stream-name`, `payload` and `content-type` are mandatory and `header` can be used multiple times.
+1. **subscribe-stream:** To subscribe for events from the given stream.
```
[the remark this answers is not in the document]
sbawaska: done.

rfc/rfc-0003-developer-utils.md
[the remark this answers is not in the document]
sbawaska: added `--from-latest` flag, displays everything by default.

markfisher:
@sbawaska I think the default should maybe be "from latest". FWIW that's the default for the `kafka-console-consumer` which has a `--from-beginning` flag instead.

rfc/rfc-0003-developer-utils.md
sbawaska: this works! changed the rfc to use `--`

scothis: the text of the rfc still uses `-c`

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,74 @@
...
+The command takes the form:  
+    ```
+    invoke-knative <deployer-name> -n <namespace> -c "<curl-params>"
+    ```
+    where `<curl-params>` are passed along to the post-request.
+
+1. **publish-stream:** To publish an event to the given stream.
+The command takes the form:
+    ```
+    publish-stream <stream-name> -n <namespace> --payload <payload-as-string> --content-type <content-type> --header "<header-name>: <header-value>"
```
scothis: It needs to be more clear what the structure of the payload is. Via a CLI the value is assumed to be text, which needs to be encoded as binary data for the message. It should also be easy to use a binary payload. 

Consider using two different flags for the payload, one for plain text, and another for a base64 encoded value.

rfc/rfc-0003-developer-utils.md
scothis: how long does the subscribe command live? Is there a way to consume a window of message and exit? If the connection is interrupted, who is responsible for reestablishing the connection?

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,74 @@
...
+    ```
+    {"payload": "base64 encoded user payload","content-type": "the content type of the message","headers": {"user provided header": "while publishing"}}
+    ```
+    The payload will be base64 encoded unless the `--payload-as-string` flag is present, in which case it will be displayed as a string.
+
+The namespace parameter is optional for all the commands. If not specified, namespace of the `riff-developer-utils` pod will be assumed.
+
+### User Impact
+The user will be able to able to invoke functions/deployers and publish and subscribe to streams using the following workflow:
+1. Run the `riff-developer-utils` pod in their cluster with something like:
```
scothis: what permissions does the pod need? by default it has no privilege to talk to the api server.

rfc/rfc-0003-developer-utils.md
scothis: With the current proposed syntax, I'm not sure the invoke-core command can accept a namespace since the flag would be conflated with curl arguments. It would also be necessary for the pod to have RBAC permission to fetch the core deployer in another namespace.

jldec:
I know that message payload encoding was discussed in RFC-0001 [here](https://github.com/projectriff/riff/pull/1359#discussion_r348414423) and [here](https://github.com/projectriff/riff/pull/1359#discussion_r346605836), but i just want to confirm the desired behavior.

I.e. If a specific message payload was published as application/json, then users running `subscribe <stream-name> --payload-as-string` will receive stringified json as the payload for that message, rather than json embedded as the value of the payload key.

E.g. if the original payload was published with `content-type: application/json` and a value of `{"foo":1, "bar":["abc",1,2,3]}` the output will be:

`{"payload":"{\\"foo\\":1,\\"bar\\":[\\"abc\\",1,2,3]}"}`

instead of

`{"payload":{"foo":1,"bar":["abc",1,2,3]}}`

is my understanding right?

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,74 @@
...
+1. **subscribe-stream:** To subscribe for events from the given stream.
+The command takes the form:
+    ```
+    subscribe-stream <stream-name> --offset <long-offset> --payload-as-string
+    ```
+    This will display all the events in the stream from the given offset in the following json format:
+    ```
+    {"payload": "base64 encoded user payload","content-type": "the content type of the message","headers": {"user provided header": "while publishing"}}
+    ```
+    The payload will be base64 encoded unless the `--payload-as-string` flag is present, in which case it will be displayed as a string.
```
scothis: the flag name `--payload-as-string` isn't sitting well with me. The purpose of base64 encoding the payload is so that it can be emitted as a string. By not base64 encoding, the payload value may or may not be a string. Nobody wants binary data in their shell.

I suggest dropping the `--payload-as-string` flag and force consumers to decode the payload if they are interested in the value.

jldec:
For examples (from the cloudevents spec) of json with embedded binary, text, and json payloads see:

https://github.com/cloudevents/spec/blob/master/json-format.md#32-examples

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,74 @@
...
+A developer would like to invoke the function for which they just created the `core deployer` or the `knative deployer`.
+When the streaming runtime is installed, they would like to test the processor by getting events into the stream and look into the contents of the stream. There is no one step mechanism for accomplishing any of these currently.
+An [earlier proposal](https://github.com/projectriff/riff/pull/1359) was to provide some of this functionality in the riff cli, this proposal expands on [Mark's comment](https://github.com/projectriff/riff/pull/1359#discussion_r348617981).
+
+### Anti-Goals
+We will only address this problem for development/demos, not production, so topics like auth/authz are out of scope for this document.
+
+## Solution
+We will ask the developers to run a `riff-developer-utils` pod in their development cluster. This pod will bundle a few binaries/scripts that users can invoke using `kubectl exec`. Since the commands will be run in-cluster, users won't need to run `kubectl port-forward`.
+We will have the following commands to start with:
```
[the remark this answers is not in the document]
scothis: the way to avoid the collision would be to avoid doing the host name lookup in the dev-utils pod. Instead pass in a hostname. But at that point, you might as well just call `curl` directly.

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,74 @@
...
+The command takes the form:  
+    ```
+    invoke-knative <deployer-name> -n <namespace> -- <curl-params>
+    ```
+    where everything after `--` is passed as parameter to curl
+
+1. **publish:** To publish an event to the given stream.
+The command takes the form:
+    ```
+    publish <stream-name> -n <namespace> --payload <payload-as-string> --content-type <content-type> --header "<header-name>: <header-value>"
```
jldec: How about supporting publishing from stdin, where that matches the output of a subscribe command. (pipe)

scothis:
Trying to bring this in for a landing. These are a few changes we can make to simplify and clarify the RFC:
- remove the core/knative invoke commands, `curl` will be available in the pod
- add a `--payload-base64` flag for encoded binary data that is mutually exclusive with `--payload` (which only handles text)
- clarify that subscribe runs until canceled (SIGINT)
- subscribe output is LD JSON with base64 encoded payloads
- `jq` and `base64` will be available in the pod for decoding messages (or do it on your own machine)
- clarify that each pod is only expected to operate on resources within the same namespace (create a new pod to interact with resource in another namespace)
- clarify that a serviceaccount and rolebinding should be created to give the pod access to the api server (the implementation should provide advice as to the RBAC permissions needed)
- other commands and improvements can be defined in future RFCs

rfc/rfc-0003-developer-utils.md
scothis: dropped the invoke command in favor of curl

rfc/rfc-0003-developer-utils.md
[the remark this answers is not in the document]
scothis: dropped the invoke commands in favor of curl

rfc/rfc-0003-developer-utils.md
scothis: added a `--payload-base64` flag that only accepts base64 encoded values, and is mutually exclusive with `--payload`

rfc/rfc-0003-developer-utils.md
scothis: Clarified that it should run until terminated (SIGTERM) or encounters an error.

rfc/rfc-0003-developer-utils.md
scothis: Punted to the implementation.

rfc/rfc-0003-developer-utils.md
scothis: Dropped, the payload will always be base64 encoded since that's the only safe output.

scothis:
@sbawaska @markfisher @jldec @glyn @ericbottard I'd like to get this approved and merged tomorrow. We descoped a good bit that was leading to ambiguities or could be tricky to implement. Future commands and behavior can be defined by future RFCs.

Please open suggestions for typos and grammatical fixes, but hold off on scope changes.

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,86 @@
...
+
+**Superseded by:** N/A
+
+**Supersedes:** [RFC 0001](https://github.com/projectriff/riff/pull/1359)
+
+**Related:**
+
+
+## Problem
+A developer would like to interact with riff resources inside the cluster in ways that may not be desirable to expose outside of the cluster. This may be reading or writing message from a stream or invoking an http workload whose ingress policy is cluster-local. There is no one step mechanism for accomplishing any of these currently.
```
markfisher:
```suggestion
A developer would like to interact with riff resources inside the cluster in ways that may not be desirable to expose outside of the cluster. This may be reading or writing messages from/to a stream or invoking an http workload whose ingress policy is cluster-local. There is no one step mechanism for accomplishing any of these currently.
```

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,86 @@
...
+
+An [earlier proposal](https://github.com/projectriff/riff/pull/1359) was to provide some of this functionality in the riff cli. The current proposal is a response to [feedback](https://github.com/projectriff/riff/pull/1359#discussion_r348617981) on that earlier proposal.
+
+### Anti-Goals
+We will only address this problem for development/demos, not production, so topics like auth/authz are out of scope for this document.
+
+## Solution
+We will ask the developers to run a `riff-dev` pod in their development cluster. This pod will bundle a few commands that users can invoke using `kubectl exec`. Since the commands will be run in-cluster, users won't need to run `kubectl port-forward`, however, the pod will need to be run with a service account that has appropriate RBAC permissions.
```
markfisher:
```suggestion
We will ask the developers to run a `riff-dev` pod in their development cluster. This pod will bundle a few commands that users can invoke using `kubectl exec`. Since the commands will be run in-cluster, users won't need to run `kubectl port-forward`. However, the pod will need to be run with a service account that has appropriate RBAC permissions.
```

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,86 @@
...
+    {"payload":"base64 encoded message payload","content-type":"the content type of the message","headers":{"header name": "header value"}}
+    ```
+    
+    Since the payload may contain binary data, it will be displayed as a base64 encoded string. Tools like `jq` and `base64 --decode` are useful to get a raw payload.
+    
+    The subscribe command will run until terminated (SIGTERM), or an error occurs on the stream.
+
+1. [**jq**](https://stedolan.github.io/jq/): Process JSON.
+
+1. [**base64:**](http://manpages.ubuntu.com/manpages/bionic/man1/base64.1.html) Encode and decode base64 strings.
```
markfisher:
```suggestion
1. [**base64**](http://manpages.ubuntu.com/manpages/bionic/man1/base64.1.html): Encode and decode base64 strings.
```

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,86 @@
...
+
+1. [**curl:**](https://curl.haxx.se) Make HTTP requests.
```
markfisher:
```suggestion
1. [**curl**](https://curl.haxx.se): Make HTTP requests.
```

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,86 @@
+# RFC-0003: A pod with utilities for better developer experience
+
+**Authors:** Swapnil Bawaskar
```
markfisher:
```suggestion
**Authors:** Swapnil Bawaskar, Scott Andrews
```

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,86 @@
...
+**Related:**
+
+
+## Problem
+A developer would like to interact with riff resources inside the cluster in ways that may not be desirable to expose outside of the cluster. This may be reading or writing message from a stream or invoking an http workload whose ingress policy is cluster-local. There is no one step mechanism for accomplishing any of these currently.
+
+An [earlier proposal](https://github.com/projectriff/riff/pull/1359) was to provide some of this functionality in the riff cli. The current proposal is a response to [feedback](https://github.com/projectriff/riff/pull/1359#discussion_r348617981) on that earlier proposal.
+
+### Anti-Goals
+We will only address this problem for development/demos, not production, so topics like auth/authz are out of scope for this document.
```
markfisher: this might require clarification in relation to the new mentions of RBAC policies and  Service Accounts

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,86 @@
...
+We will ask the developers to run a `riff-dev` pod in their development cluster. This pod will bundle a few commands that users can invoke using `kubectl exec`. Since the commands will be run in-cluster, users won't need to run `kubectl port-forward`, however, the pod will need to be run with a service account that has appropriate RBAC permissions.
+
+We will have the following commands to start with:
+
+1. **publish:** To publish an event to the given stream.
+    
+    The command takes the form:
+    
+    ```
+    publish <stream-name> --content-type <content-type> [--payload <payload-as-plain-text>] [--payload-base64 <payload-as-base64-text>] [--header "<header-name>: <header-value>"]
```
markfisher: should probably use curly-braces and a pipe to indicate that exactly one of the payload options is required (since square brackets typically mean optional)

```suggestion
    publish <stream-name> --content-type <content-type> {--payload <payload-as-plain-text> | --payload-base64 <payload-as-base64-text>} [--header "<header-name>: <header-value>"]
```
see: https://developers.google.com/style/code-syntax#set-of-two-arguments

rfc/rfc-0003-developer-utils.md
```diff
@@ -0,0 +1,86 @@
...
+    where `stream-name`, and `--content-type` are mandatory and `--header` can be used multiple times. `--payload` and `--payload-base64` are mutually exclusive. Binary data should only be passed via `--payload-base64` after it has been `base64` encoded.
+
+1. **subscribe:** To subscribe for events from the given stream.
+    The command takes the form:
+    
+    ```
+    subscribe <stream-name> [--from-beginning]
+    ```
+    
+    If the `--from-beginning` option is present, display all the events in the stream, otherwise only new events are displayed as [JSON Lines](http://jsonlines.org) in the following form:
```
markfisher:
```suggestion
    If the `--from-beginning` option is present, display all the events in the stream. Otherwise only new events are displayed as [JSON Lines](http://jsonlines.org) in the following form:
```

rfc/rfc-0003-developer-utils.md
scothis: This RFC acknowledges that RBAC permissions will be required for the pod to access resources on the cluster, but does not prescribe what that permissions are. It says:

> The implementation should recommend specific RBAC policies needed to use the commands

PR #1367, opened 2019-11-29
RFC-0005: Optional starting offset for processor inputs.

ericbottard:
Implementation at:
system: https://github.com/projectriff/system/pull/231
cli: https://github.com/projectriff/cli/pull/164
streaming-processor: https://github.com/projectriff/streaming-processor/pull/38

### DIFF
--- a/rfc/rfc-0005-customizable-consumption-offset.md
+++ b/rfc/rfc-0005-customizable-consumption-offset.md
@@ -2,7 +2,7 @@
 
 **Authors:** Eric Bottard
 
-**Status:**
+**Status:** Accepted
 
 **Pull Request URL:** https://github.com/projectriff/riff/pull/1367
 
@@ -15,8 +15,6 @@
 
 ## Problem
 Sometimes users want to replay data through functions, maybe because the function logic has changed, or maybe because downstream systems have lost state and that state needs to be re-created. The current Processor concept in riff doesn't support such a mechanism though, and always consumes data from the (current) end of streams.
-
-### Anti-Goals
 
 ## Solution
 With a top to bottom view of the user experience, the proposed solution is to
@@ -62,6 +60,3 @@ NOTE: riff currently uses in-memory positions storage for liiklus, until support
 The proposed change should be totally backwards compatible with existing riff with
 - the cli option being optional
 - its default value being `latest`, which is the current effective mode used.
-
-## FAQ
-*Answers to common questions that may arise and those that you’ve commonly been asked after requesting comments for this proposal.*
--- a/rfc/rfc-0005-customizable-consumption-offset.md
+++ b/rfc/rfc-0005-customizable-consumption-offset.md
@@ -1,10 +1,10 @@
-# RFC-#: Title
+# RFC-0005: Optional starting offset for processor inputs
 
 **Authors:** Eric Bottard
 
 **Status:**
 
-**Pull Request URL:** link
+**Pull Request URL:** https://github.com/projectriff/riff/pull/1367
 
 **Superseded by:** N/A
 
@@ -25,6 +25,23 @@ With a top to bottom view of the user experience, the proposed solution is to
 - have the Processor reconciler forward that option to the streaming-processor, so that it changes its SubscriptionRequests accordingly
 
 ### User Impact
+This RFC proposes to change the `Inputs` field of the `Processor` CRD from
+```go
+Inputs []StreamBinding `json:"inputs"`
+```
+to
+```go
+Inputs []InputStreamBinding `json:"inputs"`
+...
+type InputStreamBinding struct {
+    StreamBinding
+	// Where to start consuming this stream the first time a processor runs.
+    StartOffset int64 `json:"startOffset"`
+}
+```
+
+with `StartOffset` supporting only two special, negative values for now: earliest and latest (default).
+
 This RFC proposes to change the `riff streaming processor create` command from
 ```
 riff streaming processor create <name> ... --input [alias:]stream ...
--- a/rfc/rfc-0005-customizable-consumption-offset.md
+++ b/rfc/rfc-0005-customizable-consumption-offset.md
@@ -42,7 +42,7 @@ type InputStreamBinding struct {
 }
 ```
 
-with `StartOffset` supporting only two special, textual values for now: `earliest` and `latest` (default). Future support for parsing to a non negative number after failing to recognize those two values is possible.
+with `StartOffset` supporting only two special, textual values for now: `Earliest` and `Latest` (default). Future support for parsing to a non negative number after failing to recognize those two values is possible.
 
 This RFC proposes to change the `riff streaming processor create` command from
 ```
@@ -52,17 +52,17 @@ to
 ```
 riff streaming processor create <name> ... --input [alias:]stream[:offset] ...
 ```
-where `[..]` denotes optional values and `[:offset]` could take values `earliest` or `latest`. Those two logical values are the only ones currently supported by liiklus (and the high level kafka consumer as far as the author can remember). Nevertheless, the encoding of those two values could be done via negative numerical values, in preparation of future support of an actual numerical index.
+where `[..]` denotes optional values and `[:offset]` could take values `Earliest` or `Latest`. Those two logical values are the only ones currently supported by liiklus (and the high level kafka consumer as far as the author can remember). Nevertheless, the encoding of those two values could be done via negative numerical values, in preparation of future support of an actual numerical index. Please also note that the use of a Capitalized word allows disambiguation between the alias:stream and stream:offset cases, because capitals are forbidden in stream names.
 
 
-The actual behavior of this proposed change has to be well understood: the offset parameter only makes sense for a _new consumer group_, _i.e._ a group which has never been used and for which liiklus doesn't currently have positions. As soon as the group starts to be used and its positions stored, the `earliest` value will have no effect. Indeed, this is desired behavior otherwise everytime a processor scales down and up again, it would replay everything over and over again. The corollary from this is that to start consuming from the beginning *again*, a processor would need to be *re-created*, with a different name (because the consumer group is based on the name of the processor).
+The actual behavior of this proposed change has to be well understood: the offset parameter only makes sense for a _new consumer group_, _i.e._ a group which has never been used and for which liiklus doesn't currently have positions. As soon as the group starts to be used and its positions stored, the `Earliest` value will have no effect. Indeed, this is desired behavior otherwise everytime a processor scales down and up again, it would replay everything over and over again. The corollary from this is that to start consuming from the beginning *again*, a processor would need to be *re-created*, with a different name (because the consumer group is based on the name of the processor).
 
 NOTE: riff currently uses in-memory positions storage for liiklus, until support for Kafka storage is supported. So technically, restarting the liiklus process of a given provider would allow starting from a blank slate.
 
 ### Backwards Compatibility and Upgrade Path
 The proposed change should be totally backwards compatible with existing riff with
 - the cli option being optional
-- its default value being `latest`, which is the current effective mode used.
+- its default value being `Latest`, which is the current effective mode used.
 
 ## FAQ
 *Answers to common questions that may arise and those that you’ve commonly been asked after requesting comments for this proposal.*
--- a/rfc/rfc-0005-customizable-consumption-offset.md
+++ b/rfc/rfc-0005-customizable-consumption-offset.md
@@ -10,7 +10,7 @@
 
 **Supersedes:** N/A
 
-**Related:** rfc-0003: developer utils
+**Related:** rfc-0003: developer utils, which also support viewing stream data from the earliest offset.
 
 
 ## Problem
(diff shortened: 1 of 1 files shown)


scothis:
How long is this open for comments?

rfc/rfc-0005-customizable-consumption-offset.md
```diff
@@ -0,0 +1,49 @@
+# RFC-#: Title
```
scothis: add a title

rfc/rfc-0005-customizable-consumption-offset.md
```diff
@@ -0,0 +1,49 @@
...
+### Anti-Goals
+
+## Solution
+With a top to bottom view of the user experience, the proposed solution is to
+- introduce an offset option on each input binding when a processor is created
+- persist that option in the Processor CRD instance on the cluster
+- have the Processor reconciler forward that option to the streaming-processor, so that it changes its SubscriptionRequests accordingly
+
+### User Impact
+This RFC proposes to change the `riff streaming processor create` command from
```
scothis: We should consider the impact on the CRDs first, since that is the public API for riff.

ericbottard:
Hopefully end of day dec 3rd EST should be enough?

rfc/rfc-0005-customizable-consumption-offset.md
```diff
@@ -0,0 +1,49 @@
...
+```
+to
+```
+riff streaming processor create <name> ... --input [alias:]stream[:offset] ...
+```
+where `[..]` denotes optional values and `[:offset]` could take values `start` or `end`. Those two logical values are the only ones currently supported by liiklus (and the high level kafka consumer as far as the author can remember). Nevertheless, the encoding of those two values could be done via negative numerical values, in preparation of future support of an actual numerical index.
+
+
+The actual behavior of this proposed change has to be well understood: the offset parameter only makes sense for a _new consumer group_, _i.e._ a group which has never been used and for which liiklus doesn't currently have positions. As soon as the group starts to be used and its positions stored, the `start` value will have no effect. Indeed, this is desired behavior otherwise everytime a processor scales down and up again, it would replay everything over and over again. The corollary from this is that to start consuming from the beginning *again*, a processor would need to be *re-created*, with a different name (because the consumer group is based on the name of the processor).
```
scothis: how are the consumer groups managed when there are multiple consumers? What connects a specific workload to a specific consumer group at runtime as processors scale up and down?

rfc/rfc-0005-customizable-consumption-offset.md
ericbottard: Added section about CRD changes

rfc/rfc-0005-customizable-consumption-offset.md
```diff
@@ -0,0 +1,66 @@
...
+
+This RFC proposes to change the `riff streaming processor create` command from
+```
+riff streaming processor create <name> ... --input [alias:]stream ...
+```
+to
+```
+riff streaming processor create <name> ... --input [alias:]stream[:offset] ...
+```
+where `[..]` denotes optional values and `[:offset]` could take values `start` or `end`. Those two logical values are the only ones currently supported by liiklus (and the high level kafka consumer as far as the author can remember). Nevertheless, the encoding of those two values could be done via negative numerical values, in preparation of future support of an actual numerical index.
```
sbawaska: reading after `end` seems a little confusing. I like `earliest` and `latest`.

rfc/rfc-0005-customizable-consumption-offset.md
markfisher: currently there's a 1:1 relationship between a processor resource and a consumer group, but we should consider an intermediate resource to enable a 1:n relationship so that a processor could be replayed (or eventually played from an explicit offset)... for example in a case where the function is updated to fix a bug, and the entire event log should be re-processed

ericbottard: I'm not sure I understand the rationale. A processor is just a way to stitch together a function and some streams. If you want another run (with a different group) of that combination, just create another processor with the same combination.

Also, while I find those questions interesting (and took some time to dig into actual liiklus code), I don't think this should happen here. This (light) rfc was opened to get quick consensus on the way the start offset feature should "feel".

markfisher: One of the values of an RFC is that it can stimulate ideas that we might not have considered otherwise, and I think the notion that a Processor may have identity that spans replay is worth considering (similar to how a Deployment has identity that spans sequences of ReplicaSets). If we decide there's merit to that approach, it would mean potentially not adding the one-off configuration option for this at the Processor level. Typically a Processor will have a long lifespan (continuing to read from the "latest" end of the stream until deleted), so I think we should question whether the offset as a configuration option that is set only at initialization time is the right approach.

scothis: > (because the consumer group is based on the name of the processor)

For the same reasons that stream names should not translate directly to topics, processor names should not translate to consumer groups.

rfc/rfc-0005-customizable-consumption-offset.md
```diff
@@ -0,0 +1,66 @@
...
+Inputs []InputStreamBinding `json:"inputs"`
+...
+type InputStreamBinding struct {
+    StreamBinding
+	// Where to start consuming this stream the first time a processor runs.
+    StartOffset int64 `json:"startOffset"`
+}
+```
+
+with `StartOffset` supporting only two special, negative values for now: earliest and latest (default).
```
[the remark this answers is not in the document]
ericbottard: Those two semantic values are commonly used in e.g. the high level kafka consumer and are different from actual offset seeking available in other solutions. Moreover, I'm afraid 0 is not always strictly equivalent to earliest, in the case of log compaction for example.

To the end user, at least for the time being, I think it's better to expose semantic values and not "lie" on what we're doing.

rfc/rfc-0005-customizable-consumption-offset.md
```diff
@@ -0,0 +1,66 @@
...
+- have the Processor reconciler forward that option to the streaming-processor, so that it changes its SubscriptionRequests accordingly
+
+### User Impact
+This RFC proposes to change the `Inputs` field of the `Processor` CRD from
+```go
+Inputs []StreamBinding `json:"inputs"`
+```
+to
+```go
+Inputs []InputStreamBinding `json:"inputs"`
```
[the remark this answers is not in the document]
ericbottard: I'd be in favor of distinguishing the two. Struct embedding turns out to be impractical in this case though, so consider the codes snippets here as pseudo code explaining the philosophy of the change.

rfc/rfc-0005-customizable-consumption-offset.md
ericbottard: > For the same reasons that stream names should not translate directly to topics, processor names should not translate to consumer groups.

I take those reasons are inclusion of the namespace and collision avoidance. Still, in the same way that there is a 1:1 mapping between streams and topics, there is currently a 1:1 mapping between processor and consumer group (also, created [this](https://github.com/projectriff/system/issues/226)). 

I'm fine considering departure from that model, in this RFC or another one if we want to move forward in a timely fashion. I'm not sure I understand what you mean @markfisher by "identity that spans replay" though. Currently, the processor identity captures a function + a set of streams. Tomorrow it may also encompass the willingness to "start from the beginning for stream X, Y, Z". Are you saying _that whole set of characteristics_ is "the identity" and one may want to "do it again", ie "(re-)create a «processor» that uses the same set of start-offset semantics"? If so, how is that not creating a processor with a new name? Alternatively, should we add something like `riff streaming processor <foo> replay`?

markfisher: If we depart from the model, it can definitely be in a different RFC and a longer term initiative. I just want to make sure we don't introduce something that makes that more difficult later. Perhaps a way to consider my suggestion that seems less extreme - and more aligned with the way you just described it @ericbottard - would be to consider the new layer of resource being added _above_ the current Processor resource instead of _beneath_.

rfc/rfc-0005-customizable-consumption-offset.md
markfisher: I agree that "end" is confusing since the stream may "never end". Similarly, "start" is misleading in the case where the stream has been truncated based on the broker's retention policy.

It does seem that "earliest" and "latest" most accurately describe where we want to start consuming based on the currently stored events, and those are the 2 values accepted for the Kafka consumer's `auto.offset.reset` property as listed in the table here: https://kafka.apache.org/documentation/#consumerconfigs

rfc/rfc-0005-customizable-consumption-offset.md
```diff
@@ -0,0 +1,68 @@
...
+type InputStreamBinding struct {
+    Stream string `json:"stream"`
+    Alias string `json:"alias,omitempty"`
+
+	// Where to start consuming this stream the first time a processor runs.
+    StartOffset *int64 `json:"startOffset"`
+}
+```
+
+with `StartOffset` supporting only two special, negative values for now: `Earliest` and `Latest` (default).
```
scothis: We should not use sentinel values for APIs that are user facing. Even more so when the common tools for interacting with the API will do not have facilities to encode/resolve the value.

Consider a user viewing this raw yaml (not to mention trying to author)...


```yaml
spec:
  inputs:
  - stream: foo
    startOffset: -1
```

Compared to

```yaml
spec:
  inputs:
  - stream: foo
    startOffset: -2
```

Which resource does what? I don't know, and I wouldn't expect a user to know either.

rfc/rfc-0005-customizable-consumption-offset.md
```diff
@@ -0,0 +1,68 @@
...
+
+**Status:**
+
+**Pull Request URL:** https://github.com/projectriff/riff/pull/1367
+
+**Superseded by:** N/A
+
+**Supersedes:** N/A
+
+**Related:** rfc-0003: developer utils
```
scothis: how does this RFC relate to the developer utils?

rfc/rfc-0005-customizable-consumption-offset.md
```diff
@@ -0,0 +1,68 @@
...
+to
+```go
+Inputs []InputStreamBinding `json:"inputs"`
+...
+type InputStreamBinding struct {
+    Stream string `json:"stream"`
+    Alias string `json:"alias,omitempty"`
+
+	// Where to start consuming this stream the first time a processor runs.
+    StartOffset string `json:"startOffset"`
```
scothis: Something like [IntOrString](https://godoc.org/k8s.io/apimachinery/pkg/util/intstr#IntOrString) may be appropriate here, unless we really need the capacity of int64 vs int32.

rfc/rfc-0005-customizable-consumption-offset.md
```diff
@@ -0,0 +1,68 @@
...
+
+with `StartOffset` supporting only two special, textual values for now: `earliest` and `latest` (default). Future support for parsing to a non negative number after failing to recognize those two values is possible.
+
+This RFC proposes to change the `riff streaming processor create` command from
+```
+riff streaming processor create <name> ... --input [alias:]stream ...
+```
+to
+```
+riff streaming processor create <name> ... --input [alias:]stream[:offset] ...
```
scothis: Can we reliably distinguish a value like `foo:latest`? Is `foo` the alias or stream, is `latest` the stream or offset?

While not perfect, making the textual values `Earliest` and  `Latest` would make distinction possible since the stream name can't contain uppercase characters.

ericbottard: Very good point

markfisher: I don't think we should caps as a workaround for this. Maybe we're trying to include too much information in the same string?

markfisher: or maybe we use `[alias:]stream[@offset]` instead?

ericbottard: `@` is very good too, but reads better with strings like start/end than with earliest/latest.

markfisher: thinking "start reading the stream AT the earliest offset" works for me :-)

sbawaska: +1 for `@` as the delimiter rather than capitalizing `earliest` and `latest`.

rfc/rfc-0005-customizable-consumption-offset.md
```diff
@@ -0,0 +1,67 @@
...
+
+**Supersedes:** N/A
+
+**Related:** rfc-0003: developer utils, which also support viewing stream data from the earliest offset.
+
+
+## Problem
+Sometimes users want to replay data through functions, maybe because the function logic has changed, or maybe because downstream systems have lost state and that state needs to be re-created. The current Processor concept in riff doesn't support such a mechanism though, and always consumes data from the (current) end of streams.
+
+### Anti-Goals
```
markfisher: If this section is empty, we should probably leave out the heading. Otherwise, it might suggest that the following "Solution" is an anti-goal.

fbiville:
Merged as cf6efd1087a89d7337b5a9f4d608241db0b43f64

PR #1369, opened 2019-12-02
RFC 0006: Guidance for child resource names

scothis:
Comments due by Friday, December 6th

### DIFF
--- a/rfc/rfc-0006-child-resource-names.md
+++ b/rfc/rfc-0006-child-resource-names.md
@@ -14,20 +14,20 @@
 
 
 ## Problem
-When a controller creates a new resource, it need to either pick a static name, or let kubernetes generate a name. Static names are predictable, but may already be in use. Generated names cannot be predicated, but are guaranteed to be available.
+When a controller creates a new resource, it needs to either pick a static name or let Kubernetes generate a name. Static names are predictable, but may already be in use. Generated names cannot be predicated, but are guaranteed to be available.
 
 ### Anti-Goals
 This RFC does not prescribe which naming scheme to use for any specific resource.
 
 ## Solution
-While there will not be a cut and dry answer for each situation, there are heuristics to guide the decision:
+While there will not be a cut and dried answer for each situation, there are heuristics to guide the decision:
 - if a resource is not consumed directly by a non-controlled resource by name, generate the name (e.g. Pods in a ReplicaSet)
 - if the resource can be queried via a label selector instead of by name, generate the name (e.g. Pods fronted by a Service)
-- if a resource's name surfaces in outward facing way that is part of the public API, consider using a static name (e.g. Services in DNS)
+- if a resource's name surfaces in an outward facing way that is part of the public API, consider using a static name (e.g. Services in DNS)
 - if the name needs to be stable between different instances of the same resource, use a static name (e.g. immutable resources that cannot be updated)
-- generate the name
+- if none of the above criteria apply, generate the name
 
-Static names may use templates, but must be deterministic. based only on stable values of the parent resource (e.g. `sprintf('%s-suffix', current.name)`).
+Static names may use templates, but must be deterministic, based only on stable values of the parent resource (e.g. `sprintf('%s-suffix', current.name)`).
 
 If a static name collides with an existing resource, the parent resource should reflect the collision in its conditions. 
 
@@ -37,7 +37,7 @@ The name generally should not be chosen by the creator of the parent resource, a
 No public interface is changing, as the names of child resources are reflected on the parent resource's status. That will continue to be true.
 
 ### Backwards Compatibility and Upgrade Path
-riff generally uses generated names today, so current uses expect to use the name assigned by the system.
+riff generally uses generated names today, so current users expect to use the name assigned by the system.
 
 ## FAQ
 *Answers to common questions that may arise and those that you’ve commonly been asked after requesting comments for this proposal.*
--- a/rfc/rfc-0006-child-resource-names.md
+++ b/rfc/rfc-0006-child-resource-names.md
@@ -1,0 +1,43 @@
+# RFC-#: Child Resource Names
+
+**Authors:** Scott Andrews
+
+**Status:**
+
+**Pull Request URL:** link
+
+**Superseded by:** N/A
+
+**Supersedes:** N/A
+
+**Related:** N/A
+
+
+## Problem
+When a controller creates a new resource, it need to either pick a static name, or let kubernetes generate a name. Static names are predictable, but may already be in use. Generated names cannot be predicated, but are guaranteed to be available.
+
+### Anti-Goals
+This RFC does not prescribe which naming scheme to use for any specific resource.
+
+## Solution
+While there will not be a cut and dry answer for each situation, there are heuristics to guide the decision:
+- if a resource is not consumed directly by a non-controlled resource by name, generate the name (e.g. Pods in a ReplicaSet)
+- if the resource can be queried via a label selector instead of by name, generate the name (e.g. Pods fronted by a Service)
+- if a resource's name surfaces in outward facing way that is part of the public API, consider using a static name (e.g. Services in DNS)
+- if the name needs to be stable between different instances of the same resource, use a static name (e.g. immutable resources that cannot be updated)
+- generate the name
+
+Static names may use templates, but must be deterministic. based only on stable values of the parent resource (e.g. `sprintf('%s-suffix', current.name)`).
+
+If a static name collides with an existing resource, the parent resource should reflect the collision in its conditions. 
+
+The name generally should not be chosen by the creator of the parent resource, as it introduces complexity when reconciling updates for names changes as Kubernetes cannot move, or rename a resource.
+
+### User Impact
+No public interface is changing, as the names of child resources are reflected on the parent resource's status. That will continue to be true.
+
+### Backwards Compatibility and Upgrade Path
+riff generally uses generated names today, so current uses expect to use the name assigned by the system.
+
+## FAQ
+*Answers to common questions that may arise and those that you’ve commonly been asked after requesting comments for this proposal.*
--- a/rfc/rfc-0006-child-resource-names.md
+++ b/rfc/rfc-0006-child-resource-names.md
@@ -14,30 +14,27 @@
 
 
 ## Problem
-When a controller creates a new resource, it needs to either pick a static name or let Kubernetes generate a name. Static names are predictable, but may already be in use. Generated names cannot be predicated, but are guaranteed to be available.
+When a controller creates a new resource, it needs to either pick a static name or let Kubernetes generate a name. Static names are predictable, but may already be in use. Generated names cannot be predicated, but are guaranteed to be available. As the authors of reconcilers, we must define how each controlled resource is named.
 
 ### Anti-Goals
 This RFC does not prescribe which naming scheme to use for any specific resource.
 
 ## Solution
 While there will not be a cut and dried answer for each situation, there are heuristics to guide the decision:
-- if a resource is not consumed directly by a non-controlled resource by name, generate the name (e.g. Pods in a ReplicaSet)
-- if the resource can be queried via a label selector instead of by name, generate the name (e.g. Pods fronted by a Service)
-- if a resource's name surfaces in an outward facing way that is part of the public API, consider using a static name (e.g. Services in DNS)
-- if the name needs to be stable between different instances of the same resource, use a static name (e.g. immutable resources that cannot be updated)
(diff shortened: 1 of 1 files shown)


rfc/rfc-0006-child-resource-names.md
```diff
@@ -0,0 +1,43 @@
+# RFC-#: Child Resource Names
```
scothis:
```suggestion
# RFC-0006: Child Resource Names
```

rfc/rfc-0006-child-resource-names.md
```diff
@@ -0,0 +1,43 @@
...
+
+**Authors:** Scott Andrews
+
+**Status:**
+
+**Pull Request URL:** link
```
scothis:
```suggestion
**Pull Request URL:** [#1369](https://github.com/projectriff/riff/pull/1369)
```

rfc/rfc-0006-child-resource-names.md
```diff
@@ -0,0 +1,43 @@
...
+
+**Superseded by:** N/A
+
+**Supersedes:** N/A
+
+**Related:** N/A
+
+
+## Problem
+When a controller creates a new resource, it need to either pick a static name, or let kubernetes generate a name. Static names are predictable, but may already be in use. Generated names cannot be predicated, but are guaranteed to be available.
```
markfisher:
```suggestion
When a controller creates a new resource, it needs to either pick a static name or let Kubernetes generate a name. Static names are predictable, but may already be in use. Generated names cannot be predicated, but are guaranteed to be available.
```

rfc/rfc-0006-child-resource-names.md
```diff
@@ -0,0 +1,43 @@
...
+
+### Anti-Goals
+This RFC does not prescribe which naming scheme to use for any specific resource.
+
+## Solution
+While there will not be a cut and dry answer for each situation, there are heuristics to guide the decision:
+- if a resource is not consumed directly by a non-controlled resource by name, generate the name (e.g. Pods in a ReplicaSet)
```
markfisher:
```suggestion
- if a resource is not consumed directly by a non-controlling resource by name, generate the name (e.g. Pods in a ReplicaSet)
```

rfc/rfc-0006-child-resource-names.md
```diff
@@ -0,0 +1,43 @@
...
+- if the resource can be queried via a label selector instead of by name, generate the name (e.g. Pods fronted by a Service)
+- if a resource's name surfaces in outward facing way that is part of the public API, consider using a static name (e.g. Services in DNS)
```
markfisher:
```suggestion
- if a resource's name surfaces in an outward facing way that is part of the public API, consider using a static name (e.g. Services in DNS)
```

rfc/rfc-0006-child-resource-names.md
```diff
@@ -0,0 +1,43 @@
...
+- if the name needs to be stable between different instances of the same resource, use a static name (e.g. immutable resources that cannot be updated)
+- generate the name
```
markfisher:
```suggestion
- if none of the above criteria apply, generate the name
```

rfc/rfc-0006-child-resource-names.md
```diff
@@ -0,0 +1,43 @@
...
+
+Static names may use templates, but must be deterministic. based only on stable values of the parent resource (e.g. `sprintf('%s-suffix', current.name)`).
```
markfisher:
```suggestion
Static names may use templates, but must be deterministic, based only on stable values of the parent resource (e.g. `sprintf('%s-suffix', current.name)`).
```

rfc/rfc-0006-child-resource-names.md
```diff
@@ -0,0 +1,43 @@
...
+
+If a static name collides with an existing resource, the parent resource should reflect the collision in its conditions. 
+
+The name generally should not be chosen by the creator of the parent resource, as it introduces complexity when reconciling updates for names changes as Kubernetes cannot move, or rename a resource.
```
markfisher: but it's ok for the same name chosen by the creator of the parent resource to be used for the child resource, right? if so, should clarify that ("...if different from the parent resource name itself")

rfc/rfc-0006-child-resource-names.md
```diff
@@ -0,0 +1,43 @@
...
+
+### User Impact
+No public interface is changing, as the names of child resources are reflected on the parent resource's status. That will continue to be true.
+
+### Backwards Compatibility and Upgrade Path
+riff generally uses generated names today, so current uses expect to use the name assigned by the system.
```
markfisher:
```suggestion
riff generally uses generated names today, so current users expect to use the name assigned by the system.
```

rfc/rfc-0006-child-resource-names.md
glyn:
```suggestion
When a controller creates a new resource, it need to either pick a static name, or let kubernetes generate a name. Static names are predictable, but may already be in use. Generated names cannot be predicted, but are guaranteed to be available.
```

rfc/rfc-0006-child-resource-names.md
```diff
@@ -0,0 +1,43 @@
...
+
+
+## Problem
+When a controller creates a new resource, it need to either pick a static name, or let kubernetes generate a name. Static names are predictable, but may already be in use. Generated names cannot be predicated, but are guaranteed to be available.
+
+### Anti-Goals
+This RFC does not prescribe which naming scheme to use for any specific resource.
+
+## Solution
+While there will not be a cut and dry answer for each situation, there are heuristics to guide the decision:
```
glyn:
```suggestion
While there will not be a cut and dried answer for each situation, there are heuristics to guide the decision:
```

rfc/rfc-0006-child-resource-names.md
glyn: I'm not sure what this is getting at. It hints at various possibilities:
* a resource which is consumed directly by name or not (consumed indirectly or not by name or both or not consumed at all)
* the type of consumer ("controlled resource" which might mean created by a controller, "controlling resource" which might mean the consuming resource causes a controller to generate the consumed resource, something other than a resource such as a user)

and it only stipulates generating the name in one specific combination of possibilities.

rfc/rfc-0006-child-resource-names.md
```diff
@@ -0,0 +1,43 @@
...
+- if a resource is not consumed directly by a non-controlled resource by name, generate the name (e.g. Pods in a ReplicaSet)
+- if the resource can be queried via a label selector instead of by name, generate the name (e.g. Pods fronted by a Service)
```
glyn: I'm not sure this is always advisable. For instance, if a label selector doesn't match a unique resource, then it may be necessary to use a static name.

rfc/rfc-0006-child-resource-names.md
```diff
@@ -0,0 +1,43 @@
...
+- generate the name
+
+Static names may use templates, but must be deterministic. based only on stable values of the parent resource (e.g. `sprintf('%s-suffix', current.name)`).
+
+If a static name collides with an existing resource, the parent resource should reflect the collision in its conditions. 
+
+The name generally should not be chosen by the creator of the parent resource, as it introduces complexity when reconciling updates for names changes as Kubernetes cannot move, or rename a resource.
+
+### User Impact
+No public interface is changing, as the names of child resources are reflected on the parent resource's status. That will continue to be true.
```
jldec: Changing controller behavior to make some names deterministic when they are not deterministic today (or vice versa) does impact users. Isn't this the [issue](https://github.com/projectriff/system/issues/216) which triggered this RFC?

rfc/rfc-0006-child-resource-names.md
glyn: If the catch-all is to generate the name, then each of the other criteria should be about when to use a static name.

rfc/rfc-0006-child-resource-names.md
jldec: Users upgrading from riff v0.4 will be impacted depending on whether we continue generating non-deterministic names for services owned by deployers, or not. E.g. At the moment, the v0.5 snapshot controller breaks the assumption documented [here](https://projectriff.io/docs/v0.4/getting-started/docker-for-mac#create-a-core-deployer) that the service name can be inferred from the deployer name. See also https://github.com/projectriff/system/pull/214

rfc/rfc-0006-child-resource-names.md
sbawaska: This does not read like a problem statement to me. may be something like "we need consistency in naming all the generated resources".

rfc/rfc-0006-child-resource-names.md
scothis: List queries are inherently able to return zero-to-many resources. It may still be desirable to use a listing query to find a resource even if only a single resource is valid.

For example, our controllers often list resources controlled by the resource being reconciled to find the child. We expect either zero or one object to exist, but if multiple objects exist then we can handle that by deleting the superfluous object.

rfc/rfc-0006-child-resource-names.md
scothis: That falls under the category of a deterministic template. It's fine for the child resource name to be defined by an immutable aspect of the parent resource. Since resources can't be renamed, using the name of the parent is safe. Otherwise, there could only ever be a single instance of a named resource, since that value would always collide with the second resource.

rfc/rfc-0006-child-resource-names.md
scothis: If/when we decide to change that behavior, maybe. The contract for our resources allows children to be discovered in two ways:
1. the name of is exposed in the parent's status
2. the child can be queried by a label selector

Since a static name is not part of that contract, changing the strategy for naming resources is not a breaking change (nice how that works).

...the notable exception being the core Deployer including .spec.serviceName, which slipped in while I was on holiday. Yes, that was the direct impetus for this RFC, but defining these convention is something we should be doing anyway.

rfc/rfc-0006-child-resource-names.md
scothis: @glyn the general idea is that we should favor using generated names. Even if the item that leads to choosing a static name is true, there may be a superseding condition that says the resource should still use a generated name. For example, not all Services should us static names, even thought the name of a service is used to expose a DNS record that can be consumed by other resources.

rfc/rfc-0006-child-resource-names.md
```diff
@@ -0,0 +1,40 @@
...
+
+**Superseded by:** N/A
+
+**Supersedes:** N/A
+
+**Related:** N/A
+
+
+## Problem
+When a controller creates a new resource, it needs to either pick a static name or let Kubernetes generate a name. Static names are predictable, but may already be in use. Generated names cannot be predicated, but are guaranteed to be available. As the authors of reconcilers, we must define how each controlled resource is named.
```
sbawaska:
```suggestion
When a controller creates a new resource, it needs to either pick a static name or let Kubernetes generate a name. Static names are predictable, but may already be in use. Generated names cannot be predicted, but are guaranteed to be available. As the authors of reconcilers, we must define how each controlled resource is named.
```

rfc/rfc-0006-child-resource-names.md
scothis: The docs you linked show looking up the service name for the deployer, and then show making a request to that service. Some users will assume that any pattern that happens to be true is intentional, but as you say, they were already broken anyway.

As for projectriff/system#214, Glyn has already reverted the change introduced from that issue in favor of a static name, which is encouraged by this RFC.

rfc/rfc-0006-child-resource-names.md
scothis: refined the problem.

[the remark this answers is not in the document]
scothis: A child resource is created by a reconciler for its parent. The child will almost certainly have an owner reference to the parent, but that's not required. If the child should survive the deletion of the parent, the owner reference will not be set.

I haven't seen the term "child" used widely in kubernetes, but it's the term that kubebuilder uses, and it seems natural to me.

As an example, a Deployment creates a child ReplicaSet, which creates a child Pod. While the Pod is defended from the Deployment, it is not a child of Deployment.

rfc/rfc-0006-child-resource-names.md
```diff
@@ -0,0 +1,40 @@
+# RFC-0006: Child Resource Names
+
+**Authors:** Scott Andrews
+
+**Status:**
```
scothis:
```suggestion
**Status:** Accepted
```

PR #1370, opened 2019-12-02
RFC 0007 Guidance for resource references

scothis:
Comments due by Friday, December 6th

### DIFF
--- a/rfc/rfc-0007-resource-references.md
+++ b/rfc/rfc-0007-resource-references.md
@@ -4,7 +4,7 @@
 
 **Status:**
 
-**Pull Request URL:** link
+**Pull Request URL:** [#1370](https://github.com/projectriff/riff/pull/1370)
 
 **Superseded by:** N/A
 
--- a/rfc/rfc-0007-resource-references.md
+++ b/rfc/rfc-0007-resource-references.md
@@ -26,7 +26,7 @@ The Kubernetes API project provides two mechanisms for referencing other resourc
 
 [LocalObjectReference](https://godoc.org/k8s.io/api/core/v1#LocalObjectReference) contains only a resource's `name`. The `apiVersion`, `kind` and `namespace` must be inferred.
 
-riff should favor using ObjectReferences when the kind or namespace may vary or is unknown. LocalObjectReferences should be used when only the name of the resource will vary. A LocalObjectReference may be upgraded to an ObjectReference if explicitness is desired, however, it imposes a higher burden on those using the resource. `nil` should replace the reference if the relationship is not defined.
+riff should favor using ObjectReferences when the kind or namespace may vary or is unknown. LocalObjectReferences should be used when only the name of the resource will vary. A LocalObjectReference may be upgraded to an ObjectReference if explicitness is desired, however, it imposes a higher burden on those using the resource. `nil` should replace the reference if the referent is not defined.
 
 A validating webhook (or equivalent) should validate ObjectReferences before accepting the resource. The validation may restrict which fields are required and/or prohibited (`name` is typically required, `revision` is typically prohibited). The `apiVersion` and `kind` may be left open or restricted to known values.
 
--- a/rfc/rfc-0007-resource-references.md
+++ b/rfc/rfc-0007-resource-references.md
@@ -2,7 +2,7 @@
 
 **Authors:** Scott Andrews
 
-**Status:**
+**Status:** Accepted
 
 **Pull Request URL:** [#1370](https://github.com/projectriff/riff/pull/1370)
 
--- a/rfc/rfc-0007-resource-references.md
+++ b/rfc/rfc-0007-resource-references.md
@@ -1,0 +1,40 @@
+# RFC-0007: Resource References
+
+**Authors:** Scott Andrews
+
+**Status:**
+
+**Pull Request URL:** link
+
+**Superseded by:** N/A
+
+**Supersedes:** N/A
+
+**Related:** [RFC 0006: Guidance for child resource names](https://github.com/projectriff/riff/pull/1369)
+
+
+## Problem
+It's common for one resource to reference another. For example, a Deployer can reference a Function build to obtain the latest image, or a stream provider references the gateway Service it created. riff should have a common way to express relationships between Kubernetes resources.
+
+### Anti-Goals
+This RFC is not define how specific riff resources reference other resources, it is only providing guidance.
+
+## Solution
+The Kubernetes API project provides two mechanisms for referencing other resources. ObjectReference and LocalObjectReference.
+
+[ObjectReference](https://godoc.org/k8s.io/api/core/v1#ObjectReference) has full support for referencing any k8s resource, commonly by apiGroup, kind, namespace, and name.
+
+[LocalObjectReference](https://godoc.org/k8s.io/api/core/v1#LocalObjectReference) contains only a resource's name. The apiGroup, kind and namespace must be inferred.
+
+riff should favor using ObjectReferences when the kind or namespace may vary or is unknown. LocalObjectReferences should be used when only the name of the resource will vary. `nil` should be replace reference if the relationship is not defined.
+
+### User Impact
+Other tools in the k8s ecosystem are more likely to understand and able to traverse (Local)ObjectReferences than ad hoc reference. Replacing a string with a LocalObjectReference incurs a trivial performance overhead to marshal an object instead.
+
+### Backwards Compatibility and Upgrade Path
+Changing existing ad hoc references to use a (Local)ObjectReference is a breaking change. The impact of such a change should be considered before implementation.
+
+While the riff APIs are in alpha breaking changes without incrementing the API version are permissible.
+
+## FAQ
+*Answers to common questions that may arise and those that you’ve commonly been asked after requesting comments for this proposal.*
--- a/rfc/rfc-0007-resource-references.md
+++ b/rfc/rfc-0007-resource-references.md
@@ -22,11 +22,13 @@ This RFC does not define how specific riff resources reference other resources. 
 ## Solution
 The Kubernetes API project provides two mechanisms for referencing other resources: ObjectReference and LocalObjectReference.
 
-[ObjectReference](https://godoc.org/k8s.io/api/core/v1#ObjectReference) has full support for referencing any k8s resource, commonly by apiGroup, kind, namespace, and name.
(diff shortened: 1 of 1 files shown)


rfc/rfc-0007-resource-references.md
```diff
@@ -0,0 +1,40 @@
+# RFC-0007: Resource References
+
+**Authors:** Scott Andrews
+
+**Status:**
+
+**Pull Request URL:** link
```
scothis:
```suggestion
**Pull Request URL:** [#1370](https://github.com/projectriff/riff/pull/1370)
```

rfc/rfc-0007-resource-references.md
```diff
@@ -0,0 +1,40 @@
...
+**Supersedes:** N/A
+
+**Related:** [RFC 0006: Guidance for child resource names](https://github.com/projectriff/riff/pull/1369)
+
+
+## Problem
+It's common for one resource to reference another. For example, a Deployer can reference a Function build to obtain the latest image, or a stream provider references the gateway Service it created. riff should have a common way to express relationships between Kubernetes resources.
+
+### Anti-Goals
+This RFC is not define how specific riff resources reference other resources, it is only providing guidance.
```
markfisher:
```suggestion
This RFC does not define how specific riff resources reference other resources. It is only providing guidance.
```

rfc/rfc-0007-resource-references.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Solution
+The Kubernetes API project provides two mechanisms for referencing other resources. ObjectReference and LocalObjectReference.
```
markfisher:
```suggestion
The Kubernetes API project provides two mechanisms for referencing other resources: ObjectReference and LocalObjectReference.
```

rfc/rfc-0007-resource-references.md
```diff
@@ -0,0 +1,40 @@
...
+
+[ObjectReference](https://godoc.org/k8s.io/api/core/v1#ObjectReference) has full support for referencing any k8s resource, commonly by apiGroup, kind, namespace, and name.
+
+[LocalObjectReference](https://godoc.org/k8s.io/api/core/v1#LocalObjectReference) contains only a resource's name. The apiGroup, kind and namespace must be inferred.
+
+riff should favor using ObjectReferences when the kind or namespace may vary or is unknown. LocalObjectReferences should be used when only the name of the resource will vary. `nil` should be replace reference if the relationship is not defined.
```
markfisher:
```suggestion
riff should favor using ObjectReferences when the kind or namespace may vary or is unknown. LocalObjectReferences should be used when only the name of the resource will vary. `nil` should replace the reference if the relationship is not defined.
```

rfc/rfc-0007-resource-references.md
```diff
@@ -0,0 +1,40 @@
...
+
+### User Impact
+Other tools in the k8s ecosystem are more likely to understand and able to traverse (Local)ObjectReferences than ad hoc reference. Replacing a string with a LocalObjectReference incurs a trivial performance overhead to marshal an object instead.
```
markfisher:
```suggestion
Other tools in the k8s ecosystem are more likely to understand and be able to traverse (Local)ObjectReferences than ad hoc references. Replacing a string with a LocalObjectReference incurs a trivial performance overhead to marshal an object instead.
```

rfc/rfc-0007-resource-references.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Backwards Compatibility and Upgrade Path
+Changing existing ad hoc references to use a (Local)ObjectReference is a breaking change. The impact of such a change should be considered before implementation.
+
+While the riff APIs are in alpha breaking changes without incrementing the API version are permissible.
```
markfisher:
```suggestion
While the riff APIs are in alpha, breaking changes without incrementing the API version are permissible.
```

rfc/rfc-0007-resource-references.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Anti-Goals
+This RFC is not define how specific riff resources reference other resources, it is only providing guidance.
+
+## Solution
+The Kubernetes API project provides two mechanisms for referencing other resources. ObjectReference and LocalObjectReference.
+
+[ObjectReference](https://godoc.org/k8s.io/api/core/v1#ObjectReference) has full support for referencing any k8s resource, commonly by apiGroup, kind, namespace, and name.
+
+[LocalObjectReference](https://godoc.org/k8s.io/api/core/v1#LocalObjectReference) contains only a resource's name. The apiGroup, kind and namespace must be inferred.
```
dprotaso: I haven't seen uses of `LocalObjectReference` - do you have examples?

rfc/rfc-0007-resource-references.md
jldec: You mention "ad hoc" references later, and i assume that we use "ad hoc" references in some form today - shouldn't we include that as one of the ways to create references, and provide an example of that and an explanation of what the issues are with that?

rfc/rfc-0007-resource-references.md
jldec: While providing guidance is good, I'd suggest being more explicit here about what in riff today is motivating this, and what the likely impact of following these guidelines will be. E.g. It's difficult to infer whether this RFC is talking about issues with container-ref/application-ref/function-ref, or something else.

rfc/rfc-0007-resource-references.md
glyn: Note that the `TODO` in the `LocalObjectReference` struct may result in `LocalObjectReference` being deprecated in favour of `ObjectReference`.

scothis: @dprotaso https://sourcegraph.com/github.com/kubernetes/api/-/blob/core/v1/types.go#L5024:6&tab=references

scothis:
> (RFC meta comment - is there a way for others to commit grammar/spelling corrections - they add noise in the PR file viewer) ?

Suggestions feel appropriate for this purpose. As an RFC author, I don't want to deal with the conflicts that will be created by other people making changes directly. Conflicts in prose are harder to resolve than code.

rfc/rfc-0007-resource-references.md
scothis: Some RFCs are in the moment, while others are timeless. I've been actively avoiding being too specific about how the RFC applies directly to the current riff codebase, because those references will go stale while the principals will live on. I'm happy to talk about the current state of riff in the comments on the PR.

> It's difficult to infer whether this RFC is talking about issues with container-ref/application-ref/function-ref, or something else.

The current build field on the DeployerSpec/AdapterSpec/ProcessorSpec would be an excellent candidate for being replaced by an ObjectReference. Fields like `*Name` in the status would likewise be a strong candidate for a LocalObjectReference. 

Of course, since this RFC is only providing guidance and not mandating changes, we can approach each of these decisions individually, and more importantly, normalize future behavior.

rfc/rfc-0007-resource-references.md
scothis: by an ad hoc reference, I mean a resource reference that don't follow the proposed solution for referencing objects. Since that's not part of the proposed solution, it doesn't make sense to me to include it under the solution. The aim of this RFC is to minimize the number of references not-following the proposed solution compared to the number that do over time.

rfc/rfc-0007-resource-references.md
```diff
@@ -0,0 +1,42 @@
...
+This RFC does not define how specific riff resources reference other resources. It is only providing guidance.
+
+## Solution
+The Kubernetes API project provides two mechanisms for referencing other resources: ObjectReference and LocalObjectReference.
+
+[ObjectReference](https://godoc.org/k8s.io/api/core/v1#ObjectReference) has full support for referencing any k8s resource, commonly by `apiVersion`, `kind`, `namespace`, and `name`.
+
+[LocalObjectReference](https://godoc.org/k8s.io/api/core/v1#LocalObjectReference) contains only a resource's `name`. The `apiVersion`, `kind` and `namespace` must be inferred.
+
+riff should favor using ObjectReferences when the kind or namespace may vary or is unknown. LocalObjectReferences should be used when only the name of the resource will vary. A LocalObjectReference may be upgraded to an ObjectReference if explicitness is desired, however, it imposes a higher burden on those using the resource. `nil` should replace the reference if the relationship is not defined.
```
glyn: I'd like to think the _relationship_ is always defined, but the _referent_ may not exist.
```suggestion
riff should favor using ObjectReferences when the kind or namespace may vary or is unknown. LocalObjectReferences should be used when only the name of the resource will vary. A LocalObjectReference may be upgraded to an ObjectReference if explicitness is desired, however, it imposes a higher burden on those using the resource. `nil` should replace the reference if the referent is not defined.
```

rfc/rfc-0007-resource-references.md
```diff
@@ -0,0 +1,42 @@
...
+
+[ObjectReference](https://godoc.org/k8s.io/api/core/v1#ObjectReference) has full support for referencing any k8s resource, commonly by `apiVersion`, `kind`, `namespace`, and `name`.
+
+[LocalObjectReference](https://godoc.org/k8s.io/api/core/v1#LocalObjectReference) contains only a resource's `name`. The `apiVersion`, `kind` and `namespace` must be inferred.
+
+riff should favor using ObjectReferences when the kind or namespace may vary or is unknown. LocalObjectReferences should be used when only the name of the resource will vary. A LocalObjectReference may be upgraded to an ObjectReference if explicitness is desired, however, it imposes a higher burden on those using the resource. `nil` should replace the reference if the referent is not defined.
+
+A validating webhook (or equivalent) should validate ObjectReferences before accepting the resource. The validation may restrict which fields are required and/or prohibited (`name` is typically required, `revision` is typically prohibited). The `apiVersion` and `kind` may be left open or restricted to known values.
```
dprotaso: > *may* validate ObjectReferences



https://github.com/knative/serving/pull/5744

scothis: @dprotaso this isn't intended to be an assertion of whether or not the referenced resource exists, but whether the reference itself is semantically valid

dprotaso: sounds good - I don't have an approve button so go nuts

dprotaso: found it

rfc/rfc-0007-resource-references.md
```diff
@@ -0,0 +1,42 @@
+# RFC-0007: Resource References
+
+**Authors:** Scott Andrews
+
+**Status:**
```
scothis:
```suggestion
**Status:** Accepted
```

PR #1372, opened 2019-12-09
Controller unit testing RFC 0009

glyn:


scothis:
This is similar to the pattern we follow for testing commands in the CLI. Which it should because the CLI tests are modeled after the Knative reconciler tests. The main difference is that the trigger is a reconciliation key rather than a cli invocation and in our case, the client interface is different since we're building on controller-runtime.

See: https://github.com/projectriff/cli/blob/master/pkg/testing/command_table.go

Usage: https://github.com/projectriff/cli/blob/master/pkg/build/commands/function_create_test.go#L282-L1103

fbiville:
As discussed on projectriff.slack.com, I am overall a bit skeptical about the matrix approach for testing, as the number of attributes of the test cases seem to grow over time to a point where it sometimes becomes hard to understand which attribute to use or not for a particular situation.

Moreover, the context is spread between the custom test runner and the relevant test case one is interested in. It may be difficult to understand a test case without first understanding what the custom test runner does.

It is also not possible to improve a particular matrix test case. Structural changes apply to all of them.

In my opinion, “traditional” tests offer a greater flexibility, with a smaller learning curve (I assume most developers are familiar with given-when-then structures + before/after hooks).

Maybe there is some solution to be found in the middle, i.e. standard enough so that we do not duplicate code for common tasks and flexible enough to allow local one-off variations without surfacing these subtleties to the general "framework".

rfc/rfc-0009-controller-unit-testing.md
```diff
@@ -0,0 +1,83 @@
...
+
+The approach taken is to use a fake client and other interfaces and introduce a test framework
+which allows tests to be written in a declarative style. See "Rejected Alternatives" below for 
+some other approaches which were tried out.
+
+The test framework supports table-driven tests. Each row of the table tests a single call to `Reconcile` by
+setting up some initial Kubernetes objects, modifying the client behaviour using various hooks
+or reactors, and then asserting that various Kubernetes objects are created, updated, etc.
+
+When a helper method is used to construct a Kubernetes object, it should construct a fixed value and any variations
```
scothis: Clarify what a helper method is in this context

rfc/rfc-0009-controller-unit-testing.md
```diff
@@ -0,0 +1,83 @@
...
+Also, a local control plane is not
+deterministic when it generates object names, which complicates the setting up of test
+expectations.
+
+See here for a quick
+[spike](https://github.com/projectriff/system/pull/232) using a local control plane.
+
+### Depending on Knative testing packages
+Another approach which was quickly [spiked](https://github.com/projectriff/system/pull/233) and
+rejected was to depend on Knative testing packages. This was abandoned because of dependency
```
scothis: Knative also uses typed clientsets, which is not the same client interface we're using from controller runtime.

rfc/rfc-0009-controller-unit-testing.md
glyn: Clarified. These are actually helper functions rather than methods.

rfc/rfc-0009-controller-unit-testing.md
glyn: Good point. Changed.

glyn:
@fbiville The table approach seems to be working out quite well so far in that the test framework seems to be reusable across controllers. If we had to add features to the framework for particular controllers, that would be more of a concern.

My main concern about the table approach was that introducing minor variations on existing "rows" would result in a lot of duplication between rows, but that is being avoided by using helper functions to avoid excessive duplication. For example, @scothis has introduced a nice collection of helper functions for constructing certain common objects, in `pkg/controllers/testing/factories`.

I share your concern about the table approach being a little impenetrable for newcomers, but I think that's mitigated by using the same test framework across all the controller tests. Also, I think the table tests are pretty easy to read once you've grokked the general design of reconcilers (admittedly a bit of a hurdle for newcomers). I think using non-table tests wouldn't really help much as there is a lot of machinery associated with testing controllers and non-table tests would be more likely to diverge from each other in the use of such machinery.

rfc/rfc-0009-controller-unit-testing.md
```diff
@@ -0,0 +1,84 @@
...
+or reactors, and then asserting that various Kubernetes objects are created, updated, etc.
+
+When a helper function (a utility function provided by a test) is used to construct a Kubernetes object, it should
+construct a fixed value and any variations should be explicitly coded in the table row. This avoids details of a
+particular test being "buried" in helper functions.
+
+Since kubebuilder was used to generate the controllers, some packages contain more than one controller and so
+helper functions need to be named or organised carefully to avoid clashes. 
+
+## Rejected Alternatives
```
fbiville: maybe add the approach with Gomega/Ginkgo to the list?

rfc/rfc-0009-controller-unit-testing.md
```diff
@@ -0,0 +1,84 @@
...
+Apart from checking that various objects are created etc., it should be straigthtforward
+to test error paths by injecting errors in object lookup and creation.
+
+### Anti-Goals
+
+None.
+
+## Solution
+
+The approach taken is to use a fake client and other interfaces and introduce a test framework
```
fbiville: should other interfaces be strictly defined here?

fbiville:
```suggestion
The approach taken is to use a fake Kubernetes client and other interfaces and introduce a test framework
```

rfc/rfc-0009-controller-unit-testing.md
```diff
@@ -0,0 +1,84 @@
...
+which allows tests to be written in a declarative style. See "Rejected Alternatives" below for 
+some other approaches which were tried out.
+
+The test framework supports table-driven tests. Each row of the table tests a single call to `Reconcile` by
+setting up some initial Kubernetes objects, modifying the client behaviour using various hooks
+or reactors, and then asserting that various Kubernetes objects are created, updated, etc.
+
+When a helper function (a utility function provided by a test) is used to construct a Kubernetes object, it should
```
fbiville: should factories be enforced instead?

rfc/rfc-0009-controller-unit-testing.md
```diff
@@ -0,0 +1,84 @@
...
+None.
+
+## Solution
+
+The approach taken is to use a fake client and other interfaces and introduce a test framework
+which allows tests to be written in a declarative style. See "Rejected Alternatives" below for 
+some other approaches which were tried out.
+
+The test framework supports table-driven tests. Each row of the table tests a single call to `Reconcile` by
+setting up some initial Kubernetes objects, modifying the client behaviour using various hooks
```
fbiville: should hook and reactor be defined here?

rfc/rfc-0009-controller-unit-testing.md
```diff
@@ -0,0 +1,84 @@
...
+or reactors, and then asserting that various Kubernetes objects are created, updated, etc.
```
fbiville: tracking requests should be mentioned, I think

rfc/rfc-0009-controller-unit-testing.md
```diff
@@ -0,0 +1,84 @@
...
+
+For the purpose of this document, unit testing refers to testing the
+`Reconcile` method of a controller.
+
+Apart from checking that various objects are created etc., it should be straigthtforward
+to test error paths by injecting errors in object lookup and creation.
+
+### Anti-Goals
+
+None.
```
fbiville: what about the execution mode (as in serial vs. parallel)? It feels out of scope for this spec (hence an anti-goal) but could be specified later?

rfc/rfc-0009-controller-unit-testing.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Problem
+The Kubernetes controllers in riff need unit testing in a consistent way. Other projects
+have adopted and promoted various ways of testing controllers. For ease of maintenance,
+riff needs to adopt a single, consistent approach.
+
+For the purpose of this document, unit testing refers to testing the
+`Reconcile` method of a controller.
+
+Apart from checking that various objects are created etc., it should be straigthtforward
```
fbiville:
```suggestion
Apart from checking that various objects are created etc., it should be straightforward
```

rfc/rfc-0009-controller-unit-testing.md
fbiville: is it out of scope to talk about the developer experience of the test runner here? 
I'm thinking about the ability to focus on/skip a test, error reporting when tests fail.

fbiville: or, wrapping the 2 previous comments, that an anti-goal of this RFC is to specify the capabilities of the underlying custom test runner

rfc/rfc-0009-controller-unit-testing.md
glyn: Added. I think that helps to explain the rationale for the table-driven approach too.

rfc/rfc-0009-controller-unit-testing.md
glyn: Adopted.

glyn: Probably too much detail. I've remove mention of other interfaces.

rfc/rfc-0009-controller-unit-testing.md
glyn: I've discussed this with @scothis and am planning a quick spike to explore it.

rfc/rfc-0009-controller-unit-testing.md
glyn: Fixed.

rfc/rfc-0009-controller-unit-testing.md
glyn: Yes, I'll add an anti-goal.

rfc/rfc-0009-controller-unit-testing.md
glyn: Added.

rfc/rfc-0009-controller-unit-testing.md
glyn: I think those details are covered by the anti-goal. :-)

rfc/rfc-0009-controller-unit-testing.md
glyn: Again, this is a detail covered by the anti-goal.

PR #1373, opened 2020-01-08
RFC 0010: Drop core runtime

scothis:
Feedback is open until Tuesday, January 14th

### DIFF
--- a/rfc/rfc-0010-drop-core-runtime.md
+++ b/rfc/rfc-0010-drop-core-runtime.md
@@ -2,7 +2,7 @@
 
 **Authors:** @scothis
 
-**Status:**
+**Status:** Rejected
 
 **Pull Request URL:** [#1373](https://github.com/projectriff/riff/pull/1373)
 
--- a/rfc/rfc-0010-drop-core-runtime.md
+++ b/rfc/rfc-0010-drop-core-runtime.md
@@ -14,7 +14,7 @@
 
 ## Problem
 
-riff currently has two ways to deploy HTTP triggered workload, the Core and Knative runtimes. This dichotomy causes users extra cognitive load. For every operation they have to think about which runtime they are targeting. Moreover, the capabilities and behavior of each runtime are subtly different.
+riff currently has two ways to deploy HTTP triggered workloads: the Core and Knative runtimes. This dichotomy causes users extra cognitive load. For every operation they have to think about which runtime they are targeting. Moreover, the capabilities and behavior of each runtime are subtly different.
 
 As the Knative runtime is a super-set of the Core runtime, we should consolidate on the Knative runtime. Users who don't want to run Knative, can take a riff built container image and deploy it however they like.
 
@@ -30,7 +30,7 @@ Users would no longer be able to deploy workloads to the Core runtime. Of course
 
 ### Backwards Compatibility and Upgrade Path
 
-This is a breaking change, users will need to create the equivalent Knative runtime Deployer.
+This is a breaking change, users will need to create the equivalent of the Knative runtime Deployer.
 
 ## FAQ
 
--- a/rfc/rfc-0010-drop-core-runtime.md
+++ b/rfc/rfc-0010-drop-core-runtime.md
@@ -4,7 +4,7 @@
 
 **Status:**
 
-**Pull Request URL:** link
+**Pull Request URL:** [#1373](https://github.com/projectriff/riff/pull/1373)
 
 **Superseded by:** N/A
 
--- a/rfc/rfc-0010-drop-core-runtime.md
+++ b/rfc/rfc-0010-drop-core-runtime.md
@@ -14,7 +14,7 @@
 
 ## Problem
 
-riff currently has two ways to deploy HTTP triggered workloads: the Core and Knative runtimes. This dichotomy causes users extra cognitive load. For every operation they have to think about which runtime they are targeting. Moreover, the capabilities and behavior of each runtime are subtly different.
+riff currently has two ways to deploy HTTP triggered workloads: the Core and Knative runtimes. This dichotomy causes users extra cognitive load. For every operation they have to think about which runtime they are targeting. Moreover, the capabilities and behavior of each runtime are subtly different. For developers, having one less runtime will reduce development, QA and support overhead.
 
 As the Knative runtime is a super-set of the Core runtime, we should consolidate on the Knative runtime. Users who don't want to run Knative, can take a riff built container image and deploy it however they like.
 
--- a/rfc/rfc-0010-drop-core-runtime.md
+++ b/rfc/rfc-0010-drop-core-runtime.md
@@ -1,0 +1,37 @@
+# RFC-#: Drop Core Runtime
+
+**Authors:** @scothis
+
+**Status:**
+
+**Pull Request URL:** link
+
+**Superseded by:** N/A
+
+**Supersedes:** N/A
+
+**Related:** N/A
+
+## Problem
+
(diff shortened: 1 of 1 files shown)


rfc/rfc-0010-drop-core-runtime.md
```diff
@@ -0,0 +1,37 @@
...
+
+**Superseded by:** N/A
+
+**Supersedes:** N/A
+
+**Related:** N/A
+
+## Problem
+
+riff currently has two ways to deploy HTTP triggered workload, the Core and Knative runtimes. This dichotomy causes users extra cognitive load. For every operation they have to think about which runtime they are targeting. Moreover, the capabilities and behavior of each runtime are subtly different.
```
markfisher:
```suggestion
riff currently has two ways to deploy HTTP triggered workloads: the Core and Knative runtimes. This dichotomy causes users extra cognitive load. For every operation they have to think about which runtime they are targeting. Moreover, the capabilities and behavior of each runtime are subtly different.
```

rfc/rfc-0010-drop-core-runtime.md
```diff
@@ -0,0 +1,37 @@
...
+
+As a half measure, we could deprecate the Core runtime in riff 0.5, remove documentation from projectriff.io and hide the CLI commands. With full removal happening in riff 0.6.
+
+### User Impact
+
+Users would no longer be able to deploy workloads to the Core runtime. Of course a user could take a built image and create their own Deployment and Service to provide the equivalent of the Core runtime.
+
+### Backwards Compatibility and Upgrade Path
+
+This is a breaking change, users will need to create the equivalent Knative runtime Deployer.
```
markfisher:
```suggestion
This is a breaking change, users will need to create the equivalent of the Knative runtime Deployer.
```

rfc/rfc-0010-drop-core-runtime.md
```diff
@@ -0,0 +1,37 @@
...
+
+As a half measure, we could deprecate the Core runtime in riff 0.5, remove documentation from projectriff.io and hide the CLI commands. With full removal happening in riff 0.6.
+
+### User Impact
+
+Users would no longer be able to deploy workloads to the Core runtime. Of course a user could take a built image and create their own Deployment and Service to provide the equivalent of the Core runtime.
+
+### Backwards Compatibility and Upgrade Path
+
+This is a breaking change, users will need to create the equivalent of the Knative runtime Deployer.
```
trisberg: This is a breaking change, users will need to use a Knative runtime Deployer instead of a Core one. Functions and apps deployed with the Core deployer never scaled to zero, so to mimic this behavior there is an option to set `--min-scale=1` for the Knative deployer to prevent the scaling to 0.

rfc/rfc-0010-drop-core-runtime.md
```diff
@@ -0,0 +1,37 @@
...
+**Supersedes:** N/A
+
+**Related:** N/A
+
+## Problem
+
+riff currently has two ways to deploy HTTP triggered workloads: the Core and Knative runtimes. This dichotomy causes users extra cognitive load. For every operation they have to think about which runtime they are targeting. Moreover, the capabilities and behavior of each runtime are subtly different.
+
+As the Knative runtime is a super-set of the Core runtime, we should consolidate on the Knative runtime. Users who don't want to run Knative, can take a riff built container image and deploy it however they like.
+
```
trisberg: Should we then also rename the `knative` command to something like `serving` or `http` to avoid locking in to the specific name of the implementation used?

rfc/rfc-0010-drop-core-runtime.md
```diff
@@ -0,0 +1,37 @@
...
+**Pull Request URL:** [#1373](https://github.com/projectriff/riff/pull/1373)
+
+**Superseded by:** N/A
+
+**Supersedes:** N/A
+
+**Related:** N/A
+
+## Problem
```
glyn: Do we want to mention the extra cost of maintaining, testing, and supporting the core runtime in addition to the Knative runtime?

rfc/rfc-0010-drop-core-runtime.md
scothis: I've had the same thought, but is somewhat orthogonal to the decision here

rfc/rfc-0010-drop-core-runtime.md
scothis: The specifics of migrating core to knative are at a lower level than I'd like to address in an RFC, but yes.

rfc/rfc-0010-drop-core-runtime.md
scothis: added

rfc/rfc-0010-drop-core-runtime.md
trisberg: Maybe another RFC ;)

fbiville:
The core runtime had the advantage of being very simple to comprehend, offering a good "getting started", minimal experience.

Users should not have to know anything about Knative to deploy simple workloads. 
Therefore, we should rework the UX around the current knative deployer commands.

It may sound silly at first but what about keeping the core commands and re-implement them so that they create Knative resources with sensible defaults under the hood?

scothis:
> It may sound silly at first but what about keeping the core commands and re-implement them so that they create Knative resources with sensible defaults under the hood?

I'm not sure what you're proposing. Can you give a couple concrete examples? How would this be different than the current Knative runtime?

> Users should not have to know anything about Knative to deploy simple workloads.

It has never been a design goal to hide Knative from riff users. I'd say it's an anti-goal.

fbiville:
> It has never been a design goal to hide Knative from riff users. I'd say it's an anti-goal.

To me, riff is about allowing users to deploy (serverless) workloads to Kubernetes with as little friction as possible. 

Today, if I want to deploy a simple request-reply workload, I create a function (`riff function create...`) and a core deployer (`riff core deployer create...`) and I'm done.

Tomorrow, for the same thing, I would create the function and the Knative deployer. Then, I would need to understand what Knative Services, Routes and Revisions are and how to disable scale-to-0 (even autoscaling, maybe). That is significantly more complex than what I can do today with core.

I then have two questions:

1. are these simple workloads our target?
2. if they are, and considering that core is gone, do we plan to remove the "friction" I described above?

scothis:
To reparaphrase your comment:

Today, if I want to deploy a simple request-reply workload, I create a function (`riff function create...`) and a ~core~**knative** deployer (`riff knative deployer create...`) and I'm done.

> Tomorrow, for the same thing, I would create the function and the Knative deployer. Then, I would need to understand what the Knative autoscalerm, Services, Routes and Revisions are and how to disable scale-to-0 (even autoscaling, maybe). That is significantly more complex than what I can do today with core.

While I wouldn't recommend it for either runtime, a user can be equally naive about the implementation details of the core and knative runtimes. The surface area of the core Deployer and knative Deployer resources is by design nearly identical. The only different between the two is that the knative Deployer has knobs for min/max replicas. Since the core runtime doesn't have a scaler, there's no need for these knobs.

The core runtime is simpler because there are fewer moving parts, but that's because there are so many essential capabilities it doesn't provide. The core runtime, as is, is a day-2 dead end. As a user you must bring your own ingress, ssl termination, scaling. If you want that level of control, great, then also create a Deployment and Service.

trisberg:
There seems to be some subtle differences between services exposed by the knative runtime vs the core - our demo storefront app can't seem to access the inventory-api service if it's deployed with the knative deployer. I'll post some  more in the riff slack.

jldec:
I'm concerned that we are rushing (and maybe over-simplifying) this decision

IMO @fbiville is making a valid point that there is a loss of simplicity.

I wonder if this RFC isn't really about limitations of Knative (too little ability to configure runtime strategies) leading to issues with the design of riff runtimes (artificial dichotomy).

I suggest keeping this open and starting a discussion to explore improvements to the Knative design which enable something as simple as what we have in the riff Core runtime.

PR #1374, opened 2020-01-14
RFC 0012: source code upload

ekcasey:
[readable](https://github.com/ekcasey/riff/blob/source-code-upload/rfc/rfc-0012-source-code-upload.md)

Comments are open until Wednesday, February 19th.


rfc/rfc-0011-source-code-upload.md
```diff
@@ -0,0 +1,64 @@
...
+derivation of the target image tag, by appending a suffix to the tag. For example if the generated image will be published to `docker.io/my/app:latest`,
+the source image will be uploaded to `docker.io/my/app:latest-source`
+
+The following flags will allows for configurability:
+
+* `--source-image repository`
+
+#### Source Image Push Auth
+When pushing the source image `riff` will use the default docker keychain and will not try to use the registry secrets
+configure in `riff`
```
fbiville:
```suggestion
configured in `riff`
```

jldec:
Hi @ekcasey, would you mind changing the RFC number to 0012 in the filename, and prefixig this PR name with `RFC 0012` as well?

jldec:
@ekcasey bump

rfc/rfc-0014-source-code-upload.md
```diff
@@ -0,0 +1,66 @@
...
+1. Images built from local source will not get autmatically rebased or rebuilt based on stack or buildpack updates,
+ and are therefore susceptible to more security vulnerability
+1. Because build from source uses `pack`, it requires users to have a docker daemon
+1. It requires users to wait for large builder images to pull
+1. Results produced by pack and kpack should be identical but difference introduces risk
+
+## Solution
+
+#### Registry Type SourceResolver
+When a user specificies the `--local-path <dir>` flag during `riff function create` or `riff application create` the source
```
fbiville:
```suggestion
When a user specifies the `--local-path <dir>` flag during `riff function create` or `riff application create` the source
```

rfc/rfc-0014-source-code-upload.md
```diff
@@ -0,0 +1,66 @@
...
+code in the specified directory will be packaged as a OCI image containing a single layer and uploaded to a docker registry.
+`riff` with create a `kpack` `image` with a `registry` type `sourceresolver` https://github.com/pivotal/kpack/blob/master/docs/image.md#source-configuration.
+`kpack` can than build the image as it does in the git source type case.
```
fbiville:
```suggestion
`kpack` can then build the image as it does in the git source type case.
```

rfc/rfc-0014-source-code-upload.md
```diff
@@ -0,0 +1,66 @@
+# RFC-0011: Source Code Upload
```
trisberg: should this be RFC-0014? what happened to RFC-0012?

rfc/rfc-0014-source-code-upload.md
```diff
@@ -0,0 +1,66 @@
...
+1. Because build from source uses `pack`, it requires users to have a docker daemon
+1. It requires users to wait for large builder images to pull
+1. Results produced by pack and kpack should be identical but difference introduces risk
+
+## Solution
+
+#### Registry Type SourceResolver
+When a user specifies the `--local-path <dir>` flag during `riff function create` or `riff application create` the source
+code in the specified directory will be packaged as a OCI image containing a single layer and uploaded to a docker registry.
+`riff` with create a `kpack` `image` with a `registry` type `sourceresolver` https://github.com/pivotal/kpack/blob/master/docs/image.md#source-configuration.
```
trisberg: `riff` will create ...

rfc/rfc-0014-source-code-upload.md
```diff
@@ -0,0 +1,66 @@
...
+
+The following flags will allow for configurability:
+
+* `--source-image repository`
+
+#### Source Image Push Auth
+When pushing the source image `riff` will use the default docker keychain and will not try to use the registry secrets configured in `riff`
+
+#### Source Image Pull Auth
+Credentials for the source image must be present in the riff service account so that kpack can pull the source code
```
trisberg: This sounds like an additional manual step for someone using non-public registry like GCR or can we add something to the `riff credentials` command?

ekcasey: I believe the existing `riff credentials` commands should suffice. If I understand the `kpack` `build-init` keychain logic correctly, it should include all of the registry credentials from the service account.

The only thing users would need to do is apply another set of registry creds, if the registry they are using to store their source code is different from that to which they publish. If they are using the same regsistry for both images, no additional steps should be required.

rfc/rfc-0014-source-code-upload.md
ekcasey: I assumed we were numbering them in the order they merged. The last RFC is [0013](https://github.com/projectriff/riff/blob/master/rfc/rfc-0013-invoker-tck.md). I have no idea what happened to 0012.

Let me know if I have misunderstood the process.

jldec:
I know this is a bit nitpicky, but would anyone object if we named this RFC 0012?  
The hole in the sequence of [rfcs](https://github.com/projectriff/riff/tree/master/rfc) is causing me lost sleep.

rfc/rfc-0014-source-code-upload.md
```diff
@@ -0,0 +1,66 @@
...
+`riff` will respect include/exclude configuration from a [CNB project descriptor](https://github.com/buildpacks/rfcs/blob/master/text/0019-project-descriptor.md#buildinclude-and-buildexclude)
+if present 
+
+### User Impact
+
+#### Pros
+By making builds from local source behave more similarly to builds from a git source the behavior of `riff` will be less
+surprising to users. Users will also be able to receive CVE fixes for images built from local source through a declarative
+model, rather than requiring an imperative rebuild or rebase like the current model.
+
```
jldec: I would also add that removing the dependency on pack from the riff CLI will make it easier to widen our support to  different styles of build automation without losing the --local-path DX.

rfc/rfc-0014-source-code-upload.md
scothis: the `riff credential` commands are already able to handle secrets from multiple registries

rfc/rfc-0012-source-code-upload.md
```diff
@@ -0,0 +1,66 @@
...
+ and are therefore susceptible to more security vulnerability
+1. Because build from source uses `pack`, it requires users to have a docker daemon
+1. It requires users to wait for large builder images to pull
+1. Results produced by pack and kpack should be identical but difference introduces risk
+
+## Solution
+
+#### Registry Type SourceResolver
+When a user specifies the `--local-path <dir>` flag during `riff function create` or `riff application create` the source
+code in the specified directory will be packaged as a OCI image containing a single layer and uploaded to a docker registry.
```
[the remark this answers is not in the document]
ekcasey: Good question. I thought there was a constraint but I looked into it and there isn't any. One layer might be the easiest way to get started. In the long run we should figure out how to optimize the number and contents of layers to reduce duplicate data transfer.

rfc/rfc-0014-source-code-upload.md
ekcasey: @scothis I think we are saying the same thing

ekcasey:
> Based on the subject change, I assume this RFC is open for comments. I didn't see that announced anywhere or a notice for when comments are due and we make a decision.

Yes, sorry I missed some steps here. It is now officially open for comments until Wednesday, February 19th. I updated the issue content.

PR #1376, opened 2020-01-14
RFC 0011: Streaming Gateways

scothis:
Comments are open until Thursday, January 16th.

### DIFF
--- a/rfc/rfc-0011-streaming-gateways.md
+++ b/rfc/rfc-0011-streaming-gateways.md
@@ -47,10 +47,6 @@ This RFC will not prevent naming collisions between different provisioners withi
 
    The `Stream` reconciler can track the `Gateway` and include the `Gateway`'s Ready condition as one of its conditions. `Processor`s already track each input and output `Stream`'s Ready condition to scale down the processor when a stream goes red.
 
-1. [maybe] Pin a `Stream` to a `Gateway` instance
-
-   `Stream`s are inherently stateful. A new `Gateway`, even if it has the same name, is unlikely to contain the same state. The gateway's UID could be saved on the stream and compared. if the UID changes, the stream's Ready condition could flip False to prevent usage of the stream.
-
 1. Update Streaming CRDs status for new relationships
 
   Changes to the reconciliation object graph invalidate many of the existing status references.
--- a/rfc/rfc-0011-streaming-gateways.md
+++ b/rfc/rfc-0011-streaming-gateways.md
@@ -15,7 +15,7 @@
 
 ## Problem
 
-The relationships between a `Stream` and a `(InMemory|Kafka|Pulsar)Provider` is fragile. The stream contains a hostname for the provider's provisioner, which the stream controller calls when reconciling the stream. This has a few problems:
+The current relationship between a `Stream` and a `(InMemory|Kafka|Pulsar)Provider` is fragile. The stream contains a hostname for the provider's provisioner, which the stream controller calls when reconciling the stream. This has a few problems:
 - the `Stream` resource doesn't know when the `Provider` is unavailable, or deleted
 - the `Stream` has no way to automatically obtain a new address for the provisioner
 - the provisioner name is competing in the `Service` namespace. The likelihood of a name collision is reduced by appending a suffix to the service name. However, the user must include this suffix when creating each stream
@@ -25,7 +25,7 @@ The problem is derived from [projectriff/system#86](https://github.com/projectri
 
 ### Anti-Goals
 
-This RFC will not prevent naming collisions between different provisioners with the same name. The first gateway of a given name wins.
+This RFC will not prevent naming collisions between different provisioners within the same namespace. The first gateway of a given name wins.
 
 ## Solution
 
@@ -35,11 +35,11 @@ This RFC will not prevent naming collisions between different provisioners with 
 
 1. Introduce a `Gateway` resource which creates a single `Deployment` and `Service`
 
-   This reduces the toil of adding a new `*Gateway` reconciler as the boilerplate code will be encapsulated. It is recommend that users not create this resource directly.
+   This reduces the toil of adding a new `*Gateway` reconciler as the boilerplate code will be encapsulated. It is recommended that users not create this resource directly. Rather one of these _abstract_ `Gateway`s will be created for each _concrete_ `*Gateway` resource within the latter's reconciler as described below.
    
    1. Cohabitate the gateway (liiklus) and provisioner `Deployment`s and `Service`s
 
-      A single `Deployment` with two containers and a single `Service` with two ports can replace the current two Deployments and two Services per provider.
+      A single `Deployment` with two containers and a single `Service` with two ports can replace the current two Deployments and two Services per provider. Eventually we may extend the gateway spec and implementation (currently liiklus) so that the provisioning role is handled by the same container (a new RFC would be created to capture that plan).
 
    1. Update each existing `*Gateway` to reconcile to a `Gateway`
 
@@ -53,7 +53,7 @@ This RFC will not prevent naming collisions between different provisioners with 
 
 1. Update Streaming CRDs status for new relationships
 
-  Changes to the reconciliation object graph invalidate may of the existing status references.
+  Changes to the reconciliation object graph invalidate many of the existing status references.
 
 ### User Impact
 
@@ -61,7 +61,7 @@ The CRDs and CLI commands for streaming providers will replace `provider` with `
 
 ### Backwards Compatibility and Upgrade Path
 
-All existing behavior is being preserved, since the streaming runtime has not been released, there is no release to be backwards incompatible with.
+All existing behavior is being preserved. Since the streaming runtime has not been released, there is no release to be backwards incompatible with.
 
 ## FAQ
 
--- a/rfc/rfc-0011-streaming-gateways.md
+++ b/rfc/rfc-0011-streaming-gateways.md
@@ -2,7 +2,7 @@
 
 **Authors:** @scothis, @markfisher, @ericbottard
 
-**Status:**
+**Status:** Accepted
 
 **Pull Request URL:** [#1376](https://github.com/projectriff/riff/pull/1376)
 
--- a/rfc/rfc-0011-streaming-gateways.md
+++ b/rfc/rfc-0011-streaming-gateways.md
@@ -4,7 +4,7 @@
 
 **Status:**
 
-**Pull Request URL:** link
+**Pull Request URL:** [#1376](https://github.com/projectriff/riff/pull/1376)
 
 **Superseded by:** N/A
 
--- a/rfc/rfc-0011-streaming-gateways.md
+++ b/rfc/rfc-0011-streaming-gateways.md
@@ -1,0 +1,68 @@
+# RFC-0011: Streaming Gateways
+
+**Authors:** @scothis, @markfisher, @ericbottard
+
+**Status:**
+
+**Pull Request URL:** link
+
+**Superseded by:** N/A
+
+**Supersedes:** N/A
+
+**Related:** N/A
+
+
+## Problem
+
+The relationships between a `Stream` and a `(InMemory|Kafka|Pulsar)Provider` is fragile. The stream contains a hostname for the provider's provisioner, which the stream controller calls when reconciling the stream. This has a few problems:
+- the `Stream` resource doesn't know when the `Provider` is unavailable, or deleted
+- the `Stream` has no way automatically obtain a new address for the provisioner
+- the provisioner name is completing in the `Service` namespace. The likelihood of a name collision is reduced by appending a suffix to the service name, however, the user must include this suffix when creating each stream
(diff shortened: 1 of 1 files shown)


rfc/rfc-0011-streaming-gateways.md
```diff
@@ -0,0 +1,68 @@
...
+**Supersedes:** N/A
+
+**Related:** N/A
+
+
+## Problem
+
+The relationships between a `Stream` and a `(InMemory|Kafka|Pulsar)Provider` is fragile. The stream contains a hostname for the provider's provisioner, which the stream controller calls when reconciling the stream. This has a few problems:
+- the `Stream` resource doesn't know when the `Provider` is unavailable, or deleted
+- the `Stream` has no way automatically obtain a new address for the provisioner
```
ericbottard:
```suggestion
- the `Stream` has no way to automatically obtain a new address for the provisioner
```

rfc/rfc-0011-streaming-gateways.md
```diff
@@ -0,0 +1,68 @@
...
+- the provisioner name is completing in the `Service` namespace. The likelihood of a name collision is reduced by appending a suffix to the service name, however, the user must include this suffix when creating each stream
```
ericbottard:
```suggestion
- the provisioner name is competing in the `Service` namespace. The likelihood of a name collision is reduced by appending a suffix to the service name. However, the user must include this suffix when creating each stream
```

rfc/rfc-0011-streaming-gateways.md
```diff
@@ -0,0 +1,68 @@
...
+- as new providers are created, there is a significant amount of reconciliation logic that is duplicated with trivial differences
+
+The problem is derived from [projectriff/system#86](https://github.com/projectriff/system/issues/86).
+
+### Anti-Goals
+
+This RFC will not prevent naming collisions between different provisioners with the same name. The first gateway of a given name wins.
```
ericbottard:
```suggestion
This RFC will not prevent naming collisions between different provisioners using the same stream name. The first gateway trying a given name wins.
```

rfc/rfc-0011-streaming-gateways.md
```diff
@@ -0,0 +1,68 @@
...
+
+1. Rename `*Provider` to `*Gateway`
+
+   The high level role of the whole component is that of a gateway. 
+
+1. Introduce a `Gateway` resource which creates a single `Deployment` and `Service`
+
+   This reduces the toil of adding a new `*Gateway` reconciler as the boilerplate code will be encapsulated. It is recommend that users not create this resource directly.
+   
+   1. Cohabitate the gateway and provisioner `Deployment`s and `Service`s
```
ericbottard:
```suggestion
   1. Cohabitate the gateway (liiklus) and provisioner `Deployment`s and `Service`s
```
Since gateway and Gateway are new terms here (at least for new readers of this RFC), I'm clarifying what lowercase gateway means in this context, at least once.

rfc/rfc-0011-streaming-gateways.md
```diff
@@ -0,0 +1,68 @@
...
+
+      A single `Deployment` with two containers and a single `Service` with two ports can replace the current two Deployments and two Services per provider.
+
+   1. Update each existing `*Gateway` to reconcile to a `Gateway`
+
+1. Update `Stream` resource with a LocalObjectReference to its `Gateway`
+
+   The `Stream` reconciler can track the `Gateway` and include the `Gateway`'s Ready condition as one of it's conditions. `Processor`s already track each input and output `Stream`'s Ready condition to scale down the processor when a stream goes red.
```
ericbottard:
```suggestion
   The `Stream` reconciler can track the `Gateway` and include the `Gateway`'s Ready condition as one of its conditions. `Processor`s already track each input and output `Stream`'s Ready condition to scale down the processor when a stream goes red.
```

rfc/rfc-0011-streaming-gateways.md
```diff
@@ -0,0 +1,68 @@
...
+
+1. Introduce a `Gateway` resource which creates a single `Deployment` and `Service`
+
+   This reduces the toil of adding a new `*Gateway` reconciler as the boilerplate code will be encapsulated. It is recommend that users not create this resource directly.
+   
+   1. Cohabitate the gateway and provisioner `Deployment`s and `Service`s
+
+      A single `Deployment` with two containers and a single `Service` with two ports can replace the current two Deployments and two Services per provider.
+
+   1. Update each existing `*Gateway` to reconcile to a `Gateway`
```
ericbottard: Maybe specify what a `Gateway` resource contains here then. I assume it's more than just the host as we discussed, if we want to do the heavy lifting in the generic `Gateway` reconciler: gateway (liiklus) image, provisioner image?  (both pod templates even I guess. Obviously the actual PR will have all the details, but I guess the RFC could benefit from some high level clarifications here)

rfc/rfc-0011-streaming-gateways.md
```diff
@@ -0,0 +1,68 @@
...
+
+   The `Stream` reconciler can track the `Gateway` and include the `Gateway`'s Ready condition as one of it's conditions. `Processor`s already track each input and output `Stream`'s Ready condition to scale down the processor when a stream goes red.
+
+1. [maybe] Pin a `Stream` to a `Gateway` instance
+
+   `Stream`s are inherently stateful. A new `Gateway`, even if it has the same name, is unlikely to contain the same state. The gateway's UID could be saved on the stream and compared. if the UID changes, the stream's Ready condition could flip False to prevent usage of the stream.
+
+1. Update Streaming CRDs status for new relationships
+
+  Changes to the reconciliation object graph invalidate may of the existing status references.
```
ericbottard: Not sure what you meant here.

rfc/rfc-0011-streaming-gateways.md
```diff
@@ -0,0 +1,68 @@
...
+
+### User Impact
+
+The CRDs and CLI commands for streaming providers will replace `provider` with `gateway`. The stream create command will rename the `--provider` flag to `--gateway` with the value as the gateway's name (no suffix).
```
ericbottard:
```suggestion
The CRDs and CLI commands for streaming providers will replace `provider` with `gateway`. The stream create command will rename the `--provider` flag to `--gateway` with the value as the gateway's *name* (with no suffix needed).
```

rfc/rfc-0011-streaming-gateways.md
scothis: Some CRDs define references on their status to resources that will no longer exist, or are managed by a different resource. For example, each provider has a reference to both the gateway and provisioner Deployments and Services, since those are going to be cohabitated, we don't need both sets of references. Or, all of those references could be dropped in favor of a single gateway reference

rfc/rfc-0011-streaming-gateways.md
scothis: At a minimum I'd imaging we have a PodSpecTemplate that would become the core of the Deployment. I haven't worked through the details of the Gateway CRD shape.

rfc/rfc-0011-streaming-gateways.md
scothis: this is talking about gateway names, not stream names

rfc/rfc-0011-streaming-gateways.md
ericbottard: My initial comment was mostly triggered by the fact that the sentence is malformed. Thanks for the clarification, need to condense that in a sentence now.

rfc/rfc-0011-streaming-gateways.md
ericbottard: Ah, of course.

rfc/rfc-0011-streaming-gateways.md
```diff
@@ -0,0 +1,68 @@
...
+**Superseded by:** N/A
+
+**Supersedes:** N/A
+
+**Related:** N/A
+
+
+## Problem
+
+The relationships between a `Stream` and a `(InMemory|Kafka|Pulsar)Provider` is fragile. The stream contains a hostname for the provider's provisioner, which the stream controller calls when reconciling the stream. This has a few problems:
```
markfisher:
```suggestion
The current relationship between a `Stream` and a `(InMemory|Kafka|Pulsar)Provider` is fragile. The stream contains a hostname for the provider's provisioner, which the stream controller calls when reconciling the stream. This has a few problems:
```

rfc/rfc-0011-streaming-gateways.md
```diff
@@ -0,0 +1,68 @@
...
+- the `Stream` resource doesn't know when the `Provider` is unavailable, or deleted
+- the `Stream` has no way to automatically obtain a new address for the provisioner
+- the provisioner name is competing in the `Service` namespace. The likelihood of a name collision is reduced by appending a suffix to the service name. However, the user must include this suffix when creating each stream
+- as new providers are created, there is a significant amount of reconciliation logic that is duplicated with trivial differences
+
+The problem is derived from [projectriff/system#86](https://github.com/projectriff/system/issues/86).
+
+### Anti-Goals
+
+This RFC will not prevent naming collisions between different provisioners with the same name. The first gateway of a given name wins.
```
markfisher:
```suggestion
This RFC will not prevent naming collisions between different provisioners within the same namespace. The first gateway of a given name wins.
```

rfc/rfc-0011-streaming-gateways.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Solution
+
+1. Rename `*Provider` to `*Gateway`
+
+   The high level role of the whole component is that of a gateway. 
+
+1. Introduce a `Gateway` resource which creates a single `Deployment` and `Service`
+
+   This reduces the toil of adding a new `*Gateway` reconciler as the boilerplate code will be encapsulated. It is recommend that users not create this resource directly.
```
markfisher:
```suggestion
   This reduces the toil of adding a new `*Gateway` reconciler as the boilerplate code will be encapsulated. It is recommended that users not create this resource directly. Rather one of these _abstract_ `Gateway`s will be created for each _concrete_ `*Gateway` resource within the latter's reconciler as described below.
```

rfc/rfc-0011-streaming-gateways.md
```diff
@@ -0,0 +1,68 @@
...
+   
+   1. Cohabitate the gateway (liiklus) and provisioner `Deployment`s and `Service`s
+
+      A single `Deployment` with two containers and a single `Service` with two ports can replace the current two Deployments and two Services per provider.
```
markfisher:
```suggestion
      A single `Deployment` with two containers and a single `Service` with two ports can replace the current two Deployments and two Services per provider. Eventually we may extend the gateway spec and implementation (currently liiklus) so that the provisioning role is handled by the same container (a new RFC would be created to capture that plan).
```

rfc/rfc-0011-streaming-gateways.md
```diff
@@ -0,0 +1,68 @@
...
+
+   1. Update each existing `*Gateway` to reconcile to a `Gateway`
+
+1. Update `Stream` resource with a LocalObjectReference to its `Gateway`
+
+   The `Stream` reconciler can track the `Gateway` and include the `Gateway`'s Ready condition as one of its conditions. `Processor`s already track each input and output `Stream`'s Ready condition to scale down the processor when a stream goes red.
+
+1. [maybe] Pin a `Stream` to a `Gateway` instance
+
+   `Stream`s are inherently stateful. A new `Gateway`, even if it has the same name, is unlikely to contain the same state. The gateway's UID could be saved on the stream and compared. if the UID changes, the stream's Ready condition could flip False to prevent usage of the stream.
```
markfisher: This feels somewhat related to the notion of reclaim policies. We might want to make this behavior configurable.

rfc/rfc-0011-streaming-gateways.md
```diff
@@ -0,0 +1,68 @@
...
+
+1. Update Streaming CRDs status for new relationships
+
+  Changes to the reconciliation object graph invalidate may of the existing status references.
```
markfisher:
```suggestion
  Changes to the reconciliation object graph invalidate many of the existing status references.
```

rfc/rfc-0011-streaming-gateways.md
```diff
@@ -0,0 +1,68 @@
...
+
+### User Impact
+
+The CRDs and CLI commands for streaming providers will replace `provider` with `gateway`. The stream create command will rename the `--provider` flag to `--gateway` with the value as the gateway's name (with no suffix needed).
+
+### Backwards Compatibility and Upgrade Path
+
+All existing behavior is being preserved, since the streaming runtime has not been released, there is no release to be backwards incompatible with.
```
markfisher:
```suggestion
All existing behavior is being preserved. Since the streaming runtime has not been released, there is no release to be backwards incompatible with.
```

rfc/rfc-0011-streaming-gateways.md
scothis: let's remove this point for now and consider it as a separate rfc

PR #1378, opened 2020-01-20
Add RFC-0013: Invoker TCK

ericbottard:
Comments due by Thursday Feb 22nd


rfc/rfc-0013-invoker-tck.md
```diff
@@ -0,0 +1,37 @@
...
+
+## Solution
+This RFC proposes to introduce a TCK that would take the form of at least a runnable tool (maybe also available as a go library) that focuses on testing the conformance of packaged functions with the _interaction (networking) protocols_ prescribed by the invoker specification.
+As such, it would require that users of the TCK produce a set of Docker images implementing a handful of trivial functions (such as a "square" function, see below). 
+How these images got constructed is deemed out of scope of the TCK, and while riff and the specification propose using buildpacks to construct such images, the TCK focuses on the resulting behavior of the container, which is assumed to result from the usage of the _invoker layer_.
+The use of containers as the unit of testing has several advantages:
+- the overall environment needed for the invoker to function properly does not have to be setup as part of the templated run of a test. Invokers implementors are in full control to provide a runnable container, and there is no software conflict with the environment used to drive the TCK run.
```
fbiville: ### 
```suggestion
- the overall environment needed for the invoker to function properly does not have to be set up as part of the templated run of a test. Invokers implementors are in full control to provide a runnable container, and there is no software conflict with the environment used to drive the TCK run.
```

rfc/rfc-0013-invoker-tck.md
```diff
@@ -0,0 +1,37 @@
...
+- using a container guarantees isolation of test runs, in particular with regard to data left behind.
+
+The TCK would roughly take the following form:
+A series of tests run that cover the actual requirements of the specification. In particular, the words "MUST" throughout the spec should correspond to actual testcases. To run such tests, the invoker implementor is asked to provide a series of containers that implement simple functions. For example, to test the http request/reply interaction and correct support for marshalling/unmarshalling of `application/json`, the TCK may leverage a container that assumes input as a number and returns its square. The TCK runner would run the container image configured for "square", POST a payload on `http://<host>:8080/` and verify synchronous return of a squared number, with `Content-Type` compatible with the requested `Accept` header.
```
fbiville:
```suggestion
A series of tests run that covers the actual requirements of the specification. In particular, the words "MUST" throughout the spec should correspond to actual testcases. To run such tests, the invoker implementor is asked to provide a series of containers that implement simple functions. For example, to test the http request/reply interaction and correct support for marshalling/unmarshalling of `application/json`, the TCK may leverage a container that assumes input as a number and returns its square. The TCK runner would run the container image configured for "square", POST a payload on `http://<host>:8080/` and verify synchronous return of a squared number, with `Content-Type` compatible with the requested `Accept` header.
```

ericbottard: It's the tests that cover, so actually, should read "A series of test runs that cover"
Will fix

fbiville: ach so, gotcha

rfc/rfc-0013-invoker-tck.md
```diff
@@ -0,0 +1,37 @@
...
+- the overall environment needed for the invoker to function properly does not have to be set up as part of the templated run of a test. Invokers implementors are in full control to provide a runnable container, and there is no software conflict with the environment used to drive the TCK run.
+- using a container guarantees isolation of test runs, in particular with regard to data left behind.
+
+The TCK would roughly take the following form:
+A series of test runs that cover the actual requirements of the specification. In particular, the words "MUST" throughout the spec should correspond to actual testcases. To run such tests, the invoker implementor is asked to provide a series of containers that implement simple functions. For example, to test the http request/reply interaction and correct support for marshalling/unmarshalling of `application/json`, the TCK may leverage a container that assumes input as a number and returns its square. The TCK runner would run the container image configured for "square", POST a payload on `http://<host>:8080/` and verify synchronous return of a squared number, with `Content-Type` compatible with the requested `Accept` header.
```
glyn: How will the TCK test streaming functions which consume input but which produce no output? I would have thought it would be necessary to inject a remote endpoint into the user-provided function so that the function can say what it has received. Without this, it will be hard to know that an invoker isn't somehow buffering input inappropriately.

fbiville: Consumers are not really supported at the moment (may work with Node, does not work with Java).
I would expect consumers and suppliers to be out of scope of this RFC.
There should be a subsequent RFC to evolve both the programmatic model of invokers and the TCK, in my opinion.

scothis: > How will the TCK test streaming functions which consume input but which produce no output?

StdOut/Err are always available

> Consumers are not really supported at the moment

That sounds like a bug rather than an intentional design. Do we have issues?

A Processor currently requires at least one input stream and no output streams.

ericbottard:
> If Stdout/Stderr from a function are easily accessed by the TCK, that addresses my issue.

As Scott stated, there is always options to witness effective behavior via side effects (strdout being one). But taking a step back, I'm not even sure this really matters:
- the TCK deliberately chooses to test a built image, not really knowing how it was built. If there is a defect with the **invoker** layer, it should be caught by other sets of tests.
- the spec doesn't state that invokers shouldn't buffer. If it did, I'm not sure we could state "how much" it shouldn't, ie some layers in the system may buffer (in particular, the gRPC layer, the reactive-grpc layer, etc). Is it a problem? I don't think so. What matters is that when the stream is closed, the invoker+fn stack gets the opportunity to process everything (ie flush).

glyn:
> > If Stdout/Stderr from a function are easily accessed by the TCK, that addresses my issue.
> 
> As Scott stated, there is always options to witness effective behavior via side effects (strdout being one). But taking a step back, I'm not even sure this really matters:
> 
> * the TCK deliberately chooses to test a built image, not really knowing how it was built. If there is a defect with the **invoker** layer, it should be caught by other sets of tests.
> * the spec doesn't state that invokers shouldn't buffer. If it did, I'm not sure we could state "how much" it shouldn't, ie some layers in the system may buffer (in particular, the gRPC layer, the reactive-grpc layer, etc). Is it a problem? I don't think so. What matters is that when the stream is closed, the invoker+fn stack gets the opportunity to process everything (ie flush).

You may be right, but I thought there could be behaviours which the TCK needs to check prior to the function returning any data to the invoker. I guess we'll find out later if this is the case.

rfc/rfc-0013-invoker-tck.md
```diff
@@ -0,0 +1,37 @@
...
+The TCK will devide its test suites into several groups, corresponding to several layers of conformance. For example, support for http request/reply interaction and gRPC streaming interaction would be tested separately and reported as such. Similarly, occurences of SHOULD or MAY in the spec will be covered by tests that report conformance but don't fail the run if the tests don't pass.
+
+### User Impact
+The term "user" here means implementors of a function invoker. The requirements for such users to be able to use the TCK would likely be:
+- ability to run a compiled go command line tool,
+- availability of a local Docker daemon.
```
ojhughes: Daemon-less should be possible with [Buildah](https://buildah.io)

ericbottard: we're _running_ containers here, not building them. There may be daemonless solutions still, but I don't think it should be a strong goal.

ericbottard:
> Should the TCK cover any technical constraints such as spawning a thread per function invocation vs non blocking IO / reactive?

the tck should cover the spec, and as such, this should be covered to the best extent of the spec (uses SHOULD), yes

rfc/rfc-0013-invoker-tck.md
fbiville: > A Processor currently requires at least one input stream and no output streams.

Reframing: using a `java.util.function.Consumer` is not supported at the moment, which would be the idiomatic way to express the kind of functions you describe. I personally have not tested this kind of functions, FATS does not either it seems.

ericbottard:
Merged as 12b25d1c

PR #1379, opened 2020-02-06
Add diagram of the relationship between riff repos

scothis:


scothis:
@glyn it's hand rolled. There are too many types of dependencies for tools to generate anything meaningful without a good bit of effort.

glyn:
> @glyn it's hand rolled. There are too many types of dependencies for tools to generate anything meaningful without a good bit of effort.

@scothis Then I'm concerned that this will go out of date and end up being misleading. I guess we can update it or delete it...

PR #1382, opened 2020-03-01
[DRAFT] RFC 0014: Binding Injection

scothis:
[readable](https://github.com/scothis/riff/blob/rfc-0014-binding-injection/rfc/rfc-0014-binding-injection.md)

This is an early draft open for feedback, but not ratification.

### DIFF
--- a/rfc/rfc-0014-binding-injection.md
+++ b/rfc/rfc-0014-binding-injection.md
@@ -51,7 +51,7 @@ status:
 - `.spec.binding` is an ObjectReference to a resource implementing the binding duck type
 - `.spec.containers` optionally restricts which containers are augmented, defaults to all containers
 - `.spec.subject` is an ObjectReference to a resource implementing the podspecable duck type
-- the subject and binding target much be in the same namespace as the injection resource
+- the subject and binding target mush be in the same namespace as the injection resource
 
 The subject resource will be injected with the binding metadata and secret.
 
--- a/rfc/rfc-0014-binding-injection.md
+++ b/rfc/rfc-0014-binding-injection.md
@@ -29,30 +29,33 @@ A duck type for exposing a CNB binding is defined in RFC 0002. While not wide sp
 
 ```yaml
 apiVersion: bindings.projectriff.io/v1alpha1
-kind: Injection
+kind: ServiceBinding
 metadata:
-  name: my-binding-injection
+  name: my-service-binding
 spec:
-  provider:
-    apiVersion: streaming.projectriff.io/v1alpha1
-    kind: Stream
-    name: my-stream
   subject:
     apiVersion: knative.projectriff.io/v1alpha1
     kind: Deployer
     name: my-deployer
-  allowedContainers:
-  - user-container
+  provider:
+    apiVersion: streaming.projectriff.io/v1alpha1
+    kind: Stream
+    name: my-stream
+  containerName: user-container # optional, defaults to all containers
+  bindingMode: Secret # optional, defaults to 'Secret'
 status:
   {}
 ```
 
-- `.spec.provider` is an ObjectReference to a resource implementing the CNB binding duck type
-- `.spec.subject` is an ObjectReference to a resource implementing the podspecable duck type
-- `.spec.allowedContainers` optionally restricts which containers are augmented, defaults to all containers
+- `.spec.subject` is a reference to a resource implementing the PodSpecable duck type
+- `.spec.provider` is a reference to a resource implementing the CNB binding duck type
+- `.spec.containerName` optionally restricts which container is augmented, defaults to all containers
+- `.spec.bindingMode` optionally restricts which parts of the binding is exposed to the subject. There are two values:
+  - `"Metadata"` - only the non-sensitive, metadata element of the binding is mounted
+  - `"Secret"` (default) - the metadata and secret element of the binding are both mounted
 - the `subject` and `provider` targets must be in the same namespace as the injection resource
 
-The subject resource will be injected with the CNB binding's metadata and secret.
+The subject resource is injected with the CNB binding's metadata and secret.
 
 ```yaml
 apiVersion: streaming.projectriff.io/v1alpha1
@@ -95,19 +98,19 @@ spec:
         image: square
         env:
         - name: CNB_BINDINGS
-          value: /var/bindings
+          value: /platform/bindings
         volumeMounts:
-        - mountPath: /var/bindings/my-binding-injection/metadata
-          name: my-binding-injection-metadata
+        - mountPath: /platform/bindings/my-service-binding/metadata
+          name: my-service-binding-metadata
           readOnly: true
-        - mountPath: /var/bindings/my-binding-injection/secret
-          name: my-binding-injection-secret
+        - mountPath: /platform/bindings/my-service-binding/secret
+          name: my-service-binding-secret
           readOnly: true
     volumes:
-    - name: my-binding-injection-metadata
+    - name: my-service-binding-metadata
       configMap:
         name: my-stream-binding-metadata
-    - name: my-binding-injection-secret
+    - name: my-service-binding-secret
       secret:
         secretName: my-stream-binding-secret
 ```
@@ -115,6 +118,7 @@ spec:
 Notes:
 - each container may only define the `CNB_BINDINGS` env var once, if already defined the value must be preserved and respected
 - injected volume names must not collide with an existing volume
+  - the volume name has no semantic meaning and may be any value unique within the pod
 - injected volume mount paths must not collide with an existing volume
 - injected volume mount paths must compile with RFC 0002
 
--- a/rfc/rfc-0014-binding-injection.md
+++ b/rfc/rfc-0014-binding-injection.md
@@ -1,0 +1,131 @@
+# RFC-0014: Binding Inject
+
+**Authors:** @scothis
+
+**Status:**
+
+**Pull Request URL:** link
+
+**Superseded by:** N/A
+
+**Supersedes:** N/A
+
+**Related:** [RFC 0002 Bindings](./rfc-0002-bindings.md), [RFC 0008 Pod Templates](./rfc-0008-pod-templates.md)
+
+
+## Problem
+
+[RFC 0002 Bindings](./rfc-0002-bindings.md) defines a interface for a resource to expose metadata and secrets to be consumed by another resource. It leverages the [CNB Binding](https://github.com/buildpack/spec/blob/master/extensions/bindings.md) spec for how the service metadata and secrets are exposed inside a container. While a binding can be consumed manually, it's difficult to manage as it requires defining two volumes, and two volume mounts and a environment variable per container.
+
+### Anti-Goals
+
+This RFC has no opinion as to the type of binding being injected, or the consumer of the binding.
+
+## Solution
+
+Knative Bindings (yea, the term is overloaded) define an abstract way for reconcilers to operate on resources by duck typing known elements of the resource without requiring pre-knowledge or full type knowledge at runtime. By leveraging resources that use a PodTemplateSpec, it's possible to inject volumes, volume mounts and environment variables into an existing resource (even if owned and controlled by another resource). With [RFC 0008 Pod Templates](./rfc-0008-pod-templates.md) the riff Deployers and Processor resources implement the PodSpecable duck type.
+
+A duck type for exposing a binding is defined in RFC 0002. While not as wide spread, the Stream resource currently implements this duck type. Combining the Binding and PodSpecable duck types allows for a generic means to inject binding metadata and secrets into a consuming workload. A new CRD can express this injection:
+
+```yaml
+apiVersion: bindings.projectriff.io/v1alpha1
+kind: Injection
+metadata:
+  name: my-binding-injection
+spec:
+  binding:
+    apiVersion: streaming.projectriff.io/v1alpha1
+    kind: Stream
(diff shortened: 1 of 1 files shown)


rfc/rfc-0014-binding-injection.md
```diff
@@ -0,0 +1,131 @@
...
+    name: my-deployer
+status:
+  observedGeneration: 1
+  conditions: []
+```
+
+- `.spec.binding` is an ObjectReference to a resource implementing the binding duck type
+- `.spec.containers` optionally restricts which containers are augmented, defaults to all containers
+- `.spec.subject` is an ObjectReference to a resource implementing the podspecable duck type
+- the subject and target much be in the same namespace as the injection resource
```
dsyer: By "target" we mean "binding"?

rfc/rfc-0014-binding-injection.md
```diff
@@ -0,0 +1,131 @@
...
+  subject:
+    apiVersion: knative.projectriff.io/v1alpha1
+    kind: Deployer
+    name: my-deployer
+status:
+  observedGeneration: 1
+  conditions: []
+```
+
+- `.spec.binding` is an ObjectReference to a resource implementing the binding duck type
```
dsyer: Is there a definition of this duck type somewhere? I'm missing some detail.

rfc/rfc-0014-binding-injection.md
```diff
@@ -0,0 +1,131 @@
...
+    - name: my-binding-injection-metadata
+      configMap:
+        name: my-stream-binding-metadata
+    - name: my-binding-injection-secret
+      secret:
+        secretName: my-stream-binding-secret
+```
+
+Notes:
+- if the `CNB_BINDINGS` env var is already defined, it must be respected
```
dsyer: This seems like maybe a mistake. CNB defines a default value for the environment variable already, so why not use that? Locating the bindings directory in `/var/riff` kind of blocks other options - you can't bind to something that lives in `/var/somethingelse` as well (because `CNB_BINDINGS` is global).

rfc/rfc-0014-binding-injection.md
```diff
@@ -0,0 +1,131 @@
...
+  - otherwise it must be defined
+- volume names must not collide with an existing volume
+- volume mount paths must be under the CNB_BINDINGS value
+- the mount directory within CNB_BINDINGS must be unique
```
dsyer: I think it should have the same name as the binding object. It looks like `my-binding-injection` in this example was pulled out of a hat. Or am I missing something?

dsyer:
This looks like a good starting point. Using CNB bindings seems like a good idea, but it still forces the app to know about that in some detail. I would like to be able to inject other things to adapt apps from a specific stack to the CNB layout. An init container, and an environment variable are the two examples that come to mind that I have needed to adapt Spring Boot apps to CNB bindings.

I added some comments inline as well.

rfc/rfc-0014-binding-injection.md
dsyer: It's rfc-0002 I guess. If I have comments on that where do they go (since it is "approved" I assume it is basically immutable)?

rfc/rfc-0014-binding-injection.md
ericbottard:
```suggestion
- the subject and target must be in the same namespace as the injection resource
```

rfc/rfc-0014-binding-injection.md
ericbottard: I, OTOH think it should have the name of the *Injection* object, or something derived from it.
A single binding can be injected multiple times with different "roles". For example, a single Stream can be used as different inputs (or different outputs) of a Processor.

Have a look at the http-source [PR](https://github.com/projectriff/http-source/pull/9#issue-347115960) for cardinalities that make sense IMO.

rfc/rfc-0014-binding-injection.md
```diff
@@ -0,0 +1,131 @@
...
+
+[RFC 0002 Bindings](./rfc-0002-bindings.md) defines a interface for a resource to expose metadata and secrets to be consumed by another resource. It leverages the [CNB Binding](https://github.com/buildpack/spec/blob/master/extensions/bindings.md) spec for how the service metadata and secrets are exposed inside a container. While a binding can be consumed manually, it's difficult to manage as it requires defining two volumes, and two volume mounts and a environment variable per container.
+
+### Anti-Goals
+
+This RFC has no opinion as to the type of binding being injected, or the consumer of the binding.
+
+## Solution
+
+Knative Bindings (yea, the term is overloaded) define an abstract way for reconcilers to operate on resources by duck typing known elements of the resource without requiring pre-knowledge or full type knowledge at runtime. By leveraging resources that use a PodTemplateSpec, it's possible to inject volumes, volume mounts and environment variables into an existing resource (even if owned and controlled by another resource). With [RFC 0008 Pod Templates](./rfc-0008-pod-templates.md) the riff Deployers and Processor resources implement the PodSpecable duck type.
```
ericbottard: > Knative Bindings (yea, the term is overloaded) define an abstract way ...

A pointer to anything showing how this works would be useful here I think

fbiville: > By leveraging resources that use a PodTemplateSpec, it's possible to inject volumes, volume mounts and environment variables into an existing resource (even if owned and controlled by another resource)

Is there any (external) resource that describe this in more details?

rfc/rfc-0014-binding-injection.md
```diff
@@ -0,0 +1,131 @@
...
+    kind: Deployer
+    name: my-deployer
+status:
+  observedGeneration: 1
+  conditions: []
+```
+
+- `.spec.binding` is an ObjectReference to a resource implementing the binding duck type
+- `.spec.containers` optionally restricts which containers are augmented, defaults to all containers
+- `.spec.subject` is an ObjectReference to a resource implementing the podspecable duck type
```
ericbottard: I know it's already too late / out of scope to reconsider the word "binding" but to me, the "binding" is the act to bind, so would rather map to what you call "injection".

I also have issues with the word "subject" though. Subject to me means something that is "acting". Here, the subject is more akin to a "destination", something _where_ the injection happens.

The combination of ambiguity on both "binding" and "subject" would make this mechanism hard to use IMO, never knowing which is which. The fact that you use the word "target" below is interesting too, as I would reserve that word for what is here called the subject, because this conveys the idea of injecting _into_.

dsyer: Ditto (I had the same thought about "target").

rfc/rfc-0014-binding-injection.md
dsyer: I think that's what I meant. A `Binding` is "global" (could be shared by multiple `Injections`, or even multiple apps). I don't have the same hang up about using the name `binding` to refer to the "injection" though.

rfc/rfc-0014-binding-injection.md
scothis: For basic errata we can fx with a PR to the current rfc. To change the semantics or scope, we should create a new rfc that supersedes the rfc in question.

For now, feel free to ask here.

rfc/rfc-0014-binding-injection.md
scothis: Where does CNB define a default value for CNB_BINDINGS?

The situation you describe why this lines exists. CNB_BINDINGS can only hold a single value, if it is already defined, the new bindings should be mounted under that path.

scothis:
> I would like to be able to inject other things to adapt apps from a specific stack to the CNB layout. An init container, and an environment variable are the two examples that come to mind that I have needed to adapt Spring Boot apps to CNB bindings.

@dsyer injecting other elements is out of scope for this RFC, but is something that should be relatively easy to do by creating other Knative Bindings with a PodSpecable subject.

rfc/rfc-0014-binding-injection.md
scothis: > I also have issues with the word "subject" though. Subject to me means something that is "acting". Here, the subject is more akin to a "destination", something where the injection happens.

`subject` comes from Knative Bindings, which we should align with. I'm still not fully up to speed on the exact semantics, so the the shape of the resource will likely evolve as we better understand the implementation. This is the main reason why this RFC is a draft.

rfc/rfc-0014-binding-injection.md
scothis:
```suggestion
- the subject and binding target much be in the same namespace as the injection resource
```

rfc/rfc-0014-binding-injection.md
scothis: Since the path for the volume mount must be unique, there will necessarily need to be a bit of massaging at runtime to avoid collisions. Don't get hung up on what the name is.

rfc/rfc-0014-binding-injection.md
scothis:
```suggestion
[Knative Bindings](https://docs.google.com/document/d/1t5WVrj2KQZ2u5s0LvIUtfHnSonBv5Vcv8Gl2k5NXrCQ/edit#heading=h.lnql658xmg9p) [Knative membership required to view] (yea, the term is overloaded) define an abstract way for reconcilers to operate on resources by duck typing known elements of the resource without requiring pre-knowledge or full type knowledge at runtime. By leveraging resources that use a PodTemplateSpec, it's possible to inject volumes, volume mounts and environment variables into an existing resource (even if owned and controlled by another resource). With [RFC 0008 Pod Templates](./rfc-0008-pod-templates.md) the riff Deployers and Processor resources implement the PodSpecable duck type.
```

scothis: added a link the the Knative Bindings doc

rfc/rfc-0014-binding-injection.md
dsyer: I guess we should clarify by "must be respected" means in practice then. I was under the impression that CNB_BINDINGS had a suggested default value. Maybe I misread their spec.

scothis: boiling this down to the principles:
- we cannot change an existing CNB_BINDINGS var as there may be existing volume mounts located under that path
- the resulting container must have exactly one CNB_BINDINGS env var with the path which the volume mounts are mounted under

So, if there is an existing value, we need to add our volume mounts under the path that is defined by the existing var, otherwise we need to create the var with the path where we mounted the volumes.

rfc/rfc-0014-binding-injection.md
dsyer: OK. It seemed wrong to me in rfc-0002 that the producer and consumer both have to explicitly state the config map and secret refs. In a way, maybe this RFC (0014) is a step towards correcting that - you only need the binding ref to consume it via an `Injection`.

scothis: It's not a matter of right and wrong. RFC 0002 defined a low level mechanism to support a CNB Bindings in a k8s workload. Defining a high level the user experience wasn't the goal. RFC 0014 is proposing a higher level experience.

rfc/rfc-0014-binding-injection.md
```diff
@@ -0,0 +1,131 @@
...
+
+### Anti-Goals
+
+This RFC has no opinion as to the type of binding being injected, or the consumer of the binding.
+
+## Solution
+
+Knative Bindings (yea, the term is overloaded) define an abstract way for reconcilers to operate on resources by duck typing known elements of the resource without requiring pre-knowledge or full type knowledge at runtime. By leveraging resources that use a PodTemplateSpec, it's possible to inject volumes, volume mounts and environment variables into an existing resource (even if owned and controlled by another resource). With [RFC 0008 Pod Templates](./rfc-0008-pod-templates.md) the riff Deployers and Processor resources implement the PodSpecable duck type.
+
+A duck type for exposing a binding is defined in RFC 0002. While not as wide spread, the Stream resource currently implements this duck type. Combining the Binding and PodSpecable duck types allows for a generic means to inject binding metadata and secrets into a consuming workload. A new CRD can express this injection:
```
jldec: i suggest qualifying the word "Binding" here so that readers don't confuse the CNB service binding idea with the more generic Knative Binding. I also suggest not using the word "Binding" to qualify the kind of duck type for the same reason.

rfc/rfc-0014-binding-injection.md
dsyer: I think we managed to go from "much " to "mush". Only one more step to "must"...

rfc/rfc-0014-binding-injection.md
jldec: @dsyer to answer you question about defining duck types, my read of the Knative Bindings [doc here](https://docs.google.com/document/d/1t5WVrj2KQZ2u5s0LvIUtfHnSonBv5Vcv8Gl2k5NXrCQ/edit#heading=h.tur8dkr939w4), is that duck types are the common schema fragments which implementations of Bindings choose to standardize on.

rfc/rfc-0014-binding-injection.md
scothis: 🤦‍♂

rfc/rfc-0014-binding-injection.md
jldec: Thanks, that's better now - resolving.

scothis:
Closing, as CNB Binding is deprecated in favor of [Service Bindings](https://github.com/k8s-service-bindings/spec). Will open a new RFC

scothis:
Updated at #1390

